PR #14, opened 2018-01-18
Vect hierarchy

### LINKED ISSUE #12: Instantiate the hierarchy on row vectors
This work was done for Coquelicot's hierarchy in https://github.com/drouhling/LaSalle/blob/master/vect.v and should be adapted to the new hierarchy to remove the `Admitted` in https://github.com/math-comp/analysis/blob/e2d3d4c57f3c0797d06891669c336ba587eb2ae5/derive.v#L174

Maybe move the instantiations to the corresponding sections in `hierarchy.v`?

drouhling:
Close #12

### DIFF
--- a/hierarchy.v
+++ b/hierarchy.v
@@ -1,7 +1,8 @@
 (* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
 Require Import Reals.
 From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
-From mathcomp Require Import seq fintype bigop ssralg ssrnum finmap.
+From mathcomp Require Import seq fintype bigop ssralg ssrnum finmap matrix.
+From mathcomp Require Import zmodp.
 From SsrReals Require Import boolp reals.
 Require Import Rstruct Rbar set posnum.
 
@@ -103,6 +104,13 @@ Canonical default_filter_on_term Y (X : filteredType Y) (x : X) :=
 
 Definition filter_from {I T : Type} (D : set I) (B : I -> set T) : set (set T) :=
   [set P | exists2 i, D i & B i `<=` P].
+
+(* the canonical filter on matrices on X is the product of the canonical filter
+   on X *)
+Canonical matrix_filtered m n X (Z : filteredType X) : filteredType 'M[X]_(m, n) :=
+  FilteredType 'M[X]_(m, n) 'M[Z]_(m, n) (fun mx => filter_from
+    [set P | forall i j, locally (mx i j) (P i j)]
+    (fun P => [set my : 'M[X]_(m, n) | forall i j, P i j (my i j)])).
 
 Definition filter_prod {T U : Type}
   (F : set (set T)) (G : set (set U)) : set (set (T * U)) :=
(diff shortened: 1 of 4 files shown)


hierarchy.v
```diff
@@ -2981,7 +3100,74 @@ End NormedModule2.
...
+
+(** ** Vectors *)
+
+Section Matrix_LMod.
+
+Variables (R : ringType) (V : lmodType R) (m n : nat).
+
+Definition mscale x (A : 'M[V]_(m, n)) := \matrix_(i, j) (x *: (A i j)).
+
+Program Definition matrix_LModMixin := @LmodMixin _ _ mscale _ _ _ _.
```
CohenCyril: `matrix_lmodMixin` is already defined in mathcomp/algebra/matrix https://github.com/math-comp/math-comp/blob/d4bd8f6a29a225cf7418c54008b6493501c62bd3/mathcomp/algebra/matrix.v#L1277 isn't it the same?

hierarchy.v
```diff
@@ -2981,7 +3100,74 @@ End NormedModule2.
...
+Qed.
+Next Obligation.
+apply/matrixP => i j; rewrite ord1 mxE; apply/eqP.
+rewrite -normm_le0 -H -[j](nth_ord_enum ord0).
+rewrite -(nth_map _ 0 (fun i => `|[x 0 i]|)) ?size_enum_ord //.
+by apply: bigmaxr_ler; rewrite size_map size_enum_ord.
+Qed.
+
+Canonical rV_NormedType :=
+  NormedModule.Pack (Phant _) (@NormedModule.Class _ _
```
CohenCyril: I am a bit surprised that you need to use the internal `Pack` here. Is there a problem (bug?) using `NormedModType` instead?
(Additionally, using Pack directly is dangerous because it might mean the intermediate structures from the hierarchy have not been declared as canonical first...)

hierarchy.v
drouhling: It isn't. The mixin from mathcomp is defined for matrices over a `ringType`, while I need a `lmodType` structure for matrices over a `lmodType`.

The fact that matrices over a `ringType` have a `lmodType` structure can be derived from this and the fact that a `ringType` is a `lmodType` over itself. Maybe this should be an addition to mathcomp.

drouhling:
The generalization to matrices shouldn't take much time.

I am not sure however to understand your second point. Vectors inherit the structure of the type of the coordinates so if I have only a `uniformType` I won't be able to build this partial `normedModType` mixin.

hierarchy.v
drouhling: Unfolding the notation and filling in the holes proves that indeed every structure but the `lmodType` one is inferred. Coq accepts the following lines:
```
Canonical rV_NormedType :=
  @NormedModule.pack _ (Phant K) 'rV[V]_n.+1 _ _ _
  rV_NormedModMixin (LmodType _ _ (matrix_LModMixin _ _ _))
  (GRing.Lmodule.Class (matrix_LModMixin _ _ _)) id _ _ id _ id.
```

hierarchy.v
CohenCyril: Yes, maybe this is worth opening an issue in mathcomp. I'm still not sure how to solve this problem yet...

PR #17, opened 2018-01-24
Notations near=> and near:

CohenCyril:
- begin_near x becomes near=> x
- begin_near2 x y becomes near=> x y
- near becomes near:
- have near F x becomes near F have x
- disclaimers:
  + near=> is *not* followed by an intro pattern but just one or two variables
  + near: is *not* followed by a "revert" parttern but just one variable.


CohenCyril:
Actually, I'm merging, it makes proof more "ssreflect"-like despite the discrepancy with => and : of ssreflect (hopefully, one day we can have ssr intro patterns in Tactic Notations)

PR #21, opened 2018-02-07
Start documenting

drouhling:
Document `set.v`, `posnum.v` and `topology.v`.

I would like to discuss a few points:

- in `posnum.v`:
  - my understanding of the `!! x` notation is not perfect (in particular how exactly should it be used?) so I would like help for its documentation.
  - the `[gt0 of x]` notation is defined twice, is there any reason?

- in `topology.v`:
  - `Filtered.Source` is used several times to define canonical instances of `Filtered.source`. It would seem reasonable to export it to get rid of the `Filtered` prefix.
  - I haven't documented `in_filter` since it seems to be only a tool for the near tactics, should I?
  - I haven't documented `filter_on` and `pfilter_on` because I think something should be done about them. Sometimes we use these structures, sometimes we use the type classes `Filter` and `ProperFilter` (we also have `UltraFilter`). Shouldn't we choose one mechanism once and for all instead of mixing them, or do I miss the point of these structures?


CohenCyril:
> in posnum.v:
> 
> my understanding of the !! x notation is not perfect (in particular how exactly should it be used?) so I would like help for its documentation.

The notation `!! t` by Enrico Tassi (cf issue coq or mathcomp to be found... @gares would you know quickly which one?) triggers pretyping, to fill the holes of the term `t`. The main purpose is to trigger typeclass inference in the `body` a ssreflect `have := !! body` tactic use case.

> the [gt0 of x] notation is defined twice, is there any reason?
Yes, once inside a section (so it disappears in its end) and once outside for exporting.
Ideally the format should be factored out by a `Reserved Notation` at the begging of the file, but this is independent from this PR about doc. We should open an issue for that (regrouping all problems of the same nature).

> in topology.v:
> 
> Filtered.Source is used several times to define canonical instances of Filtered.source. It would seem reasonable to export it to get rid of the Filtered prefix.
Maybe maybe not, I do not know what's best. It's good it's been identified by documenting, but we should open an issue for that as it is independent from the doc PR.

> I haven't documented in_filter since it seems to be only a tool for the near tactics, should I?
Indeed, it's an internal construction, do not document.

> I haven't documented filter_on and pfilter_on because I think something should be done about them. Sometimes we use these structures, sometimes we use the type classes Filter and ProperFilter (we also have UltraFilter). Shouldn't we choose one mechanism once and for all instead of mixing them, or do I miss the point of these structures?
Actually, Coquelicot's Hierarchy was designed using typeclasses for filters and we kept that. It would be a major review of the code to replace them by their canonical structure counterpart, and we should definitely try, as I suspect it would shorten some proof searchs (and fix the few ones that are broken). But it might introduce other problems that I did not assess yet. So let's open an issue.

On the other hand, in the current status of the code, the canonical structures `filter_on` and `pfilter_on` are the only way to make Landau work. TypeClass resolution can call Canonical Structure inference by unification, by CS has no mean to invoke TC resolution in the current state of this code. In other words, once you go CS, you cannot rely on TC anymore to do a complementary job.

posnum.v
```diff
@@ -4,6 +4,24 @@ From Coq Require Import ssreflect ssrfun ssrbool.
 From mathcomp Require Import ssrnat eqtype choice ssralg ssrnum.
 From SsrReals Require Import boolp reals.
 
+(******************************************************************************)
+(* This file develops tools to make the manipulation of positive numbers      *)
+(* easier, thanks to canonical structures.                                    *)
+(*                                                                            *)
```
CohenCyril: - Add that goals of the form `x > 0` where `x` is canonically a posnum are solved automatically.
- Add that posnum are canonically stable by addition, multiplication, inverse, min, sqrt, and all manifest positive natural numbers (i.e. (n.+1)%:R) are canonically positive

set.v
```diff
@@ -2,6 +2,81 @@ From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
...
 
+(******************************************************************************)
+(* This file develops a basic theory of sets and types equipped with a        *)
+(* canonical inhabitant (pointed types).                                      *)
+(*                                                                            *)
+(* --> A decidable equality is defined for any type. It is thus possible to   *)
+(*     define an eqType structure for any type using the mixin gen_eqMixin.   *)
+(* --> This file adds the possibility to define a choiceType structure for    *)
+(*     any type thanks to an axiom gen_choiceMixin giving a choice mixin.     *)
+(*                                                                            *)
```
CohenCyril: Explain we chose the option to have generic mixins over global canonical structures because we must let the user keep control over the definition he wants for equality. e.g. providing any type in Type with an equality canonical structure is dangerous because for example `nat := nat` equality could be inferred in two different ways (the generic uncomputable one by direct inference, and the `eqn` by delta-reduction first)

set.v
```diff
@@ -2,6 +2,81 @@ From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
...
+(* * Sets:                                                                    *)
+(*                       set A == type of sets on A.                          *)
+(*                   (x \in P) == membership predicate for set P.             *)
```
CohenCyril: generic **boolean** membership predicate (from `ssrbool`)

set.v
```diff
@@ -2,6 +2,81 @@ From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
...
+(*                 \bigcup_i F == same as before with T left implicit.        *)
+(*          \bigcap_(i in P) F == intersection of the elements of the family  *)
+(*                                F whose index satisfies P.                  *)
+(*           \bigcap_(i : T) F == union of the family F indexed on T.         *)
+(*                 \bigcap_i F == same as before with T left implicit.        *)
+(*                   A `<=` B <-> A is included in B.                         *)
+(*                  A `<=>` B <-> double inclusion A `<=` B and B `<=` A.     *)
+(*                   f @^-1` A == preimage of A by f.                         *)
+(*                      f @` A == image of A by f.                            *)
+(*                   A !=set0 <-> A is non empty.                             *)
```
CohenCyril: "non empty" is too ambiguous, put the explicit definition in terms of sigma

set.v
```diff
@@ -2,6 +2,81 @@ From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
...
+(*                  is_prop X <-> X contains only 1 element.                  *)
```
CohenCyril: `is_prop` was a bad choice of name, we should open an issue to rename it to `is_singleton`

drouhling:
> > the [gt0 of x] notation is defined twice, is there any reason?
>
> Yes, once inside a section (so it disappears in its end) and once outside for exporting.
Ideally the format should be factored out by a Reserved Notation at the beginning of the file, but this is independent from this PR about doc. We should open an issue for that (regrouping all problems of the same nature).

In fact, both are outside, hence my question.

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                                     inferred from the type of the          *)
+(*                                     canonical filter associated to F.      *)
+(*                    [cvg F in T] <-> the canonical filter associated to F   *)
+(*                                     converges in T.                        *)
+(*                           cvg F <-> same as [cvg F in T] where T is        *)
+(*                                     inferred from the type of the          *)
+(*                                     canonical filter associated to F.      *)
+(*                         Filter F == type class proving that the set of     *)
+(*                                     sets F is a filter.                    *)
+(*                  ProperFilter' F == type class proving that the set of     *)
```
CohenCyril: `ProperFilter'` is deprecated

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*       prove G x for x "near F", i.e. for x such that H x for H arbitrarily *)
+(*       precise as long as F H. The near tactics allow for a delayed         *)
+(*       introduction of H: H is introduced as an existential variable and    *)
+(*       progressively instantiated during the proof process.                 *)
+(*   --> Notations:                                                           *)
+(*                      {near F, P} == the property P holds near the          *)
+(*                                     canonical filter associated to F; P    *)
+(*                                     must have the form forall x, Q x.      *)
+(*                                     Equivalent to F Q.                     *)
+(*            \forall x \near F, P <-> F (fun x => P).                        *)
```
CohenCyril: make the `x` apparent in `P`: `\forall x \near F, P x <-> F (fun x => P x).`

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                       \near x, P == same as \forall y \near x, P.          *)
```
CohenCyril: `\near x, P x` == same as \forall y \near x, P y.` (this was wrong without explicit `x`)

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                  {near F & G, P} == same as {near H, P}, where H is the    *)
+(*                                     product of the filters F and G.        *)
+(*   \forall x \near F & y \near G, P == same as                              *)
```
CohenCyril: explicit `x` and `y`

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                                     {near F & G, forall x y, P}.           *)
+(*         \forall x & y \near F, P == same as before, with G = F.            *)
```
CohenCyril: here too

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                   \near x & y, P == same as                                *)
```
CohenCyril: here too

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*       hypothesis ?H x, delays the proof of F ?H and asks the user to       *)
+(*       prove G x. ?H is tagged with x in order to know when to instantiate  *)
+(*       it.                                                                  *)
+(*       Also exists under the form near=> x y.                               *)
+(*     - near: x : on the goal H_i x, refines the existential variable ?H     *)
+(*       tagged with x by intersection with H_i and closes the goal. H_i must *)
+(*       contain only variables introduced before x was. The proof that H_i   *)
+(*       holds near F is delayed.                                             *)
+(*     - end_near : when every main subgoal has been proved, the user has to  *)
+(*       prove that the intersection of some family (H_i)_i is in F. The      *)
```
CohenCyril: finite familiy

topology.v
```diff
@@ -4,6 +4,186 @@ From mathcomp Require Import seq fintype ssralg finmap matrix.
...
+(*                                     excluded.                              *)
+(*                        closure A == closure of the set A.                  *)
+(*                           closed == set of closed sets.                    *)
+(*                        cluster F == set of cluster points of F.            *)
+(*                          compact == set of compact sets w.r.t. the filter- *)
+(*                                     based definition of compactness.       *)
+(*                     hausdorff T <-> T is a Hausdorff space (T_2).          *)
+(*                    cover_compact == set of compact sets w.r.t. the open    *)
+(*                                     cover-based definition of compactness. *)
+(******************************************************************************)
```
CohenCyril: Highlight Zorn's Lemma and Tychonof Theorem.

topology.v
CohenCyril: no need to put both `==` and `same as` BTW

CohenCyril:
> the [gt0 of x] notation is defined twice, is there any reason?
> 
> Yes, once inside a section (so it disappears in its end) and once outside for exporting.
> Ideally the format should be factored out by a Reserved Notation at the beginning of the file, but this is independent from this PR about doc. We should open an issue for that (regrouping all problems of the same nature).
> 
> In fact, both are outside, hence my question.

Woops, I guess its redundant then...

PR #28, opened 2018-02-08
subsetI, subIset

affeldt-aist:


CohenCyril:
Lets's have the equational formulations as in mathcomp.

CohenCyril:
I did not see you updated your PR :+1:

PR #30, opened 2018-02-08
Implement infimum

drouhling:
I need this to port my proof that segments of `R` are compact but I think this deserves a separate PR.

### DIFF
--- a/coq-alternate-reals/src/reals.v
+++ b/coq-alternate-reals/src/reals.v
@@ -31,18 +31,22 @@ Variable (R : archiFieldType).
 Definition nonempty (E : pred R) :=
   exists x : R, x \in E.
 
-(* Real upper bound set. *)
+(* Real upper bound and lower bound sets. *)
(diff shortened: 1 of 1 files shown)


CohenCyril:
Yes we should.

coq-alternate-reals/src/reals.v
```diff
@@ -178,6 +182,9 @@ Export Real.Exports.
 
 (* -------------------------------------------------------------------- *)
 Definition sup {R : realType} := Real.sup (Real.class R).
+Notation "-` E" := [pred x | - x \in E]
```
CohenCyril: This notation should be made local, or should be in the company of many similar definitions.

CohenCyril:
> Shouldn’t we move this file at the root level ?

@strub I created issue #31 for this purpose.

coq-alternate-reals/src/reals.v
CohenCyril: (And if not local, moved to boolp)

PR #39, opened 2018-02-09
Prove that segments in R are connected, closed and compact.

drouhling:
Here is the proof I mentioned on the dev mailing list.

### DIFF
--- a/hierarchy.v
+++ b/hierarchy.v
@@ -2,6 +2,7 @@
 Require Import Reals.
 From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice.
 From mathcomp Require Import seq fintype bigop ssralg ssrnum finmap matrix.
+From mathcomp Require Import interval.
 Require Import boolp reals.
 Require Import Rstruct Rbar set posnum topology.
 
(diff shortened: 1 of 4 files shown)


hierarchy.v
```diff
@@ -421,6 +421,38 @@ Canonical absRing_topologicalType.
 Coercion absRing_UniformType (K : absRingType) := UniformType K AbsRingUniformMixin.
 Canonical absRing_UniformType.
 
+Lemma locallyZ (R : absRingType) (x : R) :
```
CohenCyril: Why suffix `Z`? It is about a negation, so it should be `N` instead.

drouhling: I wasn't quite sure about that… I did the same mistake about infimum. I'm fixing this.

hierarchy.v
```diff
@@ -1874,6 +1906,138 @@ rewrite [X in closed X](_ : (eq^~ _) = ~` (xpredC (eq_op^~ y))).
 by rewrite predeqE /setC => x /=; rewrite (rwP eqP); case: eqP; split.
 Qed.
 
+(** properties of segments in [R] *)
+
+Definition seg (a b : R) := (>= a) `&` (<= b).
```
CohenCyril: Maybe you could use the development library math-comp/algebra/interval.v to write `` `[a ,b]`` instead?

CohenCyril: Oh and the abbreviation `seg` is not used that much since we are not developing the theory of segments, so maybe we should use the full name `segment` instead.

hierarchy.v
drouhling: `N` is conflicting with a few lemmas about complement. Is `Nr` (for the ring negation, hence `opp`) ok?

CohenCyril: The abbreviation for complement in mathcomp is `C`, where are the conflicts?

CohenCyril: https://github.com/math-comp/math-comp/blob/d6bc72cd477ed6fe8b95782b26a2e0fc92711395/mathcomp/ssreflect/finset.v#L89

drouhling: `openN`, `closedN`, `locallyN`…
I am renaming them with `C` then, and using `N` for this PR.

Issue #46, opened 2018-03-01
Forms library?


amahboubi:
Where is the 'forms' library imported by derive in the diff branch?

https://github.com/math-comp/analysis/blob/6b36593f4a6a612212163b25c6bad3522c7fa679/derive.v#L4

amahboubi:
Ha, I guess the answer is commit 7b3288860c8d81c98b37dee80b94d4669635be33, hence it is in the algebra packages of the 'forms' branch of mathcomp...

In my opinion it is not very convenient to rely on a such a branch...

CohenCyril:
Yes I agree, I found no better solution though :(. Do you have any idea?

amahboubi:
Not sure it is the best place to ask but what is the status of the forms branch in mathcomp? It seems the last commit was on September 2016, and the only difference is the "forms" file. Is this case?

CohenCyril:
This branch has been created in the few last week even though the commit pre-dates it by far (it is a squash of commits from the repo CohenCyril/spectral). The reason is that we have started reworking on it with Laurence and plan a PR in mathcomp. One application is to substitute the ad-hoc dot product theory of cfdot in classfun by the general theory of hermitian forms that forms should bring.

amahboubi:
So in order to prepare a PR for math-comp you created a branch of math-comp?

amahboubi:
Re-asking @CohenCyril,  apart from the diff branch of analysis, is there any other library you are aware of which depends on this math-comp/forms branch?

CohenCyril:
Yes, I was recently convinced by some best practice manual, which explains that people with commit rights should prepare their pull requests in a specific branch of a repository rather than in a separate repo.
The master branch being the "official" one, the other may be development branches.

CohenCyril:
> Re-asking @CohenCyril, apart from the diff branch of analysis, is there any other library you are aware of which depends on this math-comp/forms branch?

No, not that I am aware of (and I'm pretty sure I would be aware of it if it existed)

amahboubi:
Naive question then: why not having forms being part of analysis during this transitional period?

CohenCyril:
Why not, since the branch `diff` has not been merged in `master` yet, no "hard" decision has been taken yet.

amahboubi:
[...]

PR #47, opened 2018-03-01
Analysis depends on a specific branch of math-comp...

amahboubi:


CohenCyril:
You are right! And indeed this is very specfic to Damien working branch. I guess he plans to make a PR at somepoint and we have to solve issues related to #46 before merging into master...

CohenCyril:
Making a PR on a branch that will ultimately also be a PR is not a good practice. The reason it is not a PR yet is that @drouhling did not judge it was ready yet for integration, so the best practice is to interact with @drouhling instead, and when it is a PR, we can start reviewing/discussing and amending.

I think most of the confusion here comes from the fact that I showed you a work in progress that I could see @drouhling was working on, but that even I did not have many opportunities to discuss with him yet... We are basically standing over his shoulder right now ;)

amahboubi:
Ok. I am closing the PR.

PR #54, opened 2018-03-20
Add the Intermediate Value Theorem

drouhling:


hierarchy.v
```diff
@@ -2105,6 +2105,90 @@ rewrite ltr_subl_addl -ltr_subl_addr; apply: ltr_trans ltyz.
 by apply: ltr_distW; rewrite -absRE absrB.
 Qed.
 
+Lemma ler0P' (R : realFieldType) (x : R) :
```
CohenCyril: This name is not very good. How `about ler_gt0P` ?

hierarchy.v
```diff
@@ -2105,6 +2105,90 @@ rewrite ltr_subl_addl -ltr_subl_addr; apply: ltr_trans ltyz.
...
+  reflect (forall e, e > 0 -> x <= e) (x <= 0).
+Proof.
+apply: (iffP idP) => [lex0 e egt0|lex0].
+  by apply: ler_trans lex0 _; apply: ltrW.
+case: (lerP x 0) => // lt0x.
+have /midf_lt [_] := lt0x; rewrite ltrNge -eqbF_neg => /eqP<-.
+by rewrite add0r; apply: lex0; rewrite -[x]/((PosNum lt0x)%:num).
+Qed.
+
+Lemma IVT (f : R -> R) (a b v : R) :
```
CohenCyril: I'm not sure about this min an max stuff for `a` and `b` that make things less easy to instantiate.
Maybe have `a <= b` and make sure the user can prove `minr a b <= maxr a b` in one go.

drouhling: It might be interesting to open an issue for the other occurrences (e.g. in `MVT`).

PR #55, opened 2018-03-21
lemma : (\sum_i f i) x = \sum_i f i x

affeldt-aist:
(when the symbol + is explicit, folding can be achieved using tactics)

### DIFF
--- a/landau.v
+++ b/landau.v
@@ -1,7 +1,7 @@
 (* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
 Require Import Reals.
 From Coq Require Import ssreflect ssrfun ssrbool.
-From mathcomp Require Import ssrnat eqtype choice ssralg ssrnum.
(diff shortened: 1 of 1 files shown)


landau.v
```diff
@@ -262,6 +262,13 @@ Next Obligation. by move=> f g; rewrite funeqE => x /=; rewrite scalerDl. Qed.
 Canonical fct_lmodType U (R : ringType) (V : lmodType R) :=
   LmodType _ (U -> V) (fct_lmodMixin U V).
 
+Lemma fct_sumE (T : Type) (M : zmodType) n (f : 'I_n -> T -> M) (x : T) :
+  (\sum_(i < n) f i) x = \sum_(i < n) f i x.
```
CohenCyril: Hi @affeldt-aist, is there a reason why this theorem is not stated for any bigop?

CohenCyril: Never mind

PR #58, opened 2018-03-22
complete the theory of big-Omega/Theta

affeldt-aist:
clean the formalization (uniform naming,
all formats now in Reserved Notations first,
complete the documentation).

### DIFF
--- a/landau.v
+++ b/landau.v
@@ -69,15 +69,19 @@ Require Import Rstruct Rbar set posnum topology hierarchy.
 (*      f ~~_ F g == f == g +o_ F g (i.e., as a boolean relation)             *)
 (* --> asymptotic equivalence proved to be an equivalence relation            *)
 (*                                                                            *)
-(* The last part of this file is work in progress about the big-Omega and     *)
-(* big-Theta notations:                                                       *)
-(*                                                                            *)
-(* {Omega_F f} == the type of functions that are a big Omega of f near F      *)
+(* Big-Omega and big-Theta notations on the model of bigO and littleo:        *)
+(* {Omega_F f}      == the type of functions that are a big Omega of f near F *)
+(* [bigOmega of f]  == recovers the canonical structure of big-Omega of f     *)
 (* [Omega_F e of f] == returns a function with a bigOmega canonical structure *)
 (*                     provably equal to f if f is indeed a bigOmega of e     *)
 (*                     or e otherwise                                         *)
-(* f \is `Omega_F(e) == f : T -> eV is a bigOmega of e : T -> W               *)
-(* f =Omega_F h == f is a bigOmega of h near F                                *)
+(* f \is `Omega_F(e) == f : T -> V is a bigOmega of e : T -> W near F         *)
+(* f =Omega_F h      == f : T -> V is a bigOmega of h : T -> V near F         *)
+(* --> lemmas: relation with big-O, transitivity, product of functions, etc.  *)
+(*                                                                            *)
+(* Similar notations available for big-Theta.                                 *)
+(* --> lemmas: relations with big-O and big-Omega, reflexivity, symmetry,     *)
+(*     transitivity, product of functions, etc.                               *)
 (*                                                                            *)
 (******************************************************************************)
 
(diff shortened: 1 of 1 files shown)


landau.v
```diff
@@ -1185,78 +1182,86 @@ Qed.
...
+Reserved Notation "[bigOmega 'of' f 'for' fT ]" (at level 0, f at level 0,
+  format "[bigOmega  'of'  f  'for'  fT ]").
+Reserved Notation "[bigOmega 'of' f ]" (at level 0, f at level 0,
+  format "[bigOmega  'of'  f ]").
+Reserved Notation "[Omega_ x e 'of' f ]"
+  (at level 0, x, e at level 0, only parsing).
+(* Printing *)
 Reserved Notation "[Omega '_' x e 'of' f ]"
   (at level 0, x, e at level 0, format "[Omega '_' x  e  'of'  f ]").
+Reserved Notation "`Omega_ F g"
```
drouhling: You should use a simple quote instead of a backquote.

landau.v
```diff
@@ -289,6 +293,17 @@ Definition gen_tag : unit := tt.
 Definition a_tag : unit := tt.
 Lemma showo : (gen_tag = tt) * (the_tag = tt) * (a_tag = tt). Proof. by []. Qed.
 
+Reserved Notation "{o_ F f }" (at level 0, F at level 0, format "{o_ F  f }").
```
drouhling: Shouldn't all reserved notations be put at the beginning?

landau.v
```diff
@@ -1309,33 +1311,114 @@ apply/eqOP; exists k => //; near=> x.
...
+Hint Extern 0 (prop_near1 _) => solve[apply: bigThetaP] : core.
+Hint Extern 0 (prop_near2 _) => solve[apply: bigThetaP] : core.
+
+Notation "f '=Theta_' F h" := (f%function = mkbigTheta the_tag F f h).
+
+(* properties of big Theta *)
 
-Lemma bigThetaE (F : filter_on T) (f : T -> V) (g : T -> W) :
+Lemma bigThetaE {W} (F : filter_on T) (f : T -> V) (g : T -> W) :
   f \is `Theta_F(g) <-> f =O_F g /\ f \is `Omega_F(g).
```
drouhling: Maybe use an equality here?

landau.v
```diff
@@ -1309,33 +1311,114 @@ apply/eqOP; exists k => //; near=> x.
...
+Reserved Notation "[bigTheta 'of' f 'for' fT ]" (at level 0, f at level 0,
+  format "[bigTheta  'of'  f  'for'  fT ]").
+Reserved Notation "[bigTheta 'of' f ]" (at level 0, f at level 0,
+  format "[bigTheta  'of'  f ]").
+Reserved Notation "[Theta_ x e 'of' f ]"
+  (at level 0, x, e at level 0, only parsing).
+(*Printing*)
+Reserved Notation "[Theta '_' x e 'of' f ]"
+  (at level 0, x, e at level 0, format "[Theta '_' x  e  'of'  f ]").
+Reserved Notation "`Theta_ F g"
```
drouhling: Same remark about simple quotes and reserved notations.

drouhling:
I also have a question which is unrelated to the changes brought by this PR. Why is the `_ \is _` construction used, while it is not used for `littleo` and `bigO`?

landau.v
affeldt-aist: Indeed.

landau.v
affeldt-aist: Indeed.

landau.v
affeldt-aist: OK.

affeldt-aist:
> Why is the _ \is _ construction used, while it is not used for littleo and bigO?
The usefulness of rewriting was more obvious for littleo and bigO than for big-Omega and big-Theta,
this is why we started with the simpler \is construct only. Since then, the boilerplate of littleo and bigO
has been adapted and \is remains only because it is slightly more general than =Omega_ and =Theta_
(it allows for functions with different codomains). These are not strong reasons to keep it.

landau.v
affeldt-aist: I was wondering. There are some many. At the beginning of the relevant section seems also a good place to me.

drouhling:
OK thanks for the clarification.

PR #62, opened 2018-03-23
just moved Reserved Notation's on top of landau.v and set.v

affeldt-aist:
There is sill one Reserved Notation in the middle of hierarchy.v:
Reserved Notation  "`|[ x ]|" (at level 0, x at level 99, format "`|[ x ]|").

Putting it at the top of the file reveals that it conflicts with the tactical
=> [[...]|[...]]


CohenCyril:
Yes, this conflict an issue I know, and it is supposed to be fixed by #4

PR #65, opened 2018-03-30
Import proofs about Hausdorff spaces

drouhling:


hierarchy.v
```diff
@@ -1407,6 +1442,19 @@ Lemma flim_map_lim {T : Type} {F} {FF : ProperFilter F} (f : T -> V) (l : V) :
   f @ F --> l -> lim (f @ F) = l.
 Proof. exact: flim_lim. Qed.
 
+Lemma hausdorff_normedModType : hausdorff V.
```
CohenCyril: `normedModType_hausdorff` (indeed `haussdorff` is a uniary predicate)

hierarchy.v
```diff
@@ -656,6 +659,41 @@ Proof. exact: flim_close. Qed.
 End Locally.
 Hint Resolve locally_ball.
 
+Lemma ler_addgt0P (R : realFieldType) (x y : R) :
```
CohenCyril: Maybe we should have a convention for this kind of lemma?
Like `lerP_epsilon`?

hierarchy.v
```diff
@@ -656,6 +659,41 @@ Proof. exact: flim_close. Qed.
...
+apply/(iffP idP)=> [lexy _/posnumP[e] | lexye]; first by rewrite ler_paddr.
+case: (lerP x y) => // ltyx.
+have /midf_lt [_] := ltyx; rewrite ltrNge -eqbF_neg => /eqP<-.
+suff -> : (y + x) / 2 = y + (x - y) / 2.
+  by apply/lexye/divr_gt0 => //; rewrite subr_gt0.
+by rewrite !mulrDl addrC -mulN1r -mulrA mulN1r [RHS]addrC {3}(splitr y)
+  [RHS]GRing.subrKA.
+Qed.
+
+Lemma in_segmentP (R : realFieldType) (x y z : R) :
```
CohenCyril: `in_segmentPepsilon`?

hierarchy.v
drouhling: I am not very fond of the names containing "epsilon". Using the name of a bound variable to describe a statement seems to be a bad practice IMHO.

I could use "banana" instead of "epsilon" and still have the same statement. To my great distaste, History preferred "epsilon"…

CohenCyril: I meant there is a pattern here, and `in_segmentP` does not conform. I would rather have a **documented** (almost) arbitrary name, than two lemmas with the same pattern but nothing in common in their name. If "epsilon" is not to your liking, I can understand, maybe me can find another, and that was my question: "pos", "close", "small", "around"... ?

drouhling: I understood you question but I have no "good" solution.

Maybe something in the same vein as what we chose before with `ler_gt0P`?
Say, `ler0_gt0P` for `ler_gt0P`, `ler_gt0P` for `ler_addgt0P` and `in_segment_gt0P` for `in_segmentP`?

At least the common pattern is explicit and compact and we can easily search and replace when a "good" convention is found.

CohenCyril: Ok, ... maybe for now we can pick `addgt0P` as an idiomatic suffix?

affeldt-aist: It looks like ltn_addr/ltn_addl in ssrnat and ler_addr/ler_addl in ssrnum,
so I was wondering if a "r/l" suffix was not in order?

hierarchy.v
```diff
@@ -656,6 +659,55 @@ Proof. exact: flim_close. Qed.
 End Locally.
 Hint Resolve locally_ball.
 
+Lemma ler_addrgt0P (R : realFieldType) (x y : R) :
```
CohenCyril: The convention would be to put the "r" (or "l") after the "P", as in `maxn_idPr`

PR #66, opened 2018-04-03
Import a few theorems on sets

drouhling:


topology.v
```diff
@@ -1197,6 +1197,12 @@ by rewrite openE => Aop sAB p Ap; apply: filterS sAB _; apply: Aop.
 Qed.
 Definition locally_open := @openP.
 
+Lemma open_locally (A : set T) : open A^o.
```
CohenCyril: `open_interior` instead?

set.v
```diff
@@ -320,6 +330,14 @@ Lemma setIACA {A} (X Y Z T : set A) :
   X `&` Y `&` (Z `&` T) = X `&` Z `&` (Y `&` T).
 Proof. by rewrite -setIA [Y `&` _]setICA setIA. Qed.
 
+Lemma bigII A I (D : set I) (f : I -> set A) (X : set A) :
```
CohenCyril: head symbol is setI, so `setI_bigcapl` might be a better naming.

set.v
```diff
@@ -292,6 +292,13 @@ Qed.
 Lemma subIset {A} (X Y Z : set A) : X `<=` Z \/ Y `<=` Z -> X `&` Y `<=` Z.
 Proof. case => H a; by [move=> [/H] | move=> [_ /H]]. Qed.
 
+Lemma subset_D0 A (X Y : set A) : (X `<=` Y) = (X `\` Y = set0).
```
CohenCyril: ssreflect lemma is: `setD_eq0 A B : (A :\: B == set0) = (A \subset B).` let's have the same naming and statement conventions.

set.v
```diff
@@ -307,6 +314,9 @@ Proof. by rewrite predeqE => ?; split=> [[]|]. Qed.
 Lemma setI0 {A} (X : set A) : X `&` set0 = set0.
 Proof. by rewrite predeqE => ?; split=> [[]|]. Qed.
 
+Lemma setIl0 A (X Y : set A) : X = set0 -> X `&` Y = set0.
```
CohenCyril: Why not `setIl0 {A} (Y : set A) : set0 `&` Y = set0` instead?

topology.v
drouhling: That's what I had before, but the I noticed that `interior` totally disappeared from the code.

I chose `locally` to mimic `locally_locally`, which could have been named `locally_interior`.

set.v
drouhling: For some reason I found it more convenient for the manipulation of hypotheses in a proof I had to do. Presently I am not sure any more.

I would not add your suggestion to the library. There is already `setI0` and `setIC` for this.

CohenCyril: I indeed meant `set0I {A} (Y : set A) : set0 & Y = set0` (the commuted version of `setI0`), it is convenient not to have `setIC` to do, and it follows the scheme from mathcomp to have them both (and we also have all the variants `setTI` ...)

PR #69, opened 2018-04-16
split the generic littleo_linear0 into three lemmas,

affeldt-aist:
one of them being littleo_linear0 specialized to 0

### DIFF
--- a/derive.v
+++ b/derive.v
@@ -339,13 +339,13 @@ Proof.
 rewrite /the_littleo /insubd; case: (insubP _) => // u /asboolP lino ->.
 rewrite funeqE => x; apply/eqP; case: (ler0P `|[x]|) => [|xn0].
   by rewrite normm_le0 => /eqP ->; rewrite linear0.
-rewrite -normm_le0 -(mul0r `|[x]|) -ler_pdivr_mulr //; apply/ler_gt0P => e egt0.
+rewrite -normm_le0 -(mul0r `|[x]|) -ler_pdivr_mulr //; apply/ler0_addgt0P => e egt0.
 by rewrite ler_pdivr_mulr // littleo_linear_id.
 Qed.
 
 Lemma littleo_center (V W : normedModType R) (f : V -> W) (x : V) :
-  [o_x center x of f \o center x] = [o_ (0 : V) id of f] \o center x.
-Proof. by rewrite littleo_center0 comp_centerK (comp_centerK x id). Qed.
+  [o_x center x of f] = [o_ (0 : V) id of f \o shift x] \o center x.
+Proof. by rewrite littleo_center0 -(comp_centerK x id). Qed.
 
 Lemma diff_unique (V W : normedModType R) (f : V -> W)
   (df : {linear V -> W}) x :
--- a/derive.v
+++ b/derive.v
@@ -315,51 +315,50 @@ End DifferentialR2.
 
 Section DifferentialR3.
 
-Variables (V W : normedModType R).
-
-Lemma littleo_linear0 (V' W' : normedModType R) (f : {linear V' -> W'})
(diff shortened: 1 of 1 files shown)


derive.v
```diff
@@ -315,51 +315,50 @@ End DifferentialR2.
...
-  continuous df -> f \o shift x = cst (f x) + df +o_ (0 : V') id ->
-  'd_x f = df :> (V' -> W').
+Lemma littleo_linear0 (V W : normedModType R) (f : {linear V -> W}) :
+  [o_ (0 : V) id of f] = cst 0.
 Proof.
-move=> dfc dxf; suff dfef' : 'd_x f \- df =o_ (0 : V') id.
+rewrite /the_littleo /insubd; case: (insubP _) => // u /asboolP lino ->.
+rewrite funeqE => x; apply/eqP; case: (ler0P `|[x]|) => [|xn0].
+  by rewrite normm_le0 => /eqP ->; rewrite linear0.
+rewrite -normm_le0 -(mul0r `|[x]|) -ler_pdivr_mulr //; apply/ler_gt0P => e egt0.
```
CohenCyril: According to CI, this fails, I guess it is supposed to be `ler0_addgt0P`

derive.v
```diff
@@ -315,51 +315,50 @@ End DifferentialR2.
...
+by rewrite ler_pdivr_mulr // littleo_linear_id.
+Qed.
+
+Lemma littleo_center (V W : normedModType R) (f : V -> W) (x : V) :
+  [o_x center x of f \o center x] = [o_ (0 : V) id of f] \o center x.
```
CohenCyril: Explicit `[o_x f of e]` where `e` is not a variable are not "ready to rewrite". I think it is better to put a shift in the RHS instead.

affeldt-aist: > it is supposed to be ler0_addgt0P
> it is better to put a shift in the RHS instead
Indeed (x2). Fixed.

derive.v
```diff
@@ -315,51 +315,50 @@ End DifferentialR2.
...
-move=> dfc dxf; suff dfef' : 'd_x f \- df =o_ (0 : V') id.
+rewrite /the_littleo /insubd; case: (insubP _) => // u /asboolP lino ->.
+rewrite funeqE => x; apply/eqP; case: (ler0P `|[x]|) => [|xn0].
+  by rewrite normm_le0 => /eqP ->; rewrite linear0.
+rewrite -normm_le0 -(mul0r `|[x]|) -ler_pdivr_mulr //; apply/ler0_addgt0P => e egt0.
+by rewrite ler_pdivr_mulr // littleo_linear_id.
+Qed.
+
+Lemma littleo_center (V W : normedModType R) (f : V -> W) (x : V) :
+  [o_x center x of f] = [o_ (0 : V) id of f \o shift x] \o center x.
```
CohenCyril: I do not think this lemma is necessary, it looks subsumed by `littleo_center0`

derive.v
```diff
@@ -315,51 +315,50 @@ End DifferentialR2.
...
+rewrite ger0_norm // ltr_pdivr_mulr // -mulrA mulVf; last exact: lt0r_neq0.
 by rewrite mulr1 [X in _ < X]splitr ltr_addl.
 Qed.
 
-Lemma diff_unique (V' W' : normedModType R) (f : V' -> W')
-  (df : {linear V' -> W'}) x :
-  continuous df -> f \o shift x = cst (f x) + df +o_ (0 : V') id ->
-  'd_x f = df :> (V' -> W').
+Lemma littleo_linear0 (V W : normedModType R) (f : {linear V -> W}) :
+  [o_ (0 : V) id of f] = cst 0.
```
CohenCyril: I am afraid this is difficult to use too... Maybe `f =o id -> f = cst 0` is better?

PR #78, opened 2018-04-24
differentiable x (fun y => k y *: f)

affeldt-aist:
I have a use for these lemmas in coq-robot.


CohenCyril:
The failure comes from the fact we just moved real_closed out of math-comp/ but did not modify the opam-coq-archive package yet... I'm on it...

Issue #68, opened 2018-04-10
Have notations for deriving and derivability


CohenCyril:
Provide derivation notations such as:
- ``f^`(n)``, ``f^`()`` and ``f^`N(n)`` (as in `poly.v`) for derivation on`R` (and when available, on `C`)
- `'D(A -> V)` for the type of derviable functions from domain `A` to a normed module type `V` 
- `'D^n(A -> V)` for `n` times derivable (so that `'D^1 = 'D`)
- `'C^n_(A -> V)` (I believe `'C^0 = 'C` in the literature, but I advocate for not providing a notation without explicit `n`)

affeldt-aist:
> f^`(n), f^`() and f^`N(n) (as in poly.v) for derivation onR (and when available, on C)
I gave a try to f^`() and f^`(n) in a branch issue68. I experimented with 'D_v[f] for the directional derivative also. 

> Bonus question: should the last 3 be types or sets ?
Mmmhhh. The proposed notation looks more appropriate for a type but it seems to me that the current formalization lends itself more easily to sets.

Issue #34, opened 2018-02-08
Document derive


CohenCyril:
https://github.com/math-comp/analysis/blob/abb66fa649a4c6ee061a211c4c834242418d325c/derive.v#L1

affeldt-aist:
You'll find a tentative documentation in branch issue68 (this is related with issue #68 that discusses notations).

PR #80, opened 2018-04-24
WIP: notations and documentation in derive.v

affeldt-aist:
It addresses: one part of issue #68 (the notations) and issue #34.
The notations do not seem to generate conflicts.


derive.v
```diff
@@ -12,6 +12,32 @@ Import GRing.Theory Num.Def Num.Theory.
...
 Local Open Scope classical_set_scope.
 
+(******************************************************************************)
+(* This file provides a theory of differentiation. It includes the standard   *)
+(* rules of differentiation (differential of a sum, of a product, of          *)
+(* exponentiation, of the inverse, etc.) as well standard theorems (the       *)
+(* Extreme Value Theorem, Rolle's theorem, the Mean Value Theorem).           *)
+(*                                                                            *)
+(* Parsable notations:                                                        *)
+(*               'd_x f == the differential of a function f at a point x      *)
```
CohenCyril: I changed my mind, I think I'd prefer the notation to be `'d f x h` (with the x after the f), because it corresponds better to the mathematical practice and because a differential is supposed to be a vector field. Same for `differentiable` and `'J`.
Also I do not think there should be brackets for `'J` and `'D` (and the order of arguments is good for 'D, so we should keep it)

affeldt-aist: That indeed looks more natural.

CohenCyril: Do not forget to rebase on master btw

PR #81, opened 2018-04-24
Updating dependencies and instructions

CohenCyril:
This will not compile at first (it requires a few updates in coq-opam-archive).


strub:
This is certainly a caching issue. Try to merge master into your branch to get the new CI file.

Anyway, CI caching system is broken-by-design (TM).

CohenCyril:
Things specific to CI are now moved to issue #84

PR #92, opened 2018-08-17
Use near notation in bigO

drouhling:
To be consistent with our paper on asymptotic reasoning I swapped `bigO` and `bigOF`. This PR also:
- fixes the compilation which was broken by recent changes in finmap
- fixes the `near:` tactic: in nested uses of the near tactics, hypotheses could be removed from the context of some evars when calling the `near:` tactic
- improves the automation of positivity proofs involving points `x` such that `x \is_near +oo`


CohenCyril:
Thanks @drouhling. The right fix for finmap would be to depend on the released version 1.0.0, the dev one is now aligned with 1.1.0 which deprecated seq_fset in favour lf [fset x in s]`

PR #93, opened 2018-08-20
Fix the code wrt finmap's update

drouhling:


strub:
Which version of finmap this PR is targeting?

drouhling:
This was done with the dev version but since it also works with v1.1.0 we might as well lower the deps as in #96.

altreals/realsum.v
```diff
@@ -150,12 +150,12 @@ case: (existsTP (fun s : seq T => {subset E i <= s}))=> /= [[s le_Eis]|].
...
 pose K := (`|ifloor M|.+1 * i.+1)%N; move/(_ K)/existsp_asboolP/existsbP.
 move=> h; have /asboolP[] := xchooseP h.
-set s := xchoose h=> eq_si uq_s le_sEi; pose J := seq_fset s.
+set s := xchoose h=> eq_si uq_s le_sEi; pose J := [fset x in s].
 suff: \sum_(x : J) `|f (val x)| > M by rewrite ltrNge bM.
 apply/(@ltr_le_trans _ (\sum_(x : J) 1 / i.+1%:~R)); last first.
   apply/ler_sum=> /= m _; apply/ltrW; suff: (val m \in E i) by apply.
-  by apply/le_sEi; rewrite -seq_fsetE; apply/fsvalP.
-rewrite sumr_const cardfsE undup_id // eq_si -mulr_natr -pmulrn.
+  by apply/le_sEi; rewrite -(in_fset imfset_key); apply/fsvalP.
```
CohenCyril: I am a bit suspicious about this, why would there be an explicit key here?

hierarchy.v
```diff
@@ -2678,15 +2678,15 @@ move=> C D FC f_D; have {f_D} f_D :
     rewrite predeqE => g; split; first by move=> [_ [_ _ <-]].
     move=> Pg; exists [set g | forall j, get (Pj j) (g j)] => //.
     by exists ord0.
-  move=> _ [_ _ <-].
-  exists (seq_fset [seq (@^~ j) @^-1` (get (Pj j)) | j : 'I_n.+1]).
-    move=> B'; rewrite seq_fsetE => /mapP [j _ ->]; rewrite inE.
+  move=> _ [_ _ <-]; set s := [seq (@^~ j) @^-1` (get (Pj j)) | j : 'I_n.+1].
+  exists [fset x in s]%fset.
```
CohenCyril: Is the `%fset` scope annotation mandatory? If so, would do you know why?

altreals/realsum.v
drouhling: I tried to follow the original proof: the replacement for `seq_fsetE` is `in_fset` but rewriting `in_fset` from right to left means providing a key. I'm reworking the proof to remove the key.

Issue #94, opened 2018-09-03
opam package does not build with latest mathcomp-finmap


Armael:
Trying to install the library using the opam package currently does not work, because of finmap it seems (it works if I pin finmap 1.0.0 instead of using finmap dev).

The log:
```
make -f Makefile.coq
make[1] : on entre dans le répertoire « /home/armael/.opam/current/.opam-switch/build/coq-mathcomp-analysis.dev »
COQDEP VFILES
COQC altreals/xfinmap.v
COQC boolp.v
COQC forms.v
COQC reals.v
COQC classical_sets.v
File "./altreals/xfinmap.v", line 83, characters 30-31:
Error:
In environment
R : Type
idx : R
op : Monoid.com_law 1
T : choiceType
s : seq T
P : ?T
F : ?T0
The term "s" has type "seq T" while it is expected to have type "unit".

make[2]: *** [Makefile.coq:657: altreals/xfinmap.vo] Error 1
```

affeldt-aist:
> Trying to install the library using the opam package currently does not work,
> because of finmap it seems (it works if I pin finmap 1.0.0 instead of using finmap dev).

Indeed the version in master at the time of this writing is using finmap 1.0 and
README.md does not reflect this. This compatibility issue is due to recent changes
in finmap.

CohenCyril:
Closed by #96

PR #96, opened 2018-09-03
update INSTALL.md 

affeldt-aist:
to reflect the version changes of some of the libraries used in mathcomp-analysis in an attempt to fix issue #94

### DIFF
--- a/opam
+++ b/opam
@@ -21,5 +21,5 @@ depends: [
   "coq-mathcomp-ssreflect"   {= "1.7.0"}
   "coq-mathcomp-algebra"     {= "1.7.0"}
   "coq-mathcomp-real-closed" {= "dev"}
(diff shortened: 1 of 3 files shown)


opam
```diff
@@ -21,5 +21,5 @@ depends: [
   "coq-mathcomp-ssreflect"   {= "1.7.0"}
   "coq-mathcomp-algebra"     {= "1.7.0"}
   "coq-mathcomp-real-closed" {= "dev"}
```
CohenCyril: If we really do not depend on real-closed but jsut big-enough, then this could theoretically be replaced too...

affeldt-aist: The commit 7fe9511 of branch issue_94 does not mention real-closed anymore.

PR #90, opened 2018-07-09
WIP: Simplify filters

CohenCyril:
- better organization of locally, near etc
- typeclasses to cannoncial structures

PR #98, opened 2018-09-10
Add uniform bigO from ANR FOST as an example

drouhling:


examples/misc.v
```diff
@@ -0,0 +1,123 @@
...
+move=> x dx ltdxgetQ; apply: fOg; apply: ler_lt_trans ltdxgetQ.
+by have /andP [] := ler_norm2 dx.
+Qed.
+
+(* then we replace the epsilon/delta definition with bigO *)
+
+Local Open Scope classical_set_scope.
+
+Canonical prod_filter_on T U (F : filter_on T) (G : filter_on U) :=
+  FilterType (filter_prod F G) (filter_prod_filter _ _).
```
CohenCyril: This looks like it should be backported to `topology.v`

examples/misc.v
```diff
@@ -0,0 +1,123 @@
...
+
+Program Definition trivial_filter_on T := FilterType [set setT : set T] _.
+Next Obligation.
+split=> // [_ _ -> ->|Q R sQR QT]; first by rewrite setIT.
+by apply: eqEsubset => // ? _; apply/sQR; rewrite QT.
+Qed.
+Canonical trivial_filter_on.
```
CohenCyril: same here

examples/misc.v
```diff
@@ -0,0 +1,123 @@
...
+
+Canonical within_filter_on T D (F : filter_on T) :=
+  FilterType (within D F) (within_filter _ _).
```
CohenCyril: same here

drouhling:
Indeed. I thought this was rather the purpose of #90 but since this example file is not compiled with make there is a chance we forget about it.

I'll move these pieces of code and let you resolve the conflicts with #90 :-)

CohenCyril:
Also I rather would name the directory `misc` and the file `uniform_bigO`.

PR #101, opened 2018-09-13
Remove definitions in landau

drouhling:
This PR:
- removes `bigOW`, which has no use: use `bigO` for proofs and switch to `bigOF` when using a `bigO` hypothesis
- removes the `littleo`/`bigO`/`bigOmega`/`bigTheta` definitions: the user should always use the equational notations and use the views `eqoP`, `eqaddOP`, etc, to go back to filter reasoning when needed.

There is something I don't truly understand: I had to add occurrences of `is_true` in the definitions of `littleo`, `bigO`, etc, in order for the hints related to `littleoP`, `bigOP`, etc, to work.

### DIFF
--- a/landau.v
+++ b/landau.v
@@ -547,35 +547,17 @@ Proof. by move: x; rewrite -/(_ *: _ =1 _) {1}scaleo. Qed.
 Definition bigO (F : set (set T)) (f : T -> V) (g : T -> W) :=
   \forall k \near +oo, \forall x \near F, `|[f x]| <= k * `|[g x]|.
 
-Definition bigOW (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  exists k, \forall x \near F, `|[f x]| <= k * `|[g x]|.
-
-Lemma bigOWE (F : set (set T)) : Filter F -> bigOW F = bigO F.
-Proof.
-rewrite predeq2E => FF f g; split=> [[k] |] kP; last first.
-  by near +oo => k; exists k; near: k.
-near=> k'; near=> x; rewrite (ler_trans (near kP _ _)) // ler_wpmul2r //.
-by rewrite ltrW //; near: k'; exists k.
-Grab Existential Variables. all: end_near. Qed.
-
 Definition bigOF (F : set (set T)) (f : T -> V) (g : T -> W) :=
   exists2 k, k > 0 & \forall x \near F, `|[f x]| <= k * `|[g x]|.
 
-Lemma bigOWFE (F : set (set T)) : Filter F -> bigOW F = bigOF F.
-Proof.
-rewrite predeq2E => f g; split=> [[k] | [k k_gt0]] kP; last by exists k.
-exists (maxr k 1); first by rewrite ltr_maxr ltr01 orbT.
-by apply: filterS kP => x /ler_trans; apply; rewrite ler_wpmul2r // ler_maxr lerr.
-Qed.
-
 Lemma bigOFE (F : set (set T)) : Filter F -> bigOF F = bigO F.
-Proof. by move=> FF; rewrite -bigOWE bigOWFE. Qed.
-
-Lemma bigOWP (F : set (set T)) f g : Filter F -> bigOW F f g -> bigO F f g.
-Proof. by move=> /bigOWE->. Qed.
-
-Lemma bigOWI (F : set (set T)) f g : Filter F -> bigO F f g -> bigOW F f g.
-Proof. by move=> /bigOWE->. Qed.
+Proof.
+rewrite predeq2E => FF f g; split=> [[k _ fOg] | [k fOg]].
+  exists k => l ltkl; move: fOg; apply: filter_app; near=> x.
+  by move/ler_trans; apply; rewrite ler_wpmul2r // ltrW.
+exists (maxr 1 (k + 1)); first by rewrite ltr_maxr ltr01.
+by apply: fOg; rewrite ltr_maxr orbC ltr_addl ltr01.
+Unshelve. end_near. Qed.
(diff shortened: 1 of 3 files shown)


landau.v
```diff
@@ -346,8 +348,9 @@ Section Domination.
 
 Context {K : absRingType} {T : Type} {V W : normedModType K}.
 
-Definition littleo (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  forall eps : R, 0 < eps -> \forall x \near F, `|[f x]| <= eps * `|[g x]|.
+Notation littleo F f g :=
```
CohenCyril: `Local` notation please

landau.v
```diff
@@ -544,44 +547,21 @@ Lemma scaleox (F : filter_on T) a (f : T -> V) e x :
   a *: ([o_F e of f] x) = [o_F e of a *: [o_F e of f]] x.
 Proof. by move: x; rewrite -/(_ *: _ =1 _) {1}scaleo. Qed.
 
-Definition bigO (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  \forall k \near +oo, \forall x \near F, `|[f x]| <= k * `|[g x]|.
+Notation bigO F f g :=
```
CohenCyril: `Local`

landau.v
```diff
@@ -544,44 +547,21 @@ Lemma scaleox (F : filter_on T) a (f : T -> V) e x :
...
+  (\forall k \near +oo, \forall x \near F, is_true (`|[f x]| <= k * `|[g x]|)).
 
-Definition bigOW (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  exists k, \forall x \near F, `|[f x]| <= k * `|[g x]|.
+Notation bigOF F f g :=
```
CohenCyril: `Local`

landau.v
```diff
@@ -887,15 +855,13 @@ Hint Extern 0 (_ = 'O__ _) => apply: eqOE; reflexivity : core.
 Hint Extern 0 (_ = 'O__ _) => apply: eqoO; reflexivity : core.
 Hint Extern 0 (_ = _ + 'o__ _) => apply: eqaddoE; reflexivity : core.
 Hint Extern 0 (_ = _ + 'O__ _) => apply: eqaddOE; reflexivity : core.
-Hint Extern 0 (bigO _ _ _) => solve[apply: bigOP] : core.
+Hint Extern 0 (\forall k \near +oo, \forall x \near _,
+  is_true (`|[_ x]| <= k * `|[_ x]|)) => solve[apply: littleoP] : core.
```
CohenCyril: `bigO`

landau.v
CohenCyril: rename `bigO_ex_def` please

landau.v
CohenCyril: Name `littlleo_def`

landau.v
CohenCyril: `bigO_def`

landau.v
CohenCyril: And `Lemma bigO_ex (F : set (set T)) (g : T -> W) (f : {O_F g}) : bigO_ex_def F f g.` is missing, right?

landau.v
```diff
@@ -1227,8 +1196,9 @@ Section big_omega.
 Context {K : absRingType} {T : Type} {V : normedModType K}.
 Implicit Types W : normedModType K.
 
-Definition bigOmega {W} (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  exists2 k, k > 0 & \forall x \near F, `|[f x]| >= k * `|[g x]|.
+Notation bigOmega F f g :=
+  (exists2 k, is_true (k > 0) &
```
CohenCyril: Why an explicit `is_true`?

landau.v
```diff
@@ -1357,9 +1327,10 @@ Section big_theta.
 Context {K : absRingType} {T : Type} {V : normedModType K}.
 Implicit Types W : normedModType K.
 
-Definition bigTheta {W} (F : set (set T)) (f : T -> V) (g : T -> W) :=
-  exists2 k, ((k.1 > 0) && (k.2 > 0)) &
-    \forall x \near F, k.1 * `|[g x]| <= `|[f x]| /\ `|[f x]| <= k.2 * `|[g x]|.
+Notation bigTheta F f g :=
+  (exists2 k, is_true ((k.1 > 0) && (k.2 > 0)) &
+    \forall x \near F, is_true (k.1 * `|[g x]| <= `|[f x]|) /\
```
CohenCyril: Why explicit `is_true`?

CohenCyril:
> There is something I don't truly understand: I had to add occurrences of `is_true` in the definitions of `littleo`, `bigO`, etc, in order for the hints related to `littleoP`, `bigOP`, etc, to work

In the hint partterns, it is okay, because since any term can be provided, the system has no way to know it should be coerced to `Prop`. For the uses I made comments about, I do not understand the reason...

drouhling:
Notations are local to sections, so the `Local` keyword is not mandatory here. I can still add it for safety.

CohenCyril:
> Notations are local to sections, so the `Local` keyword is not mandatory here. I can still add it for safety.

sure :)

drouhling:
> In the hint patterns, it is mandatory to explicitly add `is_true` at the head, because since any term can be provided, the system has no way to know it should be coerced to `Prop`. For the uses I made comments about, I do not understand the reason...

I put those `is_true` because otherwise the hints inside the sections, which use the notations as in
```Hint Extern 0 (bigOmega _ _ _) => solve[apply: bigOmegaP] : core.```
won't work.

PR #107, opened 2018-09-28
"Spec theorems" for littleo, bigO, bigOmega and bigTheta

CohenCyril:


CohenCyril:
@drouhling and @affeldt-aist I changed my mind again, I'd like to keep `littleo` and `bigO` for the spec lemmas as in this commit. Indeed it is a shame to "reserve" them for local definition... What do you think?

CohenCyril:
> The result depends on how the goal is parsed, since the first occurrence is matched, and the proof script is harder to read in my opinion.

I agree, but it is generally the case with spec lemmas (`leqP`, `ltngtP`, `altP eqP`, ...) we only add arguments when there is an ambiguity. Here in most case, there are exactly as many bigOs as there are `have :=` and hence I am fine with it.

drouhling:
> > The result depends on how the goal is parsed, since the first occurrence is matched, and the proof script is harder to read in my opinion.
> 
> I agree, but it is generally the case with spec lemmas (`leqP`, `ltngtP`, `altP eqP`, ...) we only add arguments when there is an ambiguity. Here in most (all?) cases, there are exactly as many bigOs as there are `have :=` and hence I am fine with it.

I'm also fine with it, then.

You should rebase on master: there is a conflict between `Let littleo ...` from master and `Lemma littleo ...` from this branch.

PR #114, opened 2018-10-05
tentative definition of bigmaxr with bigop

affeldt-aist:
rfc

### DIFF
--- a/theories/Rstruct.v
+++ b/theories/Rstruct.v
@@ -496,65 +496,79 @@ Section bigmaxr.
 Context {R : realDomainType}.
 
 (* bigop pour le max pour des listes non vides ? *)
-Definition bigmaxr (x0 : R) lr :=
-  foldr Num.max (head x0 lr) (behead lr).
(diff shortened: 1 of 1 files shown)


drouhling:
I'll merge `bigmaxr` in `entourages`, since it is sufficient for the purpose of #112.

However it would be nice to extend it with something similar to `bigmaxr_mem` (replacing `head r s` with any default value `x` smaller than all the elements of the family), maybe reusing parts of your work. But this should not be in `Rstruct.v`, since this has nothing to do with `R`.

affeldt-aist:
A replacement for `bigmaxr_mem` (as mentioned by @drouhling ) has already been merged into `master` (in `normedtype.v`). The commit of this PR is about `Rstruct.v`, at a time when we were still relying on it. We can forget about it. We can also merge it to keep `Rstruct.v` as up-to-date as possible (I know at least one project that is using `Rstruct.v` through mathcomp-analysis).

PR #117, opened 2018-10-26
WIP Clean-up of lower classical layers

amahboubi:
- [ ]  ~~Obtain axioms from the std library~~ Keep the explicit list of additional axioms, and derive their standard consequences.
- [x] Remove ugly notation for `asbool`
- [x] Minimize the use of bool, and restrict it to plumbing with mathcomp. In particular, add better support for Prop predicates.
- [ ] Reorganize and complete the corpus of lemmas, in particular rewrite rules. 
 - [ ] Collect lemmas presently in other libraries, e.g. [this ones in hierarchy](https://github.com/amahboubi/analysis/blob/e135998d7130224228748cdd004051d91431985c/hierarchy.v#L612)
- [ ] Clean up the proof of Zorn-related lemmas
- [x] Move the generic choice mixin from `set.v` to boolp.

Several items in this list were already addressed or in progress [elsewhere](https://github.com/amahboubi/analysis/tree/clean-boolp), so the present branch will incorporate the relevant material.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -2,10 +2,12 @@
 -arg -parsing
 -arg -w -arg -duplicate-clear
 
+
 theories/boolp.v
 theories/reals.v
 theories/posnum.v
 theories/landau.v
+theories/classical_preds.v
 theories/classical_sets.v
 theories/Rstruct.v
 theories/Rbar.v
--- a/theories/Rstruct.v
+++ b/theories/Rstruct.v
@@ -363,47 +363,47 @@ Canonical R_rcfType := RcfType R Rreal_closed_axiom.
 End ssreal_struct.
 
 Open Scope ring_scope.
-Require Import reals boolp.
+Require Import reals boolp classical_preds classical_sets.
 
 Section ssreal_struct_contd.
 
-Lemma is_upper_boundE (E : pred R) x : is_upper_bound E x = (x \in ub E).
-Proof.
-rewrite /is_upper_bound inE forallbE asboolE /=; apply/eq_forall=> y.
-by rewrite -(reflect_eq implyP) (reflect_eq (RleP _ _)).
-Qed.
+Lemma is_upper_boundE (E : set R) x : is_upper_bound E x = (x \in ub E).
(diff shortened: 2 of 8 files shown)


amahboubi:
@CohenCyril , @strub : Require Import from std lib or 3 axiom prelude? I vote for the first.

amahboubi:
@CohenCyril corollary question: if we import axioms from std lib, is it important to keep your proof script for Diaconescu or shall we use the one from ClassicalEpsilon and have only one-liners here?

CohenCyril:
> Require Import from std lib or 3 axiom prelude? I vote for the first.

I vote for the second, because it makes it clear which axioms we use exactly. Historically I opted for importing from stdlib, and I changed just before the Coq Workshop presentation, when I realized the previous instantiation of that I did `mclassic` did not rely on the axioms I thought I used...

Additional argument: when we switch to stdlib2, our independence from stdlib1 will be in our favor. :smile:

CohenCyril:
@amahboubi I did not realize immediately that I depended on *proof irrelevance* rather than *diaconescu theorem + the axioms we already had.*

Sure, it is an easy dependency to remove indeed, but why restoring it and wait rather than keeping as few dependencies as possible? Especially since it consists in less than 20 loc anyway. We can keep a reference to the stdlib in comments.

BTW, Damien is currently chasing the dependency in stdlib R if I am correct.

amahboubi:
May be for a communication issue: it is the first time I read that analysis wants to keep as few dependencies as possible on stlib1. 
I am not yet 100% convinced and I do not believe in references by comments. But this specific issue is rather mundane and I think it does not worth spending much energy on it. My main ultimate goal here is to recover from the massive de-synchronization, in an efficient way.
I was not thinking of the dependency on R, although it is really great that Damien is addressing this. E.g. mathcomp is requiring a few libraries from stdlib1.

CohenCyril:
@amahboubi I did not say that analysis (as a project) wants anything. I think we should wait for the meeting to decide such strategies for this repo.
I was merely giving you my motivations for my "vote", I am sorry if that was unclear.

And wrt the dependency of mathcomp in stdlib1, this a distinct issue, which Maxime is already adressing.

PR #119, opened 2018-11-05
Replace balls with entourages

drouhling:
As explained in #112, this is a first step to remove the dependency on `R` and this makes it possible to have a model of reals using rational Cauchy filters.

NB: `hierarchy.v` is now called `normedtype.v` and only contains norm-related structures.


drouhling:
This now depends on #132.

drouhling:
This is now superseded by #242.

PR #122, opened 2018-12-10
replacing strict inequality by a large one in locally_Rbar

CohenCyril:


affeldt-aist:
What it is the status of this PR?

CohenCyril:
@affeldt-aist forgotten but I think would be a great improvment

CohenCyril:
Actually, I always think I merged it and often am disappointed I didn't :-(

affeldt-aist:
So I guess we should update the PR. Since things have changed in two years, I wonder whether it would not be better to do a new PR.

affeldt-aist:
We can maybe turn this PR into a draft PR since we have this most recent one: https://github.com/math-comp/analysis/pull/487

PR #124, opened 2018-12-18
more lemmas about classical sets

affeldt-aist:


classical_sets.v
```diff
@@ -240,9 +240,32 @@ Notation "A !=set0" := (nonempty A) : classical_set_scope.
...
+rewrite existsbE => /forallp_asboolPn H.
+move/negP : X_neq0; apply; apply/eqP; rewrite funeqE => t; rewrite propeqE.
+move: (H t); by rewrite asboolE.
+Qed.
+
 Lemma imageP {A B} (f : A -> B) (X : set A) a : X a -> (f @` X) (f a).
 Proof. by exists a. Qed.
 
+Lemma imsetP T1 T2 (D : set T1) (f : T1 -> T2) b :
+  reflect (exists2 a, a \in D & b = f a) (b \in f @` D).
```
drouhling: As we discussed yesterday, @amahboubi is removing the use of mathcomp's boolean `_ \in _` notation.

If you remove those from this statement, this lemma is almost void: the `@` notation already unfolds to the left part, up to the symmetry of equality.

I agree however that the `_ \in _` notation makes the statements closer to pen-and-paper ones.

@amahboubi, you are introducing a propositional `_ \in _` notation in #117, aren't you? Is it meant to be used everywhere where membership appears?

classical_sets.v
```diff
@@ -287,6 +310,12 @@ Proof. by move=> sXY ? nYa ?; apply/nYa/sXY. Qed.
 Lemma subsetU {A} (X Y Z : set A) : X `<=` Z -> Y `<=` Z -> X `|` Y `<=` Z.
 Proof. by move=> sXZ sYZ a; apply: or_ind; [apply: sXZ|apply: sYZ]. Qed.
 
+Lemma subUset T (X Y Z : set T) : (Y `|` Z `<=` X) = ((Y `<=` X) /\ (Z `<=` X)).
+Proof.
+rewrite propeqE; split => [H|H]; first by split => x Hx; apply H; [left|right].
```
drouhling: `H` is not a very informative name. Why not `sYZ_X`?

`split=> [H|H]` is the same as `split=> H`. The second `H` could be broken in two parts however in order to avoid the use of `proj1` and `proj2` on next line.

classical_sets.v
```diff
@@ -342,6 +371,22 @@ Proof. move=> p; rewrite /setU predeqE => a; tauto. Qed.
 Lemma setUC A : commutative (@setU A).
 Proof. move=> p q; rewrite /setU predeqE => a; tauto. Qed.
 
+Lemma in_setU T (X Y : set T) x : (x \in X `|` Y) = (x \in X) || (x \in Y) :> Prop.
```
drouhling: The same remark about `_ \in _` applies here.

This statement is also almost void: it just replaces the propositional or with the boolean one.

classical_sets.v
```diff
@@ -342,6 +371,22 @@ Proof. move=> p; rewrite /setU predeqE => a; tauto. Qed.
...
+Proof.
+rewrite propeqE; split => [ | ]; last first.
+  move/orP => -[]; rewrite 2!in_setE => ?; by [left|right].
+rewrite in_setE => -[?|?]; apply/orP; rewrite 2!in_setE; tauto.
+Qed.
+
+Lemma set0U T (X : set T) : set0 `|` X = X.
+Proof. rewrite funeqE => t; rewrite propeqE; split; by [case|right]. Qed.
```
drouhling: `by` should go right after `Proof.`

classical_sets.v
```diff
@@ -342,6 +371,22 @@ Proof. move=> p; rewrite /setU predeqE => a; tauto. Qed.
...
+
+Lemma setU0 T (X : set T) : X `|` set0 = X.
+Proof. rewrite funeqE => t; rewrite propeqE; split; by [case|left]. Qed.
```
drouhling: Same here.

classical_sets.v
amahboubi: HI @drouhling and @affeldt-aist. Yes, @drouhling , you are right. But I have nothing against merging such lemmas while I am working: if they are used for real in the library, this provides additional benchmarks.

classical_sets.v
```diff
@@ -240,9 +240,32 @@ Notation "A !=set0" := (nonempty A) : classical_set_scope.
 Lemma eqEsubset T (F G : set T) : F `<=` G -> G `<=` F -> F = G.
 Proof. by move=> H K; rewrite funeqE=> s; rewrite propeqE; split=> [/H|/K]. Qed.
 
+Lemma sub0set T (X : set T) : set0 `<=` X.
+Proof. by []. Qed.
+
+Lemma subset0 T (X : set T) : (X `<=` set0) = (X = set0).
+Proof. rewrite propeqE; split => [?|-> //]; exact/eqEsubset. Qed.
+
+Lemma set0P T (X : set T) : (X != set0) <-> (X !=set0).
```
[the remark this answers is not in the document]
affeldt-aist: applied

classical_sets.v
affeldt-aist: I was insisting on using \in so as to get more readable statements but I will do without for a while waiting for a better \in notation. I therefore remove the offending lemmas.

classical_sets.v
affeldt-aist: applied

classical_sets.v
affeldt-aist: applied (see comment above about the \in notation)

classical_sets.v
affeldt-aist: applied

classical_sets.v
affeldt-aist: applied

CohenCyril:
Unless I am mistaken @amahboubi has a lock on this file for now. If I am right, @amahboubi can you PR your changes even if they are not complete?

PR #135, opened 2019-04-09
Porting to mathcomp-1.8.0

affeldt-aist:
just so that compilation using mathcomp-1.8.0 goes through, using the branch/mathcomp-1.8.0

### DIFF
--- a/derive.v
+++ b/derive.v
@@ -1290,7 +1290,7 @@ Proof.
 move=> leab fcont; set imf := [pred t | t \in f @` [set x | x \in `[a, b]]].
 have imf_sup : has_sup imf.
   apply/has_supP; split.
-    by exists (f a); rewrite !inE; apply/asboolP/imageP; rewrite inE lerr.
+    by exists (f a); rewrite !inE; apply/asboolP/imageP; rewrite inE /= lerr.
   have [M imfltM] : bounded (f @` [set x | x \in `[a, b]] : set R^o).
     apply/compact_bounded/continuous_compact; last exact: segment_compact.
     by move=> ?; rewrite inE => /asboolP /fcont.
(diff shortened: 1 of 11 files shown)


normedtype.v
```diff
@@ -412,7 +412,7 @@ Proof. by []. Qed.
 Lemma Rhausdorff : hausdorff [topologicalType of R].
 Proof.
 move=> x y clxy.
-apply/eqP; rewrite eqr_le; apply/in_segment_addgt0Pr => _ /posnumP[e].
+apply/eqP; rewrite eqr_le; change (y \in `[x, x]); apply/in_segment_addgt0Pr => _ /posnumP[e].
```
CohenCyril: @affeldt-aist please use `rewrite -/(y \in `[x, x]) instead`

derive.v
```diff
@@ -1290,7 +1290,7 @@ Proof.
 move=> leab fcont; set imf := [pred t | t \in f @` [set x | x \in `[a, b]]].
 have imf_sup : has_sup imf.
   apply/has_supP; split.
-    by exists (f a); rewrite !inE; apply/asboolP/imageP; rewrite inE lerr.
+    by exists (f a); rewrite !inE; apply/asboolP/imageP; rewrite inE /= lerr.
```
CohenCyril: please use `inE/=` everywhere (no space needed)

normedtype.v
affeldt-aist: I needed to be more precise (i.e., rewrite -[_ && _]/(y \in `[x, x])).

derive.v
affeldt-aist: Done.

normedtype.v
CohenCyril: I changed my mind and preferred explicit arguments for `in_segment_addgt0`, I will just undo your commit if you do not mind?

PR #139, opened 2019-05-16
move structures about functions from landau.v to topology.v

### LINKED ISSUE #138: Move structures about functions out of landau.v
move

https://github.com/math-comp/analysis/blob/8768257807bbded62df52b4a97521007955cf70f/landau.v#L285-L331

at the top of topology.v?

affeldt-aist:
fixes #138


CohenCyril:
I am not sure this is the best place yet, but at least it is better than landau

PR #144, opened 2019-05-30
test nix installation on Debian

affeldt-aist:
I have not tested the temporary installation of emacs and proof general (the last point of step 4).

### DIFF
--- a/NIX.md
+++ b/NIX.md
@@ -14,23 +14,10 @@
     section of the nix
     manual](https://nixos.org/nix/manual/#ch-installing-binary).
 
-    You may experience installation errors at this stage.
-    For example, the installation may stop with the following
-    error message:
+    See [Troubleshooting](#error-when-installing-nix) in case of error.
 
-    > ...
-    > installing 'nix-2.2.2'
-    > error: cloning builder process: Operation not permitted
-    > error: unable to start build process
-    > ...
-
-    On Debian, this can be solved by the following command:
-
-    > sudo sysctl kernel.unprivileged_userns_clone=1
-
-    (Checked with Debian 9.9.)
-
-  - You need to set several environment variables before you proceed to step 2:
+  - You need to set several environment variables before you proceed to step 2.
+    As nix invites you to do, you can source the nix.sh file:
 
     > . /home/username/.nix-profile/etc/profile.d/nix.sh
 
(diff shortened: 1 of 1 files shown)


NIX.md
```diff
@@ -14,11 +14,31 @@
...
+    > error: unable to start build process
+    > ...
+
+    On Debian, this can be solved by the following command:
+
+    > sudo sysctl kernel.unprivileged_userns_clone=1
+
+    (Checked with Debian 9.9.)
+
+  - You need to set several environment variables before you proceed to step 2:
```
CohenCyril: I wanted to spare the user from a temporary procedure, which would not replace a "log out and in again" anyway, so I'd rather remove this line.

NIX.md
```diff
@@ -33,12 +53,28 @@
 
    and do whatever you are accustomed to do with Coq.
 
+   This is providing that the environment variable COQBIN does not
+   point to conflicting binaries. Check with:
+
+   > echo $COQBIN
+
+   If wrongly set, do:
```
CohenCyril:
```suggestion
If set, do:
> export COQBIN=$(which coqtop)
```

CohenCyril: Also, let's put it in a trouble shooting section.

NIX.md
```diff
@@ -14,11 +14,31 @@
     section of the nix
     manual](https://nixos.org/nix/manual/#ch-installing-binary).
 
-  - You need to **log out of your desktop session and log in again** before you proceed to step 2.
+    You may experience installation errors at this stage.
```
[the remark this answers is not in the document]
amahboubi: Or may be a later section in the same file? At least for the time being, I think it's best to have nix-related stuff in a same file.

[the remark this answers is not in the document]
CohenCyril: ok then, @affeldt-aist  could you use a markdown hyperlink to a subsequent section in the same file ?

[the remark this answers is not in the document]
affeldt-aist: DONE

NIX.md
```diff
@@ -14,23 +14,10 @@
...
-
-    On Debian, this can be solved by the following command:
-
-    > sudo sysctl kernel.unprivileged_userns_clone=1
-
-    (Checked with Debian 9.9.)
-
-  - You need to set several environment variables before you proceed to step 2:
+  - You need to set several environment variables before you proceed to step 2.
+    As nix invites you to do, you can source the nix.sh file:
```
CohenCyril: I'd rather drop this... and invite the user to always log out and in again.

NIX.md
affeldt-aist: On the other hand, the result of the previous command does invite the user to source the nix.sh file and indeed if you do so log out/log in does not seem necessary.

CohenCyril: it is necessary everytime you change terminal, so the only permanent solution is to log out and in

NIX.md
affeldt-aist: DONE

PR #147, opened 2019-07-02
upgrading nixpkgs

CohenCyril:


amahboubi:
What is the upgrade about?

CohenCyril:
This is the equivalent of `opam upgrade` except it does so for all users. Be prepared to recompile mathcomp if we merge this.

amahboubi:
How do we know the nature of the upgrade (from which version to which other)?

CohenCyril:
> How do we know the nature of the upgrade (from which version to which other)?

I know no other way than running nix-shell and observing the computed set of packages to install. (Note that this does not uninstall the previous versions, which remain available if you change the commit back to the previous one)

amahboubi:
I meant, for the present commit log, and for analysis, which version of the dependencies are supported by this new config. E.g. real-closed is now available, etc.

CohenCyril:
> I meant, for the present commit log, and for analysis, which version of the dependencies are supported by this new config. E.g. real-closed is now available, etc.

cf https://github.com/NixOS/nixpkgs/blob/73392e79aa62e406683d6a732eb4f4101f4732be/pkgs/development/coq-modules/mathcomp/extra.nix#L46-L61

amahboubi:
Could you rephrase your commit log to reflect the upgrades it allows?

amahboubi:
> cf https://github.com/NixOS/nixpkgs/blob/73392e79aa62e406683d6a732eb4f4101f4732be/pkgs/development/coq-modules/mathcomp/extra.nix#L46-L61

Could we have a nix upgrade documentation somewhere? E..g. what is the role of this file? Who should edit it and when?

CohenCyril:
> Could we have a nix upgrade documentation somewhere? E..g. what is the role of this file? Who should edit it and when?

a doc on
- upgrading the local nix shell (shell.nix) on analysis?
- or upgrading the package definitions (mathcomp/extra.nix) on nixpkgs?

amahboubi:
Both.

amahboubi:
@CohenCyril , again a description would have been really helpful here, listing the packages and/or versions made available via this update.
Also, such a PR should not say un-merged that long (without a reason).

PR #149, opened 2019-07-08
remove: line not necessary with opam 2; loosen up restriction about mathcomp's version

affeldt-aist:
This the same PR as #146 but recreated to trigger the CI.
Closes #150 and closes #148.


affeldt-aist:
I updated INSTALL.md according to the opam file and following #150.

PR #162, opened 2019-11-09
Add preimage_setI/U lemmas.

vlj:
This adds a lemma that prove that the intersection/union of preimages is the preimage of intersections/unions

### DIFF
--- a/theories/classical_sets.v
+++ b/theories/classical_sets.v
@@ -279,6 +279,14 @@ move=> Xx; have : setT x by [].
 by rewrite -fsurj => - [y _ fy_eqx]; exists y => //; rewrite /preimage fy_eqx.
 Qed.
 
+Lemma preimage_setU {A B I} (P : set I) (f:A->B) F:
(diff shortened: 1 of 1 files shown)


theories/classical_sets.v
```diff
@@ -279,6 +279,14 @@ move=> Xx; have : setT x by [].
 by rewrite -fsurj => - [y _ fy_eqx]; exists y => //; rewrite /preimage fy_eqx.
 Qed.
 
+Lemma preimage_setU {A B I} (P : set I) (f:A->B) F:
+  f @^-1` (\bigcup_ (i in P) F) = \bigcup_(i in P) (f @^-1` F).
+Proof. by rewrite predeqE; split; rewrite /bigsetU /preimage. Qed.
```
CohenCyril: Hi! Blind guess `by rewrite predeqE.` would close the goal...

vlj: you're right, thanks for the tip !

PR #163, opened 2019-12-03
Fix notation and scope usage and generalize ereals

pi8027:
cc: @affeldt-aist

### DIFF
--- a/theories/Rstruct.v
+++ b/theories/Rstruct.v
@@ -372,7 +372,7 @@ Canonical R_rcfType := RcfType R Rreal_closed_axiom.
 
 End ssreal_struct.
 
-Open Scope ring_scope.
+Local Open Scope ring_scope.
 Require Import reals boolp.
 
 Section ssreal_struct_contd.
--- a/theories/altreals/distr.v
+++ b/theories/altreals/distr.v
@@ -627,8 +627,8 @@ CoInductive dlim_spec f (x : T) : R -> Type :=
 Lemma dlimP f x : dlim_spec f x (dlim f x).
 Proof.
 rewrite dlimE; case: nlimP => [l h|?] /=; last by apply/DLimOut.
-have: (0%:E <= l)%O by apply/ncvg_geC: h => n; apply/ge0_mu.
-have: (l <= 1%:E)%O by apply/ncvg_leC: h => n; apply/le1_mu1.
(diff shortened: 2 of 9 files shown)


affeldt-aist:
Why not PR on branch analysis_270? (analysis_270_integration is regularly rebased on anslysis_270)

affeldt-aist:
Also, is it important to redefine the comparison operators in scope ereal_scope? Why not use order_scope (or %O) when one needs them?

pi8027:
@affeldt-aist 

> Why not PR on branch analysis_270? (analysis_270_integration is regularly rebased on anslysis_270)

There are 3 branches including `270` and this was so confusing to me (I still don't understand the difference between them), but I can reopen this for `analysis_270`.

> Also, is it important to redefine the comparison operators in scope ereal_scope? Why not use order_scope (or %O) when one needs them?

`order_scope` defines the `0` and `1` notations which mean the bottom and top of lattices. It's better to do not use `order_scope` to reason about ereals.

affeldt-aist:
> There are 3 branches including `270` and this was so confusing to me (I still don't understand the difference between them), but I can reopen this for `analysis_270`.

Sorry about that. `analysis_270_*` are branches that build upon `analysis_270`, each working on a different aspect but the common basis is `analysis_270`. (We could at least delete `analysis_270_integral`). 
 
> > Also, is it important to redefine the comparison operators in scope ereal_scope? Why not use order_scope (or %O) when one needs them?
> 
> `order_scope` defines the `0` and `1` notations which mean the bottom and top of lattices. It's better to do not use `order_scope` to reason about ereals.

Ok, I see. I was happy enough using 0%R and 1%R here and there, since so far the need for order_scope was limited.

pi8027:
> Ok, I see. I was happy enough using 0%R and 1%R here and there, since so far the need for order_scope was limited.

IMO, error messages produced by `(0 < x)%O` where `x : (R : numDomainType)` is difficult enough to let us lose time a lot. Those specialized notations save our time.

affeldt-aist:
> IMO, error messages produced by `(0 < x)%O` where `x : (R : numDomainType)` is difficult enough to let us lose time a lot. Those specialized notations save our time.

Ok. Otesudesuga, is it ok to PR on analysis_270?

pi8027:
Yes.

PR #166, opened 2019-12-03
compatibility with mathcomp 1.10.0

affeldt-aist:
Just to have a version that compiles with mathcomp 1.10.0.

Dropped the artificial dependency with bigenough in the opam file too.

The opam package can be shipped just after merge.


affeldt-aist:
I haven't figured out how to compile it also with mathcomp.1.9.0 because of L2161 of topology.v.

affeldt-aist:
> I haven't figured out how to compile it also with mathcomp.1.9.0 because of L2161 of topology.v.

Thanks @pi8027!

PR #168, opened 2020-01-10
Revise the use of scopes

pi8027:
- As I suggested in https://github.com/math-comp/analysis/pull/163#issuecomment-561187282, it is better to do not use `order_scope`.
- Both `ring_scope` and `ereal_scope` have the ordering notations, thus they should not be opened simultaneously.

cc: @affeldt-aist

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,7 +4,7 @@
   sha256 = "0jsisiw8yckq96r5rgdmkrl3a7y9vg9ivpw12h11m8w6rxsfn5m5";
 }),
 coq-version ? "8.9",
-mc ? {url = "https://github.com/math-comp/math-comp"; ref = "experiment/order";},
+mc ? {url = "https://github.com/math-comp/math-comp"; ref = "master";},
 print-env ? false
 }:
 let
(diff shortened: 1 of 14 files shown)


pi8027:
`real_scope` is declared and opened in several files but actually never be used. What is this?

affeldt-aist:
> What is this?

It's wip ^_^. I look at that on Monday, today's hospital day :-(.

affeldt-aist:
> it is better to do not use `order_scope`.

It is not obvious to me why as a user I should avoid the most generic notation
offered by order_scope. Since ereals have been declared as an ordered type,
isn't it for them in particular to enjoy the notation?

pi8027:
@affeldt-aist
```coq
From mathcomp Require Import all_ssreflect all_algebra.

(*
At least, using `order_scope` to reason about rings is a potential pitfall:
*)
Fail Check (fun (R : numDomainType) (x : R) => 0 <= x)%O.
(*
The command has indeed failed with message:
In environment
R : numDomainType
x : R
The term "x" has type "Num.NumDomain.sort R" while it is expected to have type "Order.POrder.sort ?t".
*)
```

pi8027:
Also, I think that simultaneously opening both `order_scope` and `ring_scope` makes printing ambiguous. I will try to produce an example if needed.

affeldt-aist:
@CohenCyril lectured me on `*_display`'s (in particular the usage of `%R`) and on the benefit of using `%E` when explicitly dealing with ereals.

PR #170, opened 2020-02-13
Analysis 270 hahnbanach

mkerjean:
- hahn-banach theorem (depending only of MathComp + Axioms)
- hahn-banach theorem applied to normed spaces.
- hahn_banach_applications also contains lemmas that could be added to normedtype.v (Section LinearContinuousBounded).


affeldt-aist:
This PR has been subsumed by the branch `mathcomp_master_hahnbanach`.

PR #173, opened 2020-04-11
Add setICl (intersection of set with its complement)

vlj:
Hi,

I'd like to add two lemma to topology, closureC (closure of complement is complement of interior) and interiorC (complement of interior is closure of complement). For this I need these two intermediate lemma which I think may more generally be useful for anyone.

I'm still learning the naming convention of mathcomp, but there is a clash between C for Complement and C for Commutativity. So while setCC may fits mathcomp convention, I didn't know how to name the third lemma since setIC was already taken.

### DIFF
--- a/theories/classical_sets.v
+++ b/theories/classical_sets.v
@@ -323,6 +323,9 @@ Qed.
 Lemma setDE {A} (X Y : set A) : X `\` Y = X `&` ~` Y.
 Proof. by []. Qed.
 
+Lemma nonsubset {A} (X Y:set A): ~ (X `<=` Y) -> X `&` ~` Y !=set0.
+Proof. by rewrite -setD_eq0 setDE -set0P => /eqP. Qed.
+
 Lemma setIid {A} (X : set A) : X `&` X = X.
 Proof. by rewrite predeqE => ?; split=> [[]|]. Qed.
 
@@ -337,6 +340,9 @@ Proof. by rewrite predeqE => ?; split=> [[]|]. Qed.
 
 Lemma set0I A (Y : set A) : set0 `&` Y = set0.
 Proof. by rewrite setIC setI0. Qed.
+
+Lemma setICl {A} (X : set A) : ~` X `&` X = set0.
+Proof. by rewrite predeqE => ?; split => // -[]. Qed.
 
 Lemma setIA {A} (X Y Z : set A) : X `&` (Y `&` Z) = X `&` Y `&` Z.
 Proof. by rewrite predeqE => ?; split=> [[? []]|[[]]]. Qed.


affeldt-aist:
We can maybe look at `finset.v` for the naming convention (`setCK` and `setICr`).

vlj:
Renamed setCK and setICr, and added another lemma (setC0).

affeldt-aist:
`setICr` should maybe be `setICl` because the complement operation is on the left-hand. These conventions are documented in `https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md`.

Also you may want to squash your commits into one commit.

theories/classical_sets.v
```diff
@@ -338,6 +344,9 @@ Proof. by rewrite predeqE => ?; split=> [[]|]. Qed.
 Lemma set0I A (Y : set A) : set0 `&` Y = set0.
 Proof. by rewrite setIC setI0. Qed.
 
+Lemma setICr {A} (X : set A) : ~` X `&` X = set0.
+Proof. by rewrite predeqE => ? //=; split =>  [[inX notinX] |]. Qed.
```
affeldt-aist, line 348: Proof. by rewrite predeqE => ?; split => // -[]. Qed.

theories/classical_sets.v
```diff
@@ -369,6 +378,9 @@ Proof. by rewrite funeqE => t; rewrite propeqE; split; [case|left]. Qed.
 Lemma setU_eq0 T (X Y : set T) : (X `|` Y = set0) = ((X = set0) /\ (Y = set0)).
 Proof. by rewrite -!subset0 subUset. Qed.
 
+Lemma setC0 {TT}: ~` set0  = @setT TT.
+Proof. by rewrite predeqE /setC /setT => x; split => //= H. Qed.
```
affeldt-aist, line 382: Proof. by rewrite predeqE => ?; split => // _ []. Qed.

theories/classical_sets.v
```diff
@@ -323,6 +326,9 @@ Qed.
 Lemma setDE {A} (X Y : set A) : X `\` Y = X `&` ~` Y.
 Proof. by []. Qed.
 
+Lemma nonsubset {A} (X Y:set A): ~ (X `<=` Y) -> X `&` ~` Y !=set0.
+Proof. by rewrite -setD_eq0 setDE -set0P  => /eqP. Qed.
```
affeldt-aist, line 330: Only one space before `=>`.

vlj:
Renamed to setICl, removed setCC, setC0 and setCK since they are already in master, and fixed -[] and the extra space. Everything is squashed into a single commit now.

PR #176, opened 2020-04-13
Mathcomp master hausdorff close

affeldt-aist:
These are cherry-picked commits from the branch `mathcomp_master_integral_ereal` that generalize lemmas from `topology.v` and `normedtype.v` using the notation of Hausdorff spaces.

The contents of this PR need to discussed and merged with those of PR #172


affeldt-aist:
My mistake: should be based on `mathcomp_master`.

PR #175, opened 2020-04-13
Update of master w.r.t. MathComp 1.11

affeldt-aist:
Update of master w.r.t. MathComp 1.11.

This PR should fix the following issues:
- fixes #4
- fixes #110
- fixed #61

PR #179, opened 2020-04-14
Mathcomp master hahnbanach

mkerjean:
This contains two files:

- `hahn_banach.v` contains a new formalization of Zorn's Lemma depending of the Choice in Prop and to Excluding middle, as well as a proof of Hahn_Banach theorem in its analytical form. This file is independent from MathComp Analysis libraries.

- `hahn_banach_applications.v`  contains a formalization of Hahn Banach's theorem on normed vector spaces, using the Mathematical Components libraries. 

This PR depends on PR#183


affeldt-aist:
When rebasing this PR / branch on PR #175, we have but back `theories/hahn_banach_applications.v` back in the `_CoqProject` and this has revealed a compilation problem apparently simply due to a misuse of `sup`.

CohenCyril:
should be reopened on top of master

CohenCyril:
Should be reopened on top of master

PR #177, opened 2020-04-13
Mathcomp master hausdorff close

affeldt-aist:
This PR:
- generalizes lemmas from `topology.v` and `normedtype.v` using the `hausdorff` predicate:
  + generalize the predicate `close x y` for two points of a topological space (was previously defined at the level of pseudometric spaces)
  + lemmas `flim_unique`, `locally_flim_unique`, `lim_id`, `flim_lim`, `flim_map_lim`, etc. are now defined at the level of topogical spaces
  + alternative definition of `hausdorff` for topological spaces (lemma `open_hausdorff`)
  + alternative definition of `hausdorff` for pseudometric spaces (lemma `ball_hausdorff`)
  + proof that normed modules are Hausdorff (lemma `normedModType_hausdorff`)
- fixes extended reals 
  + fix the definition of filters for extended reals (`ereal_locally`, `ereal_locally'`)
  + fix the definitions of inequalities
  + equip extended reals with a structure of topological space
  + proof that extended reals are Hausdorff

Start of this PR: 4 cherry-picked commits from the branch `mathcomp_master_integral_ereal`
ca4995dcdd835507bc37e58ef22da9b9ffd67d5c
87788a5670d3d13e96a3f04d90daff4b5806d148
42f184e428a1f4f1384c95de0ac1cb3cba153188
8dcc72a2ecf50ded58f8337137ee721a7e8f32d3
Add modifications introduced by commits
5e75df19fe7cacd6f19ac15aaaede249c3d620e1
d64953e35400f52eb86141be2f2732f31853067a
0d11d38d9cc6a3934abdf469b2f2429eadb5cfa3
29d807d7fc3fbcff1797da88670ab7edecb5e2f0
8f4c265681e03f8e014640ce13d604f3c97ab5f3
80f11f66164b519035b089611e61399d179de268
(Yet, the changes made to files specific to `mathcomp_master_integral_ereal` such as `integral.v` have not been reported here.)

Relation with PR #174: The contents of these 4 commits have been modified to be a superset of PR #174 (checked with @mkerjean):
- This PR provides proofs for `flimi_cluster` (now `flimi_close`) and `flim_clusterP` (now `flim_closeP`).
- `locally_flim_unique`, `lim_id`, `flim_map_lim`, `flimi_unique`, and `flimi_map_lim` have been generalized like they were in PR #174.

PR #174, opened 2020-04-13
Mathcomp master hausdorff

mkerjean:
A certain number of lemmas on the unicity of limits are using the predicate "close" on pseudoMetricTypes and normedModTypes (e.g flim_map_lim).

This PR aims at generalizing these lemmas to the level of Hausdorff topological types.

PR #182, opened 2020-04-14
lemmas for within and locally'

mkerjean:
A few lemmas for within.

### DIFF
--- a/theories/topology.v
+++ b/theories/topology.v
@@ -1852,6 +1852,21 @@ Typeclasses Opaque locally'.
 Canonical locally'_filter_on (T : topologicalType)  (x : T) :=
   FilterType (locally' x) (locally'_filter _).
 
+Lemma subset_filtermap (T U : Type) (f : T -> U):
+  forall (F G : set (set T)), G `=>` F -> f @ G `=>` f @ F.
+Proof. by move=> F G H A fFA ; exact: H (preimage f A) fFA. Qed.
+
+Lemma flim_within_filter {T U} {f : T -> U} (F : set (set T)) {FF : (Filter F) }
+  (G : set (set U)) : forall (D : set T), (f @ F) --> G -> (f @ within D F) --> G.
+Proof. move=> ?;  exact: flim_trans (subset_filtermap (flim_within _)). Qed. 
+
+Lemma cvg_within {T} {U : topologicalType} (f : T -> U) (F : set (set T))
+  (D : set T): Filter F -> cvg (f @ F) -> cvg (f @ within D F).
+Proof. by move => FF /cvg_ex [l H]; apply/cvg_ex; exists l; exact: flim_within_filter. Qed.
+
+Lemma locally_locally' {T : topologicalType} (x : T) : locally' x `=>` locally x.
+Proof. exact: flim_within. Qed.
+
 (** ** Closed sets in topological spaces *)
 
 Section Closed.


theories/topology.v
```diff
@@ -1852,6 +1852,39 @@ Typeclasses Opaque locally'.
 Canonical locally'_filter_on (T : topologicalType)  (x : T) :=
   FilterType (locally' x) (locally'_filter _).
 
+
+Lemma flimS {T U : topologicalType} {f : T -> U}:
+  forall (F G: set (set T)), (G `=>` F) -> (f @ G `=>` f @ F).
+Proof. by move => F G H A fFA ; exact ((H (preimage f A)) fFA). Qed.
```
affeldt-aist, line 1858: This lemma may be written as follows (it actually appears in PR #177 as a copy of a lemma that was in PR #174)

Lemma flim_incl (T U : Type) (f : T -> U) (F G : set (set T)) :
  G `=>` F -> f @ G `=>` f @ F.
Proof. move=> GF ? ?; exact: GF. Qed.

theories/topology.v
```diff
@@ -1852,6 +1852,39 @@ Typeclasses Opaque locally'.
...
+
+
+Lemma flim_within_filter {T U : topologicalType} {f : T -> U} (F : set (set T)) {FF: (Filter F) }
+      (G : set (set U)): forall (D : set T), ((f @ F) --> G) -> ((f @ within D F) --> G).
```
affeldt-aist, line 1862: This lemma also can be expressed at the level of `Type` (and proved `by move=> ?; exact: (flim_trans (flim_incl (flim_within _))).`).

theories/topology.v
```diff
@@ -1852,6 +1852,39 @@ Typeclasses Opaque locally'.
...
+  move => D ; rewrite /flim !filter_of_filterE.
+  have lem : (f @ within D F) `=>` (f @ F).
+  by apply: flimS; apply: flim_within.  
+  move => fFG; apply: flim_trans. 
+  by apply: (flimS lem).
+  by [].
+Qed. 
+
+
+Lemma cvg_within {T U : topologicalType} (f : T -> U) (F: set (set T)) (D: set T):
```
affeldt-aist, line 1873: Can't `T` be generalized to `Type`?

theories/topology.v
```diff
@@ -1852,6 +1852,39 @@ Typeclasses Opaque locally'.
...
+  (Filter F) -> cvg (f @ F) -> cvg (f @ within D F).
+Proof.
+  move => FF /cvg_ex [l H].
+  apply/cvg_ex; exists l. 
+  by apply: flim_within_filter.
+Qed.
+
+
+Lemma locally_locally' {T : topologicalType}:
+  forall (x : T), (locally' x) `=>` (locally x).
```
affeldt-aist, line 1883: We can remove parentheses around in the `=>` expression.

theories/topology.v
mkerjean: If it appears in PR #177 I should delete it then ?

affeldt-aist: No, I remove it from PR #177 because it is not used there.

CohenCyril: why is it named `flimS`, `S` is generally used for "subset" but `flim` is actually an inclusion of filter. This one should rather be named `flim_map` or `subset_filtermap`.

theories/topology.v
mkerjean: ok

theories/topology.v
mkerjean: ok

theories/topology.v
mkerjean: done

PR #183, opened 2020-04-14
LinearContinuousBounded

mkerjean:
Lemmas showing that for functions between normed vector spaces, boundedness and continuity are equivalent.

### DIFF
--- a/theories/normedtype.v
+++ b/theories/normedtype.v
@@ -3304,15 +3304,18 @@ Grab Existential Variables. by end_near. Qed.
 Lemma continuousfor0_continuous (f : {linear V -> W}) :
   {for 0, continuous f} -> continuous f.
 Proof.
-move=> /linear_continuous0 /linear_boundedP [r []].
-(*rewrite le_eqVlt => /orP[/eqP <- f0|r0 fr x].
-- suff : f = (fun _ => 0) :> (_ -> _) by move=> ->; exact: cst_continuous.
-  by rewrite funeqE => x; move: (f0 x); rewrite mul0r normr_le0 => /eqP.
-- rewrite cvg_to_locally => e e0; rewrite nearE /= nbhs_ballP.
-  exists (e / r); first by rewrite /= divr_gt0.
-  move=> y; rewrite -!ball_normE //= => xy; rewrite -linearB.
-  by rewrite (le_lt_trans (fr (x - y))) // mulrC -ltr_pdivl_mulr.
-Qed.*) Admitted.
+move=> /linear_continuous0 /linear_boundedP [y [yreal fr]].
+near (@pinfty_nbhs R) => r.
+have r0  : 0 < r.
+  by near: r; exists 1; rewrite real1; split => // z; apply le_lt_trans.
+move=> x; apply/cvg_ballP => _/posnumP[e]; rewrite nearE /= nbhs_ballP.
+exists (e%:num / r); first by rewrite /= divr_gt0.
+move=> z; rewrite -!ball_normE //= => xy; rewrite -linearB.
+move: xy; rewrite ltr_pdivl_mulr //; apply le_lt_trans.
+rewrite mulrC; apply fr.
+near: r; exists (y + 1); rewrite realD // ?real1 //; split => // x.
+by apply le_lt_trans; rewrite ler_addl.
+Grab Existential Variables. by end_near. Qed.
 
 Lemma linear_bounded_continuous (f : {linear V -> W}) :
   bounded_on f (nbhs (0 : V)) <-> continuous f.
@@ -3322,30 +3325,18 @@ exact: continuousfor0_continuous.
 Qed.
 
 Lemma bounded_funP (f : {linear V -> W}):
-  (forall r, exists M, forall x, `|x| <= r -> `|f x| <= M) <-> bounded_on f (nbhs (0 : V)).
+  (forall r, exists M, forall x, `|x| <= r -> `|f x| <= M) <->
+  bounded_on f (nbhs (0 : V)).
 Proof.
-split => [/(_ 1) [M Bf]|/linear_boundedP [r [r0 Bf]] e] ; last first.
-  (*by exists (e * r) => x xe; rewrite (le_trans (Bf _)) // mulrC ler_pmul.*) admit.
-apply/linear_boundedP.
-have M0 : 0 <= M by move: (Bf 0); rewrite linear0 !normr0 => /(_ ler01).
-exists M; split => //.
-admit.
-(*move: M0; rewrite le0r => /orP[/eqP|] M0 x.
-- rewrite M0 mul0r normr_le0.
-  have [/eqP ->|x0] := boolP (x == 0); first by rewrite linear0.
-  have /Bf : `| `|x|^-1 *: x | <= 1.
-    rewrite normmZ normrV ?unitf_gt0 ?normrE //=.
-    by rewrite mulrC mulrV ?unitf_gt0 ?normrE.
-  rewrite linearZ M0 normr_le0 scaler_eq0 invr_eq0 => /orP[|//].
-  by rewrite normr_eq0 => /eqP ->; rewrite linear0.
-- have [/eqP ->|x0] := boolP (x == 0).
-  + by rewrite normr0 mulr0 normr_le0 linear0.
-  + rewrite -ler_pdivr_mulr ?normr_gt0 //= mulrC.
-     have xx1 : `| `|x|^-1 | = `|x|^-1.
-       by rewrite normrV ?unitfE ?normr_eq0 // normr_id.
-     rewrite -xx1 -normmZ -linearZ; apply: (Bf (`|x|^-1 *: x)).
-     by rewrite normmZ xx1 mulVr // unitfE normr_eq0.
-Qed.*) admit.
-Admitted.
+split => [/(_ 1) [M Bf]|/linear_boundedP fr y].
+  apply/ex_bound; exists M; apply/nbhs_ballP; exists 1 => // x.
+  rewrite -ball_normE /ball_ /= sub0r normrN => x1.
+  exact/Bf/ltW.
+near (@pinfty_nbhs R) => r; exists (y * r) => x xe.
+rewrite mulrC (@le_trans _ _ (r * `|x|)) //.
+  by move: {xe} x; near: r.
+rewrite ler_pmul //; near: r; exists 1.
+by rewrite real1; split => // x /ltW; apply le_trans.
+Grab Existential Variables. by end_near. Qed.
 
 End LinearContinuousBounded.
--- a/theories/normedtype.v
+++ b/theories/normedtype.v
@@ -3237,14 +3237,14 @@ Section LinearContinuousBounded.
 
 Variables (R : numFieldType) (V W : normedModType R).
 
-Lemma linear_boundedP (f : {linear V -> W}) : bounded_on f (nbhs (0 : V)) <->
+Lemma linear_boundedP (f : {linear V -> W}) : bounded_near f (nbhs (0 : V)) <->
   \forall r \near +oo, forall x, `|f x| <= r * `|x|.
 Proof.
 split=> [|/pinfty_ex_gt0 [r r0 Bf]]; last first.
   apply/ex_bound; exists r; apply/nbhs_ballP; exists 1 => // x.
   rewrite -ball_normE //= sub0r normrN -(gtr_pmulr _ r0) => /ltW.
   exact/le_trans/Bf.
-rewrite /bounded_on => /pinfty_ex_gt0 [M M0 /nbhs_ballP [_/posnumP[e] efM]].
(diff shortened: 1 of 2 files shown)


theories/normedtype.v
```diff
@@ -2600,3 +2600,96 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
   (f : Rdefinitions.R -> Rdefinitions.R) (x : Rdefinitions.R) :
   locally (f x) P -> Ranalysis1.continuity_pt f x -> \near x, P (f x).
 Proof. by move=> Lf /continuity_pt_flim; apply. Qed.
+
+
+Section LinearContinuousBounded.
+
+Variables (R: numFieldType) (V  W: normedModType R).
+
+Definition continuous_at (x : V) (f : V -> W) :=  f @ x --> (f x).
```
CohenCyril, line 2609: `continuous_at x f` should be perhaps rephrased `{at x, continuous f}`

theories/normedtype.v
```diff
@@ -2600,3 +2600,96 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
...
+    by apply: ltW.
+    by apply: mulr_gt0.
+    apply: mulr_gt0 ; last by [].
+       apply : mulr_gt0 ; last by [].
+      by rewrite invr_gt0 //=.
+   suff -> : a = (`|x|^-1 * t/2) *: x + 0  by  rewrite linearP (linear0 f) addrC add0r.
+     by rewrite addrC add0r.
+Qed.     
+
+(* rewrite using landau notations ? *)
```
CohenCyril, line 2660: I thinks some elements of answer and proof can be found here: https://github.com/math-comp/analysis/blob/master/theories/derive.v#L152-L166

CohenCyril: and maybe better her:
https://github.com/math-comp/analysis/blob/master/theories/landau.v#L1061-L1092

mkerjean: > I think some inspiration in statements and proofs can be found here: https://github.com/math-comp/analysis/blob/master/theories/derive.v#L152-L166

This is stated to be proof that the identity is locally bounded. Is it something else ?

mkerjean: > and maybe better here:
> https://github.com/math-comp/analysis/blob/master/theories/landau.v#L1061-L1092

Indeed, this lemma implies bounded_continuous0. I will update the PR accordingly. I was unable to find it at the time. Could the name of the lemma contain "bounded" to help ? 
I have tried several times to use landau notations unsuccessfully, do you strongly prefer these lemmas to be using landau notations ?

theories/normedtype.v
mkerjean: Will do. Where is this notation defined ? Could it be added to the header ?

theories/normedtype.v
mkerjean: If you would prefer these lemmas to rewritten using landau notations and the lemma `linear_for_continuous`, it means that they should be part of `landau.v` instead of  `normedtype.v`. Should I do that anyway ?

theories/normedtype.v
CohenCyril: It comes from mathcomp https://github.com/coq/coq/blob/master/theories/ssr/ssrbool.v#L251

theories/normedtype.v
CohenCyril: > This is stated to be proof that the identity is locally bounded. Is it something else ?

No it is stated that it contains such a proof, inside the proof script.

> Indeed, this lemma implies bounded_continuous0. I was unable to find it at the time.

More interestingly the proof contains some of the steps mentioned in the comment. Splitting it amounts to stating the implications independently and extracting their proof from lemme `linear_for_continuous`.

> Could the name of the lemma contain "bounded" to help ?
> 

I think it should be eventually restated as an equivalence between `continous` and `{for 0, locally_bounded f}` where
```coq
locally_bounded f := forall x, \forall A \near +oo, \forall y \near x, `|f x| <= A.
```
Additionally, we should have a (trivial) equivalence lemma with `{for x, locally_bounded f} <-> (f : _ -> _) =O_x (cst 1)` because I suspect many things will be easier to prove with landau.

theories/normedtype.v
mkerjean: Could it be added to the header ?

theories/normedtype.v
mkerjean: I am still not sure the lemma should not contains `bounded` in its name. For now, I am unable to use \near or landau notations.

CohenCyril: > I am still not sure the lemma should not contains bounded in its name - as it is is suggests that all linear maps are continuous.

If we keep the current shape of statement, it should be named `linear_locally_bounded_continuous`, however, if we turn it into the equivalence `continuous f = {for 0, locally_bounded f}` then the convention is that only the LHS of the equation should appear in the name, as it will most of the time be used from left to right, to reduce continuity of linear application to locally bounding in 0.

> I have repeatedly been unable to use landau notations so they don't always make proofs easier for their authors.

@mkerjean I'd be happy to see where you tried and what failed, in order to improve on the documentation.

theories/normedtype.v
```diff
@@ -2600,3 +2600,94 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
...
+
+
+Section LinearContinuousBounded.
+
+Variables (R: numFieldType) (V  W: normedModType R).
+
+Lemma continuous_atP x (f : V -> W) :
+ {for x, continuous f} <-> (forall eps : R,  0 < eps -> \forall y \near f @ x, ball (f x) eps y) .
+Proof.
+   split; by  move => /flim_ballP.
```
CohenCyril, line 2612: isn't it `exact: flim_ballP`?

theories/normedtype.v
```diff
@@ -2601,93 +2601,55 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
...
-      by rewrite invr_gt0 //=.
-   suff -> : a = (`|x|^-1 * t/2) *: x + 0  by  rewrite linearP (linear0 f) addrC add0r.
-     by rewrite addrC add0r.
-Qed.     
+  {for 0, continuous f} -> exists r, r > 0 /\ forall x, `|f x| <= `|x| * r.
+Proof.
+move=> /continuous_atP /(_ _ ltr01).
+rewrite linear0 /= nearE => /locally_ex[tp ball_f].
+pose t := tp%:num; exists (2 * t^-1); split => // x.
+have [->|/eqP x0] := boolp.EM (x = 0); first by rewrite linear0 !normr0 mul0r.
```
CohenCyril, line 2618: You would prefer `have [->|x0] := eqVneq x 0`, which does not explicitly rely on excluded middle (`R` could very well have "true" decidable equality).
```suggestion
have [->|x0] := eqVneq x 0; first by rewrite linear0 !normr0 mul0r.
```

theories/normedtype.v
```diff
@@ -2601,93 +2601,55 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
   locally (f x) P -> Ranalysis1.continuity_pt f x -> \near x, P (f x).
 Proof. by move=> Lf /continuity_pt_flim; apply. Qed.
 
-
 Section LinearContinuousBounded.
 
-Variables (R: numFieldType) (V  W: normedModType R).
+Variables (R : numFieldType) (V W : normedModType R).
 
 Lemma continuous_atP x (f : V -> W) :
```
CohenCyril, line 2608: since it is exacly `flim_ballP`, I am wondering if we should have this lemma at all...

theories/normedtype.v
```diff
@@ -2601,93 +2601,55 @@ Lemma locally_pt_comp (P : Rdefinitions.R -> Prop)
...
-  rewrite invf_div => Ht. 
-    by apply: ltW.
-    by apply: mulr_gt0.
-    apply: mulr_gt0 ; last by [].
-       apply : mulr_gt0 ; last by [].
-      by rewrite invr_gt0 //=.
-   suff -> : a = (`|x|^-1 * t/2) *: x + 0  by  rewrite linearP (linear0 f) addrC add0r.
-     by rewrite addrC add0r.
-Qed.     
+  {for 0, continuous f} -> exists r, r > 0 /\ forall x, `|f x| <= `|x| * r.
```
CohenCyril, line 2613: I personally prefer ``\forall r \near +oo, forall x, `|f x| <= `|x| * r.`` because then, one can extract an epsilon as big as they want, not necessarily greater than `0` (there might still be missing theorems to handle it, but it is worth adding them), and to prove, one could either choose between the ``exists r, forall x, `|f x| <= `|x| * r`` presentation (no constraint needed on `r`), or using `near=> r` to prove it, which can lead to huge simplifications in some cases.

theories/normedtype.v
mkerjean: Indeed, I will rewrite that.

theories/normedtype.v
mkerjean: For the accessibility of the code ? It would maybe allow to handle basic continuity notions without handling filter.

theories/normedtype.v
mkerjean: Indeed, thanks for the remark.

mkerjean: I still find the notation near uneasy to use, but I guess I can see the interest now. thanks for the explanation.

theories/normedtype.v
CohenCyril: ok

theories/normedtype.v
CohenCyril: You could just click on "commit suggestion"

CohenCyril:
Should be reopened on top of master (please add a reference to this PR so that we can continue the discussions)

theories/normedtype.v
```diff
@@ -4362,7 +4362,7 @@ Proof.
   by move => e0; apply: subset_closure; apply: ballxx.
 Qed.
 
-Lemma le_orP (R : numDomainType) (x y : R) : (y <= x) = (x == y) || (y < x).
+Lemma le_orE (R : numDomainType) (x y : R) : (y <= x) = (x == y) || (y < x).
 Proof. by rewrite lt_def; case: eqP  => // ->; rewrite lexx. Qed.
```
pi8027, line 4366: `rewrite le_eqVlt eq_sym.` ?

theories/normedtype.v
```diff
@@ -4381,35 +4382,46 @@ move=> x e e0; apply: eqEsubset =>y.
...
 move => xy; rewrite  /closed_ball /closed_ball_  /= closureI /closed /closure.
-rewrite le_orP; case/predU1P; last first.
+rewrite le_orE; case/predU1P; last first.
   by rewrite -ball_normE => xye B0 [/(_ y xye) B0B _].
 rewrite -ball_normE // => xye B [Be Bc].
 apply: Bc => B0 /nbhs_ballP [s s0]; rewrite -ball_normE //= => B0y.
-pose r := 2^-1 * (minr e s); simpl in (type of r).
-exists (x + (r *: (`| x - y|^-1 *: (x -y)))); split.
-  apply: Be; move=> //=.
+have: (s > e) \/ (s <= e). (* todo: lemma in realType*)
```
pi8027, line 4391: Doesn't `case: (leP s e)` work?

mkerjean: It does thanks. It did not find it through `Search` tough.

theories/normedtype.v
mkerjean: thanks ! What does `eqVlt` stands for ?

pi8027: `V` means "or".

mkerjean:
Should I open a new PR for `closed_ball` ?

affeldt-aist:
> Should I open a new PR for `closed_ball` ?

I think that that was one of the conclusions of the last meeting (the minutes seem to say so).

theories/normedtype.v
```diff
@@ -3165,39 +3148,38 @@ by rewrite rpredM// ?ger0_real ?invr_ge0// ltW.
...
-        exists r : R, 0 <= r /\ (forall x : V, `|f x| <= r * `|x|).
-Proof.
-split => [/linear_bounded | B].
-  by move=> /pinfty_ex_gt0  [M [M0 B]]; exists M; split; first by apply: ltW.
-apply/linear_bounded; near_simpl. Admitted.
-
 Lemma linear_continuous0 (f : {linear V -> W}) :
   {for 0, continuous f} -> bounded_on f (nbhs (0 : V)).
-Proof.
+Proof. (*TODO : rewrite with near ? *)
```
mkerjean, line 3153: Should this proof be rewritten with near tactics ? I tried for one hour and gave up.

affeldt-aist:
(I just rebased to get rid of conflicts so that we can merge.)

mkerjean:
Should this be merged ?

CohenCyril (approved):
Yes! looks good to me!

PR #186, opened 2020-04-20
Add setCU (complement of union is intersection of complements) lemma.

vlj:
I used boolean reflection to prove this lemma, I'm not sure this is the best way.


affeldt-aist:
We can save one line!
```
rewrite predeqE => z.
by apply: asbool_eq_equiv; rewrite asbool_and !asbool_neg asbool_or negb_or.
```
:-)

vlj:
I updated the PR following affeldt-aist advice, I didn't see the negb_or lemma.

PR #193, opened 2020-05-01
Renaming flim to cvg

### LINKED ISSUE #29: naming of flim_add
https://github.com/math-comp/analysis/blob/aa351f34edbf0c45fc88cd5528749803bf1e8b3f/hierarchy.v#L1490

Is that a good name? With that name, I was rather expecting a lemma such as
Lemma lim_add (F : filter_on X) (f g : X -> Y) (a b : Y) :
  f @ F --> a -> g @ F --> b -> (f + g) @ F --> a + b.
(no prod, following flim_const).

CohenCyril:
Fixes #29

- `cvg` corresponds to `_ --> _`
- `lim` corresponds to `lim _`
- `continuous`  corresponds to continuity
- some suffixes _opp, _add ... renamed to N, D, ...

Apply this substitution **before** rebasing on top of this commit:
```bash
sed -i -f etc/pr193.sed **/*.v
```

@math-comp/analysis what do you think about this conventions?

### DIFF
--- a/etc/pr193.sed
+++ b/etc/pr193.sed
@@ -1,0 +1,25 @@
+s/cvg_within/cvg_app_within/g;
+s/cvg/PREPROCESS/g;
+s/\([^n]\)lim_add/\1cvgD/g;
+s/\([^n]\)lim_opp/\1cvgN/g;
+s/\([^n]\)lim_mult/\1cvgM/g;
+s/\([^n]\)lim_mul/\1cvgM/g;
+s/\([^n]\)lim_scale/\1cvgZ/g;
+s/\([^n]\)lim_sub/\1cvgN/g;
+s/\([^n]\)lim_cst/\1cvg_cst/g;
+s/\([^n]\)lim_cont/\1cvg_cont/g;
+s/Definition flim/Definition cvg_to/g;
+s/(flim F G)/(cvg_to F G)/g;
+s/(flim \[/(cvg_to \[/g;
+s/flim/cvg/g;
+s/approaches/cvg/g;
+s/approach/cvg/g;
+s/\(is_\)\?filter_\?lim/cvg/g;
+s/filtermap/fmap/g;
+s/close_lim/close_cvg/g;
+s/Flim/Cvg/g;
+s/_cont\([^a-z]\)/_continuous\1/g
+s/fcvg/cvg/g;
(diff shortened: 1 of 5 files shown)


affeldt-aist:
Wouldn't `appr` be more natural as a prefix?

mkerjean:
Do you have an example of simplifying allowed by this renaming ?

CohenCyril:
@affeldt-aist 
> Wouldn't `appr` be more natural as a prefix?

Maybe, I don't know, `appr` makes me think of `app` as in `oapp`, buut maybe it is not that bad...

@mkerjean 
> Do you have an example of simplifying allowed by this renaming ?

The main simplification is documentation and consistency of notations. Right now `lim_add`, `lim_opp`... should be named `flim_add`, `flim_opp` instead. But the visual distinction between `lim` and `flim` is not clear enough in my opinion, and explaining `flim` goes through the history of coquelicot, where it was called `filterlim`, and since `f @ x --> y` is supposed to be handled without the explicit use of filters in most cases, it seems unnatural to rely on a name which contains the word `filter` in it, hence the switch to `approaches` or rather an abbreviation of it.

affeldt-aist:
What about using "tends to" instead of "appraoches" and abbreviate it as "tto"?

CohenCyril:
So should we have a poll?
/polls `flim` `apr` `appr` `tto` `tendsto` `cvg`

CohenCyril:
I force pushed to show how it would look with `cvg`

affeldt-aist:
But then shouldn't, e.g., `lim_scale` become `cvgZ`?

CohenCyril:
If it's too confusing, `cvg := F --> lim F` could be renamed `is_cvg := F --> lim F`.

CohenCyril:
> But then shouldn't, e.g., `lim_scale` become `cvgZ`?

I did several mistakes in the sed script, they should be fixed now. Can you refresh and tell me it you see a problem?

CohenCyril:
The checks finally pass, and I am happy with the result. @mkerjean what do you think?

mkerjean:
`cvg` seems great (sorry, I've been busy with job applications).

CohenCyril:
> `cvg` seems great

Thanks for your input :smiley: 

> (sorry, I've been busy with job applications).

good luck!!

PR #196, opened 2020-05-06
Generic domination, boundedness and lipschitz

CohenCyril:
- See header for the notations and their localization.
- Added a bunch of combinator to extract existential witnesses.
- Added `filter_forall`, (commutation between a filter and finite forall).


CohenCyril:
@mkerjean @affeldt-aist I implemented the generic boundedness and lipshitz notations in an ever more general way than described here: https://github.com/math-comp/analysis/pull/183#discussion_r408341345
Indeed `[locally G]` is a combinator that can be used either on boundedness or lipshitz predicates to make them local instead of global.

CohenCyril:
@affeldt-aist can I merge?

affeldt-aist:
> @affeldt-aist can I merge?

Yes.

PR #197, opened 2020-05-06
contributing file

affeldt-aist:
The goal is to explain good practices that are specific to mathcomp-analysis. Still embryonic.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,9 +10,11 @@ guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md).
 Always submit a pull request for code and wait for the CI to pass before merging.
 Text markup files may be edited directly though, should you have commit rights.
 
-## `=>` vs. `-->` vs. `cvg` vs. `lim`
+## `-->` vs. `cvg` vs. `lim`
 
-TODO: explain in particular the lemmas `cvgP`, `cvg_ex`
+- `F --> x` means `F` tends to `x`. _This is the preferred way of stating a convergence._ **Lemmas about it use the string `cvg`.**
+- `lim F` is the limit of `F`, it makes sense only when `F` converges and defaults to a distinguished point otherwise. _It should only be used when there is no other expression for the limit._ **Lemmas about it use the string `lim`.**
+- `cvg F` is defined as `F --> lim F`, and is equivalent through `cvgP` and `cvg_ex` to the existence of some `x` such that `F --> x`. _When the limit is known, `F --> x` should be preferred._ **Lemmas about it use the string `is_cvg`.**```
 
 ## `near` tactics vs. `filterS`, `filterS2`, `filterS3` lemmas
 
@@ -34,7 +36,13 @@ and its variants (`filterS2`, `filterS3`, etc.).
 
 ## Landau notations
 
-TODO
+Landau notations can be written in four shapes:
+- `f =o_F e` (i.e. functional with a simple right member, thus a binary notation)
+- `f = g +o_F e` (i.e. functional with an additive right member, thus ternary)
+- `f x =o_(x \near F) (e x)` (i.e. pointwise with a simple right member, thus binary)
+- `f x = g x +o_(x \near F) (e x)` (i.e. pointwise with an additive right member, thus ternary)
+
+The outcome is an expression with the normal Leibniz equality `=` and term `'o_F` which is not parsable. See [this paper](https://doi.org/10.6092/issn.1972-5787/8124) for more explanation and the header of the file [landau.v](https://github.com/math-comp/analysis/blob/master/theories/landau.v).
 
 ## Naming convention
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
+# Contribution Guide for the mathcomp-analysis	library (WIP)
+
+The purpose of this file is to document scripting techniques to be
```
CohenCyril, line 3:
```suggestion
The purpose of this file is to document coding styles to be
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+Text markup files may be edited directly though, should you have commit rights.
+
+## `=>` vs. `-->` vs. `cvg` vs. `lim`
+
+TODO: explain in particular the lemmas `cvgP`, `cvg_ex`
+
+## `near` tactics vs. `filterS`, `filterS2`, `filterS3` lemmas
+
+When dealing with limits, mathcomp-analysis favors filters
+that are behind goals such as
```
CohenCyril, line 20:
```suggestion
phrasing, as in
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+```
+{near \oo, G}.
```
CohenCyril, line 22:
```suggestion
\forall x \near \oo, P x.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+```
+In the presence of such goals, the `near` tactics can be used to
+recover epsilon-delta reasoning
+(see [this paper](https://doi.org/10.6092/issn.1972-5787/8124)).
+
+However, when the proof does not require changing the epsilon it
+is often worth using filters and lemmas such as
```
CohenCyril, line 29:
```suggestion
is might be worth using filter combinators, i.e. lemmas such as
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+
+TODO
+
+## Naming convention
+
+### homo and mono notations
+
+Statements of `{homo ...}` or `{mono ...}` shouldn't be named after `homo`, or `mono`
+(just as for `{morph ...}` lemmas). Instead use the head of the unfolded statement
+(for `homo`) or the head of the LHS of the equality (for `mono`).
```
CohenCyril, line 45:
```suggestion
(for `homo`) or the head of the LHS of the equality (for `mono`), as in
```coq
Lemma le_contract : {mono contract : x y / (x <= y)%O}.
```
```

File: CONTRIBUTING.md [hunk 6]
```diff
@@ -0,0 +1,73 @@
...
+However, when the proof does not require changing the epsilon it
+is might be worth using filter combinators, i.e. lemmas such as
+```
+filterS : forall T (F : set (set T)), Filter F -> forall P Q : set T, P `<=` Q -> F P -> F Q
+```
+and its variants (`filterS2`, `filterS3`, etc.).
+
+## Landau notations
+
+TODO
```
CohenCyril, line 37:
```suggestion
Landau notations can be written in four shapes:
- `f =o_F e` (i.e. functional with a simple right member, thus a binary notation)
- `f = g +o_F e` (i.e. functional with an additive right member, thus ternary)
- `f x =o_(x \near F) (e x)` (i.e. pointwise with a simple right member, thus binary)
- `f x = g x +o_(x \near F) (e x)` (i.e. pointwise with an additive right member, thus ternary)

The outcome is an expression with the normal Leibniz equality `=` and term `'o_F` which is not parsable. See [this paper](https://doi.org/10.6092/issn.1972-5787/8124) for more explanation and the header of the file [landau.v](https://github.com/math-comp/analysis/blob/master/theories/landau.v).
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+used when contributing to mathcomp-analysis. It comes as an addition
+to mathcomp's [contribution
+guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md).
+
+## Policy for Pull Requests
+
+Always submit a pull request for code and wait for the CI to pass before merging.
+Text markup files may be edited directly though, should you have commit rights.
+
+## `=>` vs. `-->` vs. `cvg` vs. `lim`
```
CohenCyril, line 13:
```suggestion
## `-->` vs. `cvg` vs. `lim`
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+
+TODO: explain in particular the lemmas `cvgP`, `cvg_ex`
```
CohenCyril, line 15: - `F --> x` means `F` tends to `x`. _This is the preferred way of stating a convergence._ **Lemmas about it use the string `cvg`.**
- `lim F` is the limit of `F`, it makes sense only when `F` converges and defaults to a distinguished point otherwise. _It should only be used when there is no other expression for the limit._ **Lemmas about it use the string `lim`.**
- `cvg F` is defined as `F --> lim F`, and is equivalent through `cvgP` and `cvg_ex` to the existence of some `x` such that `F --> x`. _When the limit is known, `F --> x` should be preferred._ **Lemmas about it use the string `is_cvg`.**

PR #203, opened 2020-05-11
Supremums ereal

affeldt-aist:
The goal of this PR is to prove that a set of extended reals always has a supremum. For this purpose, I needed to generalize the definition of greatest elements, supremums, etc. from `reals.v`. I tried for a while to stick to the original `pred`-base definitions but `set`s from `classical_sets.v` turned out to be more handy and simplify proofs a bit.The first commit does the generalization, the second one adds the lemmas about extended reals.

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,7 +4,22 @@
 
 ### Added
 
+- in `boolp.v`, lemmas for classical reasoning `existsNP`, `existsP`,
+  `forallNP`, `forallP`, `imply_classic`, `orC`.
+- in `classical_sets.v`, definitions for supremums: `ul`, `lb`,
+  `supremum`
+- in `ereal.v`:
+  + technical lemmas `lee_ninfty_eq`, `lee_pinfty_eq`, `lte_subl_addr`
+  + lemmas about supremum: `ereal_supremums_neq0`
+  + definitions:
+    * `ereal_sup`, `ereal_inf`
+  + lemmas about `ereal_sup`:
+    * `ereal_sup_ub`, `ub_ereal_sup`, `ub_ereal_sup_adherent`
+
 ### Changed
+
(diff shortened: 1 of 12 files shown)


affeldt-aist:
@amahboubi @strub do you have an opinion about replacing `pred` with `set` has done in this commit https://github.com/math-comp/analysis/pull/203/commits/48661cc903ff48df35739e5df60df6ba93db049c ? It looked a bit more comfortable to me.

affeldt-aist:
@CohenCyril do you mean this commit https://github.com/math-comp/analysis/commit/f8d361a93db5906a73327d4fcf1fe9070e1960b8 when you refer to the the new `inE`?

affeldt-aist:
@CohenCyril so your opinion is to stick to `pred`-based defnitions?

affeldt-aist:
(Trying to summarize.) `pred`-based predicates and the `\in` notation might be viable thanks to 
- https://github.com/math-comp/analysis/pull/202 (already merged)
- https://github.com/math-comp/math-comp/pull/501 (to be merged)

For the time being, this PR sticks to `set`-based predicates, avoids the `\in` notation, and
tries not to unfold `lb`. `ub`, etc. That does not reduce the size of proofs much, but makes
appear a few factorizations and enables some generalizations.

affeldt-aist:
As an additional piece of information about `pred` vs. `set`, the commit https://github.com/math-comp/analysis/pull/203/commits/31f83c25912286ac73822b68b690df356807ae1b replays `sequences.v` with `set`-based predicates.

affeldt-aist:
We agreed on merging this PR during the last meeting but it seems that it incidentally implements one aspect of PR #117 (this was observed by @mkerjean ).

@amahboubi Are you also ok with merging?

amahboubi:
No objection.

CohenCyril:
> No objection.

Thanks @amahboubi, I will merge then.

PR #205, opened 2020-05-15
Topological structures on numfields

mkerjean:
This PR aims to get rid of the notation ^ o, by endowing any numFieldType with a canonical normedModType structure.


theories/normedtype.v
```diff
@@ -133,6 +133,82 @@ Definition pseudoMetric_of_normedDomain
...
+Definition numfield_lmodMixin :=
+  let mkMixin := @GRing.Lmodule.Mixin K  _ (@GRing.mul K) in
+  mkMixin (@GRing.mulrA K) (@GRing.mul1r K) (@GRing.mulrDr K)
+          (fun v a b => GRing.mulrDl a b v).
+Canonical numfield_lmodType := LmodType K K numfield_lmodMixin.
+Canonical numfield_regular_lalgType := LalgType K K (@GRing.mulrA _).
+(*Not enough to  to make _*_ and _*:_ definitionnaly equal *) (*TODO *)
+End vecspace_of_numfield.
+
+Section pseudoMetric_of_normedDomainnumField. (*NEW *)
```
CohenCyril, line 152: This section looks dangerous to me. Is it necessary to make things work?

CohenCyril:
I am not sure it is the problem, but I see one possible explanation: you are missing the closure under the join, e.g. solution to `GRing.Ring.sort ?1 = GRing.LModule.sort ?2`, must go through `GRing.LAlgebra` and `GRing.Field.sort ?1 = GRing.LModule.sort ?2` must go through `FieldExtension`. So it seems to me that providing canonical structures only up to `lmodType` may cause unification failures and the fix is to state and make canonical that `F` is a (trivial) field extension over itself...

mkerjean:
> I am not sure it is the problem, but I see one possible explanation: you are missing the closure under the join, e.g. solution to `GRing.Ring.sort ?1 = GRing.LModule.sort ?2`, must go through `GRing.LAlgebra` and `GRing.Field.sort ?1 = GRing.LModule.sort ?2` must go through `FieldExtension`. So it seems to me that providing canonical structures only up to `lmodType` may cause unification failures and the fix is to state and make canonical that `F` is a (trivial) field extension over itself...

Would that be a way to resolve the rift between `NormedModule.lmodType` and  `numfield_lmodType` ? I'm puzzled about wether a new structure mimicking numFieldType but with a forgetful functor to lmodType is necessary. 
Also where can I find `FieldExtenstion` ?

CohenCyril:
> > I am not sure it is the problem, but I see one possible explanation: you are missing the closure under the join, e.g. solution to `GRing.Ring.sort ?1 = GRing.LModule.sort ?2`, must go through `GRing.LAlgebra` and `GRing.Field.sort ?1 = GRing.LModule.sort ?2` must go through `FieldExtension`. So it seems to me that providing canonical structures only up to `lmodType` may cause unification failures and the fix is to state and make canonical that `F` is a (trivial) field extension over itself...
> 
> Would that be a way to resolve the rift between `NormedModule.lmodType` and `numfield_lmodType` ? I'm puzzled about wether a new structure mimicking `numFieldType` but with a forgetful functor to lmodType is necessary.

I do not think a new structure is necessary. But I could be wrong... but it's difficult to say until we check whether all the joins can be defined...

> Also where can I find `FieldExtenstion` ?

https://github.com/math-comp/math-comp/blob/master/mathcomp/field/fieldext.v#L10

theories/normedtype.v
```diff
@@ -260,6 +260,68 @@ Definition pseudoMetric_of_normedDomain
...
+Canonical numfield_algType :=  AlgType K K (scalerCA_numField).
+
+End vecspace_of_numfield.
+
+
+Section numField_canonical. (*New *)
+Variable R : numFieldType.
+(*Canonical topological_of_numFieldType := [numFieldType of R^o].*)
+Canonical numFieldType_pointedType :=
+  [pointedType of R for pointed_of_zmodule R].
```
pi8027, line 297: This `Canonical` declaration makes `numFieldType` inheriting from `pointedType`, but by construction rather than inclusion. Actually, this can be generalized to make `zmodType` inheriting from `pointedType`, and then you have to declare unification hints between `Pointed.sort` and all the subclasses of `zmodType`. Even if you do not generalize it, similar unification hints for `numClosedFieldType`, `realFieldType`, `archiFieldType`, `rcfType`, and `realType` have to be declared.

pi8027: I guess similar issues appear almost everywhere, and it is too painful to fix.

theories/normedtype.v
```diff
@@ -260,6 +260,68 @@ Definition pseudoMetric_of_normedDomain
...
+  forall x y : K, x * y = x *: y. (*TODO: naming *)
+Proof. by []. Qed.
+
+Lemma scalerCA_numField (x y z : K) :
+  x *: (y * z) = y * (x *: z).
+Proof.
+ by rewrite -!numfield_scale_mul mulrA mulrA [in _ * y]mulrC.
+Qed.
+
+Canonical numfield_algType :=  AlgType K K (scalerCA_numField).
```
pi8027, line 288: There is an easier way to do this by relying on the regular (^o) instances.
```suggestion
Canonical numfield_lmodType := [lmodType K of K for [lmodType K of K^o]].
Canonical numfield_lalgType := [lalgType K of K for [lalgType K of K^o]].
Canonical numfield_algType := [algType K of K for [algType K of K^o]].
```

mkerjean: I guess I would like not to rely on the canonical ^o instances, as I'm afraid this will introduce the usual confusion on the lmodType structure of K^o.

theories/normedtype.v
```diff
@@ -260,6 +260,68 @@ Definition pseudoMetric_of_normedDomain
   := PseudoMetricMixin ball_norm_center ball_norm_symmetric ball_norm_triangle erefl.
 End pseudoMetric_of_normedDomain.
 
+Canonical pseudoMetric_of_normedDomain. (* new. ok ? *)
```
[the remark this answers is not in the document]
mkerjean: Indeed, thanks.

theories/normedtype.v
mkerjean: By almost everywhere, do you mean unification hints between `numfieldType` (and its restructions) and pointed/filtered/etc ?

theories/normedtype.v
pi8027: The above `numfield_lmodType` is constructed using `regular_lmodType`, but the unification hint synthesized from it is `Num.NumField.sort <- GRing.Lmodule.sort`. So it makes `numFieldType` inheriting from `lmodType` by construction.

theories/normedtype.v
pi8027: I meant almost every new canonical declarations (but actually they are not that many). But `numfield_lmodType` below has the same issue.

theories/normedtype.v
pi8027: As in Cyril's [comment](https://github.com/math-comp/analysis/pull/205#issuecomment-634182427) above, `numFieldType` should inherit from `fieldExtType` (by construction) after this PR. By transitivity, the structures (A) should inherit from all the structures (B) below.
- (A): `numFieldType`, `numClosedFieldType`, `realFieldType`, `archiFieldType`, `rcfType`, and `realType`,
- (B): `lmodType`, `lalgType`, `algType`, `comAlgType`, `unitAlgType`, `comUnitAlgType`, `vectType`, `falgType`, and `fieldExtType`.

So we need the following 6 × 9 = 54 unification hints for them.
```coq
Canonical numField_lmodType (K : numFieldType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical numField_lalgType (K : numFieldType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical numField_algType (K : numFieldType) :=
  [algType K of K for [algType K of K^o]].
Canonical numField_comAlgType (K : numFieldType) := [comAlgType K of K].
Canonical numField_unitAlgType (K : numFieldType) := [unitAlgType K of K].
Canonical numField_comUnitAlgType (K : numFieldType) := [comUnitAlgType K of K].
Canonical numField_vectType (K : numFieldType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical numField_falgType (K : numFieldType) := [FalgType K of K].
Canonical numField_fieldExtType (K : numFieldType) :=
  [fieldExtType K of K for regular_fieldExtType K].

Canonical numClosedField_lmodType (K : numClosedFieldType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical numClosedField_lalgType (K : numClosedFieldType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical numClosedField_algType (K : numClosedFieldType) :=
  [algType K of K for [algType K of K^o]].
Canonical numClosedField_comAlgType (K : numClosedFieldType) :=
  [comAlgType K of K].
Canonical numClosedField_unitAlgType (K : numClosedFieldType) :=
  [unitAlgType K of K].
Canonical numClosedField_comUnitAlgType (K : numClosedFieldType) :=
  [comUnitAlgType K of K].
Canonical numClosedField_vectType (K : numClosedFieldType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical numClosedField_falgType (K : numClosedFieldType) := [FalgType K of K].
Canonical numClosedField_fieldExtType (K : numClosedFieldType) :=
  [fieldExtType K of K for regular_fieldExtType K].

Canonical realField_lmodType (K : realFieldType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical realField_lalgType (K : realFieldType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical realField_algType (K : realFieldType) :=
  [algType K of K for [algType K of K^o]].
Canonical realField_comAlgType (K : realFieldType) := [comAlgType K of K].
Canonical realField_unitAlgType (K : realFieldType) := [unitAlgType K of K].
Canonical realField_comUnitAlgType (K : realFieldType) :=
  [comUnitAlgType K of K].
Canonical realField_vectType (K : realFieldType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical realField_falgType (K : realFieldType) := [FalgType K of K].
Canonical realField_fieldExtType (K : realFieldType) :=
  [fieldExtType K of K for regular_fieldExtType K].

Canonical archiField_lmodType (K : archiFieldType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical archiField_lalgType (K : archiFieldType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical archiField_algType (K : archiFieldType) :=
  [algType K of K for [algType K of K^o]].
Canonical archiField_comAlgType (K : archiFieldType) := [comAlgType K of K].
Canonical archiField_unitAlgType (K : archiFieldType) := [unitAlgType K of K].
Canonical archiField_comUnitAlgType (K : archiFieldType) :=
  [comUnitAlgType K of K].
Canonical archiField_vectType (K : archiFieldType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical archiField_falgType (K : archiFieldType) := [FalgType K of K].
Canonical archiField_fieldExtType (K : archiFieldType) :=
  [fieldExtType K of K for regular_fieldExtType K].

Canonical rcf_lmodType (K : rcfType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical rcf_lalgType (K : rcfType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical rcf_algType (K : rcfType) := [algType K of K for [algType K of K^o]].
Canonical rcf_comAlgType (K : rcfType) := [comAlgType K of K].
Canonical rcf_unitAlgType (K : rcfType) := [unitAlgType K of K].
Canonical rcf_comUnitAlgType (K : rcfType) := [comUnitAlgType K of K].
Canonical rcf_vectType (K : rcfType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical rcf_falgType (K : rcfType) := [FalgType K of K].
Canonical rcf_fieldExtType (K : rcfType) :=
  [fieldExtType K of K for regular_fieldExtType K].

Canonical real_lmodType (K : realType) :=
  [lmodType K of K for [lmodType K of K^o]].
Canonical real_lalgType (K : realType) :=
  [lalgType K of K for [lalgType K of K^o]].
Canonical real_algType (K : realType) :=
  [algType K of K for [algType K of K^o]].
Canonical real_comAlgType (K : realType) := [comAlgType K of K].
Canonical real_unitAlgType (K : realType) := [unitAlgType K of K].
Canonical real_comUnitAlgType (K : realType) := [comUnitAlgType K of K].
Canonical real_vectType (K : realType) :=
  [vectType K of K for [vectType K of K^o]].
Canonical real_falgType (K : realType) := [FalgType K of K].
Canonical real_fieldExtType (K : realType) :=
  [fieldExtType K of K for regular_fieldExtType K].
```

pi8027: I think it would be better to declare these canonicals also as implicit coercions, and if Coq reports "ambiguous paths", we should stop coding and discuss.

CohenCyril: @pi8027 spot on!

CohenCyril: > I would like not to rely on the canonical ^o instances, as I'm afraid this will introduce the usual confusion on the lmodType structure of K^o.

In the sentence `[lmodType K of X for Y]`, `Y` is only used to get the mixin and the canonical structure is put on `X`. So using `^o` in `Y` leverages what we already know about `^o` to endow `X` with it.

CohenCyril: @pi8027 
> ```coq
> Canonical numField_fieldExtType (K : numFieldType) :=
>   [fieldExtType K of K for regular_fieldExtType K].
> ```

I would rather not rely on the name of a canonical structure, doesn't `[fieldExtType K of K^o]` work? (i.e. isn't `regular_fieldExtType K` canonical?)

pi8027: > I would rather not rely on the name of a canonical definition, doesn't `[fieldExtType K of K^o]` work? (i.e. isn't `regular_fieldExtType K` canonical?)

No, it doesn't work. I think it is broken on the MathComp side. Let me check.

mkerjean: > > I would like not to rely on the canonical ^o instances, as I'm afraid this will introduce the usual confusion on the lmodType structure of K^o.
> 
> In the sentence `[lmodType K of X for Y]`, `Y` is only used to get the mixin and the canonical structure is put on `X`. So using `^o` in `Y` leverages what we already know about `^o` to endow `X` with it.

My goal i this PR was to suppress any use of `K^o` as a topological object. Is there any advantage in using `K^o` as an intermediate object ?

mkerjean: > I think it would be better to declare these canonicals also as implicit coercions, and if Coq reports "ambiguous paths", we should stop coding and discuss.

Won't this happen if we keep the constructions on `K^o` as they are ?

CohenCyril: > > > I would like not to rely on the canonical ^o instances, as I'm afraid this will introduce the usual confusion on the lmodType structure of K^o.
> > 
> > 
> > In the sentence `[lmodType K of X for Y]`, `Y` is only used to get the mixin and the canonical structure is put on `X`. So using `^o` in `Y` leverages what we already know about `^o` to endow `X` with it.
> 
> My goal i this PR was to suppress any use of `K^o` as a topological object. Is there any advantage in using `K^o` as an intermediate object ?

Yes, as Kazuhuiko demonstrated, most of the work has already done for `^o` so using it as an intermediate object saves time and proofs.

mkerjean: Except from time, of course. It seems to me that using `K^o` as an intermediate object makes the file less readable and accessible to new users. I would be willing to rewrite everything on numFields directly.

CohenCyril: > Except from time, of course. It seems to me that using `K^o` as an intermediate object makes the file less readable and accessible to new users. I would be willing to rewrite everything on numFields directly.

Not only time but also maintainability (= robustness to change + number of loc to update). One should not write twice the same piece of code.
Instead I advocate for a comment that explains what these lines are doing.

mkerjean: Indeed, and this is why my intention was to remove anything related to `^o` in the mathcomp/analysis setting. Would you be against that ?

CohenCyril: We cannot really remove `^o` since it is in core mathcomp and useful there. Actually, all the results about `^o` from @pi8027's snippet come from mathcomp core. There is nothing we can do in mathcomp/analysis to get rid of them, so we should rather reuse them (for free).

mkerjean: Isn't there a risk that defining the topology using the `^o` would still leads to an iteration of the `^o` in the types of objects handled ?

pi8027: Using `^o` as an intermediate object to declare these canonicals makes sure that they are convertible with the "regular" instances, and, FYI, with PR math-comp/math-comp#545, we can write:
``` coq
Canonical numField_fieldExtType (K : numFieldType) := [fieldExtType K of K].
```

theories/normedtype.v
```diff
@@ -260,6 +260,253 @@ Definition pseudoMetric_of_normedDomain
...
+Canonical real_vectType (K : realType) :=
+  [vectType K of K for [vectType K of K^o]].
+Canonical real_falgType (K : realType) := [FalgType K of K].
+Canonical real_fieldExtType (K : realType) :=
+  [fieldExtType K of K for regular_fieldExtType K].
+
+
+Coercion numField_lmodType : numFieldType >-> lmodType.
+Coercion numField_lalgType : numFieldType >-> lalgType.
+Coercion numField_comAlgType : numFieldType >-> comAlgType. (* ambiguous path *)
```
pi8027, line 367: So we have the following ambiguous paths here:
```
New coercion path [numField_comAlgType; GRing.ComAlgebra.comRingType] : Num.NumField.type >-> GRing.ComRing.type is ambiguous with existing 
[Num.NumField.comRingType] : Num.NumField.type >-> GRing.ComRing.type. [ambiguous-paths,typechecker]
```

I think the problem is twofold:
1. `regular_comRingType` uses `GRing.mulrC`, which is opaque.
2. Since `lalgType` and its subclasses do not take the ring structures as their base, an eta expansion appears in the former path.

pi8027: FYI, I'm working on a fix of this issue on the MathComp side, which subsumes math-comp/math-comp#545 and math-comp/math-comp#546.

pi8027: Now math-comp/math-comp#546 solves this issue. Adding 54 implicit coercions from `numFieldType` and its subclasses to `fieldExtType` and its superclasses does not introduce any ambiguous path.

pi8027:
PR math-comp/math-comp#591 reimplements the solution to the issue of ambiguous paths (math-comp/math-comp#546) using primitive class records. I suggest using the new one to work on this PR. If there is a problem, please let me know.

theories/normedtype.v
mkerjean: > PR [math-comp/math-comp#591](https://github.com/math-comp/math-comp/pull/591) reimplements the solution to the issue of ambiguous paths ([math-comp/math-comp#546](https://github.com/math-comp/math-comp/pull/546)) using primitive class records. I suggest using the new one to work on this PR. If there is a problem, please let me know.

Since PR591 has been merged, I am trying to use the master branch of mathcomp to work on the present PR205, but real-closed doesn't seem to compile over mathcomp-master. Do you have any idea why ? 

```
File "./theories/polyrcf.v", line 228, characters 0-63:
Error: The LHS of leNgt
    (_ <= _)%O
does not match any subterm of the goal
```

pi8027: math-comp/real-closed#28 is compatible with the master branch of MathComp.

mkerjean: Is the master branch of analysis supposed to be compatible to be compatible with the master branch of analysis ? I'm having trouble rebasing section [numfiled_topological](https://github.com/math-comp/analysis/blob/0ab9e13f9ce156aea29ebc48b1f66103bb1d54e2/theories/normedtype.v#L391) but I'm also stumbling on issues like that : 

```
File "./theories/normedtype.v", line 525, characters 2-76:
Error: The LHS of ler_paddr
    (_ <= _ + _)
does not match any subterm of the goal
```

mkerjean: @pi8027 Since the rebasing of analysis on the master branch of math-comp is failing at the moment, I would like to rebase directly the present PR on your local solutions, but the branch merged in [math-comp/math-comp#591](https://github.com/math-comp/math-comp/pull/591) doesn't seems to be available anymore. Where can I find it ?

pi8027:
@mkerjean Do you need any help? I would like to take a look at this PR after merging #275.

mkerjean:
> @mkerjean Do you need any help? I would like to take a look at this PR after merging #275.

Yes ! I think this PR begins to look better. I moved your [previous](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/topology.v#L4329) definitions at the end of `topology.v` , as they are now needed by `ereal.v`. 

I would be especially happy to have your opinion on [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) new definitions : should normedModType structures on `rcfType`, `realType` and others be defined this way ? Shoud we define them without a clone of the structure using `_^o` ?

mkerjean:
@affeldt-aist The way I did it, taking `^o` out of `ereal.v` as the unfortunate consequence that `ereal_hausdorff` needs to be explicitely used sometimes ([here](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/measure.v#L312) for example). Do you see why ?

affeldt-aist:
> @affeldt-aist The way I did it, taking `^o` out of `ereal.v` as the unfortunate consequence that `ereal_hausdorff` needs to be explicitely used sometimes ([here](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/measure.v#L312) for example). Do you see why ?

It seems to be fixable by changing the `Hint`.

theories/topology.v
```diff
@@ -4327,26 +4338,202 @@ Qed.
...
+     (Uniform.class (numFieldType_uniformType R)) (@pseudoMetric_of_normedDomain R R)).
+(* (*NEW*) *)
+
+Canonical numClosedFieldType_pointedType (R: numClosedFieldType) :=
+  [pointedType of R for [pointedType of [numFieldType of R]]].
+Canonical numClosedFieldType_filteredType (R: numClosedFieldType) :=
+  [filteredType _ of R for [filteredType _ of [numFieldType of R]]].
+Canonical numClosedFieldType_topologicalType (R: numClosedFieldType) : topologicalType :=
+  (* [topologicalType of R for [topologicalType of [numFieldType of R]]]. *)
+  (*leads to ambiguous path *)
```
pi8027, line 4366: With MathComp 1.12, I do not observe this issue of ambiguous paths. See https://github.com/pi8027/analysis/tree/numfield_topology

theories/normedtype.v
```diff
@@ -96,10 +97,366 @@ End add_to_mathcomp.
...
+      (@nbhs_ball_normE _ [normedZmodType R of R])
+      (pseudoMetric_of_normedDomain [normedZmodType R of R]))).
+Canonical numFieldType_regular_uniformType : uniformType := UniformType R^o
+  (uniformityOfBallMixin (@nbhs_ball_normE _ [normedZmodType R of R])
+    (pseudoMetric_of_normedDomain [normedZmodType R of R])).
+Canonical numFieldType_regular_pseudoMetricType :=
+  @PseudoMetric.Pack R R^o (@PseudoMetric.Class R R
+  (Uniform.class (numFieldType_uniformType R)) (@pseudoMetric_of_normedDomain R R)).
+Definition numdFieldType_lalgType : lalgType R := @GRing.regular_lalgType R.
+End numFieldType_regular_canonical.
```
pi8027, line 452: They seem redundant with canonical declarations in topology.v, and can be removed if I'm not mistaken. (Also, those declarations in topology.v can probably be removed.)

mkerjean: Yes they are redundant, but they were put in `topology.v` because `ereal.v` need them, and it only depends of `topology.v`. So they should be removed from `normedtype.v` if I'm right ?

pi8027:
FTR, `sequences.v` still has some remaining `^o`.

mkerjean:
> FTR, `sequences.v` still has some remaining `^o`.

I'll try to do that today, thanks

theories/normedtype.v
mkerjean: I misunderstood your comment: you are right, these lines should be removed both from `topology.v` and `normedtype.v`. However, these is still an issue in  `sequence.v` concerning `harmonic` on which I am working, and for now the `Canonical numFieldType_regular_topologicalType ` is needed to make sequence.v compile.

CohenCyril:
Left todo before merge :
- [ ] encapsulate "non-forgetful inheritance" instances inside a module (e.g `NumFieldNormed`) and document it in the headers and import the module in the subsequent files of the library,
- [ ] add a `CHANGELOG_UNRELEASED.md` entry,
- [ ] squash & rebase on top of master

theories/sequences.v
```diff
@@ -485,17 +484,17 @@ Qed.
...
 Lemma harmonic_gt0 {R : numFieldType} i : 0 < harmonic i :> R.
 Proof. by rewrite /= invr_gt0 ltr0n. Qed.
 
 Lemma harmonic_ge0 {R : numFieldType} i : 0 <= harmonic i :> R.
 Proof. exact/ltW/harmonic_gt0. Qed.
 
-Lemma cvg_harmonic {R : archiFieldType} : harmonic --> (0 : R^o).
-Proof.
+Lemma cvg_harmonic {R : archiFieldType} : (@harmonic R) --> (0 : R).
+Proof. (*TODO : remove explicit R for harmonic*)
```
mkerjean, line 497: Do you understand why the argument of harmonic needs to be explicit now @pi8027 @affeldt-aist ? How would the new topological definitions affect this ?

pi8027: Because a `fieldType` is not canonically a `filteredType`? The error message seems to make sense, but I don't get how to fix this.
```coq
Error:
In environment
R : Num.ArchimedeanField.type
The term "Phantom (sequence (GRing.Field.sort ?R)) (@harmonic ?R)" has type
 "phantom (sequence (GRing.Field.sort ?R)) (@harmonic ?R)" while it is expected to have type
 "phantom (@Filtered.sort ?X ?fX) ?x".
```

pi8027: By annotating the type of `harmonic` by `^o`, we were able to specialize its argument `R` to a `numFieldType` when it gets unified with a `filteredType`. So keeping `^o` for `harmonic` may make sense. Specializing `harmonic` to `R : numFieldType` also solves this inference issue, but it may break something else.

theories/normedtype.v
pi8027: If I remember/understand correctly what @CohenCyril suggested in the last analysis call, we should actually keep them somewhere to allow users to choose the inference style: annotating things with `^o`, or using "non-forgetful inheritance" which this PR provides.

pi8027:
FYI, I intend to modify `hierarchy.ml` at some point to make sure there is no missing join in this PR, but I'm not sure if it is doable.

theories/sequences.v
mkerjean: Specializing `harmonic` to numFieldType has the consequence that one must make explicit `R` when applying `cvg_harmonic` later.

pi8027: So, I guess it is better not to change the definition of `harmonic`, at least for now.

mkerjean:
> FYI, I intend to modify `hierarchy.ml` at some point to make sure there is no missing join in this PR, but I'm not sure if it is doable.

That woudl be great! I'm afraid that [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) definitions should be done directly one the hierarchy of topological structures. Do you want to merge already the cleanup you did in your fork of the PR ?

pi8027:
> That woudl be great! I'm afraid that [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) definitions should be done directly one the hierarchy of topological structures.

I'm not sure if I understood what you say, but is it the same as the following one?:

> I would be especially happy to have your opinion on [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) new definitions : should normedModType structures on `rcfType`, `realType` and others be defined this way ? Shoud we define them without a clone of the structure using `_^o` ?

OK, I actually did not answer this. I think we should define a `regular` instance for each topological structure (pointed, filtered, topological, normed modules, etc.), but not for each numeric field structure (`rcfType`, `realType`, etc.), and then define instances for non-forgetful inheritance based on them. These `regular` instances solve unification problems of the form `Pointed.sort ?1 = regular ?2`, `Filtered.sort ?1 = regular ?2`, and so on, and then force `?2` to be canonically a `numFieldType`. Since other numeric field structures are canonically `numFieldType`, we do not need `regular` instances for them. Also, if we define such `regular` instances, they will be redundant with ones for `numFieldType`.

> Do you want to merge already the cleanup you did in your fork of the PR ?

No (I mean, not now). I'd like to wait for #275.

mkerjean:
> > That woudl be great! I'm afraid that [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) definitions should be done directly one the hierarchy of topological structures.
> 
> I'm not sure if I understood what you say, but is it the same as the following one?:
> 
> > I would be especially happy to have your opinion on [these](https://github.com/math-comp/analysis/blob/b3a67378443dc35cd788b02dc384a5f3a4949935/theories/normedtype.v#L1258) new definitions : should normedModType structures on `rcfType`, `realType` and others be defined this way ? Shoud we define them without a clone of the structure using `_^o` ?
> 
> OK, I actually did not answer this. I think we should define a `regular` instance for each topological structure (pointed, filtered, topological, normed modules, etc.), but not for each numeric field structure (`rcfType`, `realType`, etc.), and then define instances for non-forgetful inheritance based on them. These `regular` instances solve unification problems of the form `Pointed.sort ?1 = regular ?2`, `Filtered.sort ?1 = regular ?2`, and so on, and then force `?2` to be canonically a `numFieldType`. Since other numeric field structures are canonically `numFieldType`, we do not need `regular` instances for them. Also, if we define such `regular` instances, they will be redundant with ones for `numFieldType`.
> 
Indeed, it was the same question. Thanks you for explanations. So we are ok for now, and  wait for PR#275 to be merged to rebase.

> > Do you want to merge already the cleanup you did in your fork of the PR ?
> 
> No (I mean, not now). I'd like to wait for #275.

theories/sequences.v
pi8027: Now I think this is an issue of missing join: `numFieldType` should be the join of `fieldType` and `filteredType`, in other words, a unification problem `GRing.Field.sort ?1 = Filtered.sort ?2` has to be solved by instantiating `?1` and `?2` with a fresh `?3 : numFieldType`.

mkerjean: This should resolve harmonic problem right ? I still struggle with joins, would it be something like that : 
```
Definition filtered_fieldType  (R : numFieldType)
  := @GRing.Field.Pack (numFieldType_filteredType R) (Num.NumField.class R).
```

pi8027: @mkerjean Yes. If you keep this part as is, I will take care of this issue.

pi8027:
I'm working on a fix for this PR. I suggest not to touch this branch until further notice so that we do not duplicate our efforts.

theories/sequences.v
mkerjean: I  won't work on this PR for at least two weeks, and not until you tell me it is safe to do so.

pi8027:
I still have to fix canonical declarations, but I do not see any serious unfixable issues. Here is the diagram:
hierarchy

pi8027:
FTR, removing this line breaks the `ex_ball_sig` lemma in `normedtype.v`. So, at least `numDomainType` has to inherit from `pointedType`. https://github.com/math-comp/analysis/blob/823ad8d21b4305b111152e59c3fe036d49a8d960/theories/topology.v#L3902

pi8027:
I made some progress. See: https://github.com/pi8027/analysis/tree/numfield_topology.

Here is the list of remaining issues I discovered:
- Currently, the smallest `ssralg`/`ssrnum` structure that inherits from `pointedType` is `numFieldType`, and `R_pointedType` has been removed. So I had to put `^o` back in two places: https://github.com/math-comp/analysis/blob/e954498980cca73b34f32b5ca86c92b6973479ec/theories/normedtype.v#L301 https://github.com/math-comp/analysis/blob/e954498980cca73b34f32b5ca86c92b6973479ec/theories/normedtype.v#L322 I'd like to hear @CohenCyril's opinion on whether we should make `zmodType` inheriting from `pointedType` to fix this or not. I think that a downside to doing so is that the `countalg` and `finalg` structures also have to inherit from `pointedType` to make the hierarchy globally consistent.
- I had to patch the proof of `cvg_norm` but do not understand why. https://github.com/math-comp/analysis/commit/e954498980cca73b34f32b5ca86c92b6973479ec#diff-ec3855f35a085739515162cb2620b5daea6488b737c11df25d618370adc5e240R2595-R2596
- There is only one remaining explicit application of `harmonic`. I think this is unfixable. https://github.com/math-comp/analysis/blob/e954498980cca73b34f32b5ca86c92b6973479ec/theories/sequences.v#L693

mkerjean:
> I made some progress. See: https://github.com/pi8027/analysis/tree/numfield_topology.
Could you describe the changes you made to  `normedtype.v` in this branch, in particular concerning the module `nonforgetful_inheritance`? Is it mainly a reordering of the deleted canonical definitions, or are they new structures at stakes ?

mkerjean:
> I made some progress. See: https://github.com/pi8027/analysis/tree/numfield_topology.

Could you describe the changes you made to `normedtype.v` in this branch, in particular concerning the module `nonforgetful_inheritance` ? Is it mainly a reordering of the deleted canonical definitions, or are they new structures at stakes ?

pi8027:
> > I made some progress. See: https://github.com/pi8027/analysis/tree/numfield_topology.
> 
> Could you describe the changes you made to `normedtype.v` in this branch, in particular concerning the module `nonforgetful_inheritance` ? Is it mainly a reordering of the deleted canonical definitions, or are they new structures at stakes ?

@mkerjean Since Coq gives precedence to the oldest coercion path between two coercion classes, the coercion from `realType` (and other structures that inherit from `numFieldType`) to `normedModType` (and `pseudoMetricNormedZmodType`) has to be declared before one from `numFieldType` to `normedModType`. So I reordered them upside down (from `realType` to `numFieldType` ordering). A bunch of missing (non-trivial) joins have also been added, e.g., the join of `numDomainType` and `normedModType` should be `numFieldType`.

pi8027:
I will unlikely be able to make more progress until Feb 2. Also, I need feedback from Cyril to make more progress anyway. (No pressure intended.)

mkerjean:
> I will unlikely be able to make more progress until Feb 2. Also, I need feedback from Cyril to make more progress anyway. (No pressure intended.)

The status of this PR could be discussed in the next Analysis meeting. I am waiting for it to be merged to rebase the PRs related to Hahn Banach, Banach Steinhauss and holomorpy on it.

CohenCyril:
Ah sorry, I've been kind of busy...
The problem is that forgetful inheritance from mathcomp analysis to mathcomp (core) is not possible in the current state of affairs, and altering mathcomp with anything that might lead to to blowup in Galois is risky, so I'm afraid we cannot do a proper inheritance between `numDomainType` and `pointedType` and we'll have to go to the non forgetful one, unless I missed something...

mkerjean:
Did you had a chance to merge your fork of the PR @pi8027  ? Should this branch be merged with master afterwards @CohenCyril  ? See the [todos](https://github.com/math-comp/analysis/wiki/2021-01-26-Meeting) from the last meeting.

affeldt-aist:
small questions about this PR:
- should `default.nix` be changed? (l.49 mentions `numfield_topology` which is this very branch) @CohenCyril 
- in `ereal.v`, `lee_fin`, `lte_fin`, `lee_tofin`, `lte_tofin` should use `%O` or` %R`? @pi8027

affeldt-aist:
I tried to reflect the changes in the changelog but I am not sure it is good enough. You may want to take a look at it.

affeldt-aist:
@pi8027 maybe we should put a date to the diagram to avoid any confusion, or do you have a more recent one?

pi8027:
> in `ereal.v`, `lee_fin`, `lte_fin`, `lee_tofin`, `lte_tofin` should use `%O` or` %R`? @pi8027

The LHS and RHS of `lee_fin` are inequations of extended reals and a numeric domain, and should use `ereal_scope` and `ring_scope`, respectively. However, the latter one should be `order_scope` if we generalize it to `porderType`.

> @pi8027 maybe we should put a date to the diagram to avoid any confusion, or do you have a more recent one?

https://github.com/math-comp/analysis/pull/205#issuecomment-749539719 is up to date.

pi8027:
In the case you need the dot file, here it is: [hierarchy.dot](https://github.com/math-comp/analysis/files/6127770/hierarchy.txt).

affeldt-aist:
> The LHS and RHS of `lee_fin` are inequations of extended reals and a numeric domain, and should use `ereal_scope` and `ring_scope`, respectively. However, the latter one should be `order_scope` if we generalize it to `porderType`.

I propose to create an issue for that and to merge this PR now. What do you think? @CohenCyril @pi8027 @mkerjean

pi8027:
@affeldt-aist The module `nonforgetful_inheritance` has to be renamed to `numFieldTopology` according to the last meeting.

pi8027 (approved):
Fine with me.

PR #206, opened 2020-05-18
Add a distrLattice canonical structure to set T.

vlj:
Hi,

This add a canonical distrLattice structure to setT and subset relation.
I have far less experience in adding a canonical structure than I have writing lemma so the PR is likely a bit rough, any guidance is appreciated.

Some comments:
- For some reasons coq fails to unify the type of the lemma setU setI ... when passed as argument to MeetJoinMixin. I added "wrapper lemma" SetOrder_setCI SetOrder_setCU ... to force the type. I'm not sure this is the right thing to do (I could also use @setU (set T) directly in the argument but it would be quite hard to read) and the naming is maybe wrong.
- subset is a Prop while distrLattice expects a rel (ie result type is bool). I wrapped the subset relation in asboolsubset. This means that to use lemma in order (like in my case ltux) one need to map subset and asboolsubset forth and back so I provided subsetE. I don't know if it's the usual idiom.

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -11,6 +11,8 @@
     `setDUr`, `setDUl`, `setIDA`, `setDD`
   + definition `dep_arrow_choiceType`
   + lemma `bigcup_set0`
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`
+  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`
 - in `normedtype.v`:
   + lemma `closed_ereal_le_ereal`
   + lemma `closed_ereal_ge_ereal`
@@ -47,6 +49,9 @@
   + lemma `oppe_continuous`
   + lemmas `setUCl`, `setDv`
   + lemmas `image_preimage_subset`, `image_subset`, `preimage_subset`
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`, `setDset0`, `subsetNeqSetDnonempty`, `proper_neqAsubset`
+  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`.
 
 ### Changed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -13,6 +13,8 @@
   + lemma `bigcup_set0`
   + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`
   + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`
+  + lemma `not_exists2P`
+
 - in `normedtype.v`:
   + lemma `closed_ereal_le_ereal`
   + lemma `closed_ereal_ge_ereal`
@@ -49,9 +51,10 @@
   + lemma `oppe_continuous`
   + lemmas `setUCl`, `setDv`
   + lemmas `image_preimage_subset`, `image_subset`, `preimage_subset`
+  + definition `proper` and its notation `<`
   + lemmas `setUK`, `setKU`, `setIK`, `setKI`
-  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`, `setDset0`, `subsetNeqSetDnonempty`, `proper_neqAsubset`
-  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`.
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`, `properEneq`
+  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType` on classical `set`.
 
 ### Changed
 
@@ -59,6 +62,7 @@
   + the index in `bigcup_set1` generalized from `nat` to some `Type`
   + lemma `bigcapCU` generalized
   + lemmas `preimage_setU` and `preimage_setI` are about the `setU` and `setI` (instead of `bigcup` and `bigcap`)
+  + `eqEsubset` changed from an implication to an equality
 - lemma `asboolb` moved from `discrete.v` to `boolp.v`
 - lemma `exists2NP` moved from `discrete.v` to `boolp.v`
 - lemma `neg_or` moved from `discrete.v` to `boolp.v` and renamed to `not_orP`
@@ -72,6 +76,8 @@
   + the first argument of `real_of_er` is now maximal implicit
   + the first argument of `is_real` is now maximal implicit
   + generalization of `lee_sum`
+- in `boolp.v`:
+  + rename `exists2NP` to `forall2NP` and make it bidirectionnal
 
 ### Renamed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -11,7 +11,7 @@
     `setDUr`, `setDUl`, `setIDA`, `setDD`
   + definition `dep_arrow_choiceType`
   + lemma `bigcup_set0`
-  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`, `subsetPset`, `properPset`
   + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`
   + lemma `not_exists2P`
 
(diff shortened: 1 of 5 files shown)


CohenCyril:
Hi @vlj thanks for your contribution, I will take a look a deeper look next week.

CohenCyril:
@vlj  thanks for your contribution, I think once the proofs are tidied we can merge, indeed proofs in math-comp/analysis should adhere to [math-comp core contributing style](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md) and more particularily [the proof style](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md#proof-style).
In particular:
- The only period symbol (`.`) in a line should be the last character of the line.
- When exactly two subgoals are generated, the first one should be indented and second one should not. No bullets should be used.
Sorry for the delay, and please do not hesitate to ask questions, I will try to be more reactive now!

CohenCyril:
> About the first convention is it accepted to use the semicolon symbol even if only a single goal is focused?

Yes, in mathcomp proof style conventions, the semicolon is not used regardless of how many goals are focused, and its purpose is to group tactics in a line that makes a single, substantial step of reasoning.

theories/classical_sets.v
```diff
@@ -980,3 +980,86 @@ Definition supremum (x0 : T) E :=
...
+Definition asboolsubset X Y := `[< X `<=`Y >].
+
+Definition proper_asboolsubset X Y := ~~(Y == X) && asboolsubset X Y.
+
+Lemma setI_meet X Y :
+  asboolsubset X Y = (setI X Y == X).
+Proof.
+apply/asbool_equiv_eqP.
+  apply/eqP.
+split.
```
CohenCyril, line 999: These three last lines should be regrouped in one single line.

CohenCyril:
Hi @vlj, I refactored and cleaned your code. The wrapper lemmas you defined were not necessary if you partially apply the wanted lemmas. Some intermediate definitions were superfluous. About the `bool` vs `Prop` thing, it is indeed expected that lattice relations are in `bool` while the classic subset relation is in `Prop`. Indeed `asbool` was needed but should be used minimally.
Some pieces of code are still missing to turn sets into a full-fledged distributive lattice with top and bottom and complement. We could merge as such and open a new PR to complement it...
From the naming of your (former) wrapper lemmas, I also suspect that many of our set lemmas in `classical_sets.v` are not following the naming conventions from `finset.v` or `order.v` (also cf #252), hence one should open yet another PR to fix them. (ping @affeldt-aist)

affeldt-aist:
`setIDl` was renamed to `setIUl` by a recent PR (hence the rebase/push force).
`classical_sets.v` was lacking `setUK`, etc. lemmas.
Should `setI_meet` be renamed `setIidPlE` and moved near `setIidPl`?

vlj:
Thanks for the refactoring and renaming !
I added the canonical structure for top and bottom. I didn't manage to get Order.BLattice.Mixin to accept the sub0set lemma so I introduced an intermediate fact that uses the asboolP. There is no counterpart for the setTsub so I directly write the one that fits the input of Order.TBLattice.Mixin.

vlj:
I added complement and sub.

CohenCyril (changes requested):
The `Exports` module misses a few canonicals: `Internal.blatticeType` ... `Internal.ctbDistrLatticeType`. Also, I think `subEset` should be renamed `subsetEset`, and you should add as well ``properEset : (X < Y)%O = (X `<` Y)``, ``subEset : (X `\` Y)%O = (X `\` Y)``, ``complEset : (~` X)%O = ~` X``, ``botEset : 0%O = set0`` , ``topEset : 1%O = setT``.

CohenCyril:
Additionally, the changelog should document the new canonical instances and the `E` lemmas.

vlj:
I merged and added the lemmas suggested.

CohenCyril:
Great! Can you squash and rebase?

CHANGELOG_UNRELEASED.md
```diff
@@ -9,6 +9,8 @@
 - in `classical_sets.v`:
   + lemmas `setIIl`, `setIIr`, `setCS`, `setSD`, `setDS`, `setDSS`, `setCI`,
     `setDUr`, `setDUl`, `setIDA`, `setDD`
+  + lemmas `setUK`, `setKU`, `setIK`, `setKI`, `subsetEset`, `subEset`, `complEset`, `botEset`, `topEset`, `meetEset`, `joinEset`
+  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType`.
```
CohenCyril, line 13:
```suggestion
  + Canonical `porderType`, `latticeType`, `distrLatticeType`, `blatticeType`, `tblatticeType`, `bDistrLatticeType`, `tbDistrLatticeType`, `cbDistrLatticeType`, `ctbDistrLatticeType` on classical `set`
```

theories/classical_sets.v
```diff
@@ -238,8 +239,10 @@ Notation "\bigcap_ ( i : T ) F" :=
 Notation "\bigcap_ i F" := (\bigcap_(i : _) F) : classical_set_scope.
 
 Definition subset {A} (X Y : set A) := forall a, X a -> Y a.
+Definition proper_subset {A} (X Y : set A) := (Y != X) && `[< subset X Y >].
```
CohenCyril, line 242:
```suggestion
Definition proper_subset {A} (X Y : set A) := (Y != X)  /\ X `<=` Y.
```

theories/classical_sets.v
```diff
@@ -238,8 +239,11 @@ Notation "\bigcap_ ( i : T ) F" :=
 Notation "\bigcap_ i F" := (\bigcap_(i : _) F) : classical_set_scope.
 
 Definition subset {A} (X Y : set A) := forall a, X a -> Y a.
-
 Notation "A `<=` B" := (subset A B) : classical_set_scope.
+
+Definition proper_subset {A} (X Y : set A) := (Y != X) /\ X `<=` Y.
```
CohenCyril, line 244: Actually, on second thoughts, this definition would be even more usable like this:
```suggestion
Definition proper_subset {A} (X Y : set A) := (X `<=` Y) /\ (exists2 a, Y a & ~ X a).
```
With lemma ``proper_neqAsubset :  (X `<` Y) =  (X `!=` Y) /\  (X `<=` Y)``

vlj:
Working on it, but it's WIP.

vlj:
Is mathcomp analysis supposed to be built against dev branch of mathcomp ? I need the contra_not_neq lemma which seems to have been introduced recently in ssrbool.

CohenCyril:
Indeed, @vlj we build master against stable releases, you may add it in `boolp.v` as long as you open an issue after the merge of the PR to remind us to remove it as soon as we switch to mathcomp 1.12 when it is release

vlj:
I modified the proper_subset definition. I had some troubles writing the properEset and I'm not sure it's the easiest way to do ; it uses a couple intermediate lemma, setDset0 :  (Y `\` X) != set0 <-> exists2 a : T, Y a & ~ X a. and subsetNeqSetDnonempty (H : X `<=` Y) : X != Y <-> (Y `\` X) != set0. I think setDset0 may be useful for other purpose, I don't know if subsetNeqSetDnonempty  is too specific or not. I'm also not very experimented with the asbool_* lemmas so maybe it's not optimal either.

vlj:
Added proper_neqAsubset lemma.

affeldt-aist:
> The `Exports` module misses a few canonicals: `Internal.blatticeType` ... `Internal.ctbDistrLatticeType`. Also, I think `subEset` should be renamed `subsetEset`, and you should add as well `` properEset : (X < Y)%O = (X `<` Y) ``, `` subEset : (X `\` Y)%O = (X `\` Y) ``, `` complEset : (~` X)%O = ~` X ``, `botEset : 0%O = set0` , `topEset : 1%O = setT`.

It looks like these changes have been performed.

theories/classical_sets.v
affeldt-aist: I tried yet another definition, the one mimicking `fintype`. See the last commit.

pi8027:
@affeldt-aist Could you request me a review here just not to forget about it? Thanks.

affeldt-aist:
> @affeldt-aist Could you request me a review here just not to forget about it? Thanks.

Sure. Thank you!

theories/classical_sets.v
```diff
@@ -1136,3 +1167,90 @@ exists n.+1; split => // m Em; case/existsNP : En => k /not_implyP[Ek /negP].
...
+
+Fact SetOrder_joinKI (Y X : set T) : X `&` (X `|` Y) = X.
+Proof. by rewrite setUC setKU. Qed.
+
+Fact SetOrder_meetKU (Y X : set T) : X `|` (X `&` Y) = X.
+Proof. by rewrite setIC setKI. Qed.
+
+Definition orderMixin := @MeetJoinMixin _ _ _ setI setU setI_meet
+  (fun _ _ => erefl) (@setIC _) (@setUC _) (@setIA _) (@setUA _)
+  SetOrder_joinKI SetOrder_meetKU (@setIUl _) setIid.
```
pi8027, line 1191: I think it is better to make `Order.lt` convertible with `(fun X Y => `[<X `<` Y>])` here. The third argument of `MeetJoinMixin` has to be explicitly provided and `(fun _ _ => erefl)` has also to be replaced.

theories/classical_sets.v
```diff
@@ -1136,3 +1167,90 @@ exists n.+1; split => // m Em; case/existsNP : En => k /not_implyP[Ek /negP].
...
+Local Canonical latticeType := LatticeType (set T) orderMixin.
+Local Canonical distrLatticeType := DistrLatticeType (set T) orderMixin.
+
+Fact SetOrder_sub0set (x: set T): (set0 <= x)%O.
+Proof. apply/asboolP; apply: sub0set. Qed.
+
+Fact SetOrder_setTsub (x:set T): (x <= setT)%O.
+Proof. apply/asboolP; by []. Qed.
+
+Local Canonical blatticeType :=
```
pi8027, line 1203:
```suggestion
Local Canonical bLatticeType :=
```

theories/classical_sets.v
```diff
@@ -1136,3 +1167,90 @@ exists n.+1; split => // m Em; case/existsNP : En => k /not_implyP[Ek /negP].
...
+  BLatticeType (set T) (Order.BLattice.Mixin  SetOrder_sub0set).
+Local Canonical tblatticeType :=
```
pi8027, line 1205:
```suggestion
Local Canonical tbLatticeType :=
```

theories/classical_sets.v
```diff
@@ -243,15 +244,21 @@ Notation "\bigcap_ ( i : T ) F" :=
 Notation "\bigcap_ i F" := (\bigcap_(i : _) F) : classical_set_scope.
 
 Definition subset {A} (X Y : set A) := forall a, X a -> Y a.
-
 Notation "A `<=` B" := (subset A B) : classical_set_scope.
+
+Definition proper {A} (X Y : set A) := X `<=` Y /\ ~ (Y `<=` X).
+Notation "A `<` B" := (proper A B) : classical_set_scope.
```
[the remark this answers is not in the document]
affeldt-aist: Indeed. But maybe this should be another PR. Is it ok to raise an issue about that and do it later, or do you think it should be done now?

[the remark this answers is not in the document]
pi8027: It is OK to do so, but I think ssrmatching becomes less useful when we duplicate definitions like this, and then we often have to rewrite ones to others manually. Also, there is a substantial overlap between the order theory and `classical_sets.v`.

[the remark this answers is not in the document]
CohenCyril: @pi8027 in fact, it still makes a difference even in mathcomp analysis: because the propositional version is directly usable as an implication (for `subset`) or a conjunction of an implication (and a negation of one) for `proper`, we should have both, and there should be a reflection lemma between the two.
However, the big problem form is the naming, which conflicts entirely with finset, ... we should rename them all, but that's beyond the scope of this PR.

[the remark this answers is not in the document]
affeldt-aist: > there should be a reflection lemma between the two

Are the `subsetEset`, `properEset`, etc. lemmas sufficient for this purpose? (Provided we `Export SetOrder.Exports.`.)

CohenCyril (approved):
LGTM now!

PR #213, opened 2020-05-26
tentative how-to-release file

affeldt-aist:
This is the same kind of memo as MathComp has for release managers. It might be useful to double-check release steps.

### DIFF
--- a/HOWTORELEASE.md
+++ b/HOWTORELEASE.md
@@ -31,7 +31,7 @@
 1. Update the date of the opam file in master
 
 2. Submit a new opam package to [https://github.com/coq/opam-coq-archive](https://github.com/coq/opam-coq-archive)
-- details omitted (for now?)
+- details omitted for now.
 
 3. Annouce on Zulip:
 - "The opam package the X.Y.Z version of mathcomp-analysis is available as an opam package in U."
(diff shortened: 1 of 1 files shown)


affeldt-aist:
TODO: make the tag format precise. @CohenCyril

CohenCyril:
> TODO: make the tag format precise. @CohenCyril

done in b264352aaa016e4cdf06054f412af0018bb57fc5

CohenCyril:
@affeldt-aist should we merge this in the Wiki?

affeldt-aist:
> should we merge this in the Wiki?

You mean merge in master (in which case I am ok) or copy-paste in the wiki (in which case I would ask why not merge in master and put a link from the wiki to the .md file)?

CohenCyril:
> > should we merge this in the Wiki?
> 
> You mean merge in master (in which case I am ok) or copy-paste in the wiki (in which case I would ask why not merge in master and put a link from the wiki to the .md file)?

Since this is a release process tied to github and not to the distributed content, I think it should not be in math-comp/analysis, but math-comp/analysis.wiki....

affeldt-aist:
I moved the file to `analysis.wiki`. We can maybe close without merging?

PR #218, opened 2020-06-04
Adapting to new Order.max

CohenCyril:
Patch for math-comp/math-comp#516


affeldt-aist:
It looks like it is pulling 4b3281e45c570ca3ee21659b09e8ba8707f250d2.

PR #219, opened 2020-06-05
rename lemmas for classical reasoning to avoid clash with fintype

### LINKED ISSUE #217: Rename existsP
https://github.com/math-comp/analysis/blob/d2d010f71587ac48338213e3c7e652351baac503/theories/boolp.v#L564

Naming conflicts with `fintype.existsP`; the same for `forallP`.

affeldt-aist:
fixes #217


affeldt-aist:
Is it ok to merge quickly?

PR #223, opened 2020-06-09
Boole inequality

affeldt-aist:
Tentative formalization of Boole's inequality (wip).

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -6,7 +6,12 @@ Require Import boolp reals ereal.
 Require Import classical_sets posnum topology normedtype sequences.
 
 (******************************************************************************)
-(*          Tentative formalization of Boole's inequality (WIP)               *)
+(* This file provides basic elements of a theory of measure illustrated       *)
+(* by a formalization of Boole's inequality.                                  *)
+(*                                                                            *)
+(* {measure set T -> {ereal R}} == type of a measure over sets of elements of *)
+(*                                 type T where R is expected to be a         *)
+(*                                 a realFieldType or a realType              *)
 (******************************************************************************)
 
 Set Implicit Arguments.
@@ -21,7 +26,12 @@ Local Open Scope ring_scope.
 Lemma setCE T (A : set T) : ~` A = setT `\` A.
 Proof. by rewrite predeqE => t; split => // -[]. Qed.
 
-Lemma bigsetU_incl T m n (U : nat -> set T) : (m <= n)%N ->
+Lemma setUCr T (A : set T) : A `|` ~` A = setT.
+Proof.
+by rewrite predeqE => t; split => // _; case: (pselect (A t)); [left|right].
+Qed.
+
+Lemma subset_bigsetU T m n (U : nat -> set T) : (m <= n)%N ->
   \big[setU/set0]_(i < m) U i `<=` \big[setU/set0]_(i < n) U i.
 Proof.
 by rewrite !bigcup_ord => mn x [i im ?]; exists i => //; rewrite (leq_trans im).
@@ -65,25 +75,25 @@ End Measurable.
 Export Measurable.Exports.
 
 Section measurable_interface.
-Variables X : measurableType.
+Variables T : measurableType.
 
-Lemma measurable0 : measurable (set0 : set X).
-Proof. by case: X => ? []. Qed.
+Lemma measurable0 : measurable (set0 : set T).
+Proof. by case: T => ? []. Qed.
 
-Lemma measurableC (A : set X) : measurable A -> measurable (~` A).
-Proof. by case: X A => ? []. Qed.
+Lemma measurableC (A : set T) : measurable A -> measurable (~` A).
+Proof. by case: T A => ? []. Qed.
 
-Lemma measurable_bigU (U : (set X)^nat) :
+Lemma measurable_bigU (U : (set T)^nat) :
   (forall i, measurable (U i)) -> measurable (\bigcup_i (U i)).
-Proof. by case: X U => ? []. Qed.
+Proof. by case: T U => ? []. Qed.
 
 End measurable_interface.
 
 Section measurable_lemmas.
-Variables X : measurableType.
-Implicit Types A B : set X.
+Variables T : measurableType.
+Implicit Types A B : set T.
 
-Lemma measurableT : measurable (setT : set X).
+Lemma measurableT : measurable (setT : set T).
 Proof. by rewrite -setC0; apply measurableC; exact: measurable0. Qed.
 
 Lemma measurableU A B : measurable A -> measurable B -> measurable (A `|` B).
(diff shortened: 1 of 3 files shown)


theories/measure.v
```diff
@@ -0,0 +1,526 @@
...
+have -> : \bigcup_n A n = \bigcup_n (B_of A) n := A_B_of ndA.
+have mB : forall i, measurable (B_of A i).
+  by elim=> [|i ih] //=; apply: measurableD.
+apply: cvg_trans (is_measure_sigma_additive mB Binter).
+rewrite (_ : (fun n => \sum_(i < n.+1) mu (B_of A i))%E = mu \o A) //.
+rewrite funeqE => n; rewrite -is_measure_additive// bigcup_ord.
+by rewrite -bigcup_ord -AE.
+Qed.
+
+Lemma bool_le (A : (set X) ^nat) : (forall i, measurable (A i)) ->
```
CohenCyril, line 345: Maybe `Theorem Boole_inequality` instead. Otherwise by convention `bool` means `bool : Type`

theories/measure.v
```diff
@@ -300,7 +305,7 @@ rewrite big_ord_recr [in RHS]big_ord_recr /= predeqE => x; split=> [[Ax|An1x]|].
 move=> [summyB|[An1x NAnx]]; by [rewrite ih; left | right].
 Qed.
 
-Lemma homo_bigUB_of (A : (set X) ^nat) n :
+Lemma homo_bigUB_of (A : (set T) ^nat) n :
```
CohenCyril, line 308: Statements of `{homo : ... }` or `{mono : ...}` shouldn't be named after `homo`, or `mono`, just as for `{morph ...}` lemmas. Instead please use the head of the unfolded statement (for homo) or the head of the LHS of the equality (for mono). In this case: `sub_bigUB_of`.

CohenCyril: When a `mono` lemma subsumes `homo`, it gets priority for the short name, and if really needed the corresponding `homo` lemma can be suffixed with `W`. If the `mono` lemma is only valid on a subdomain, then the `homo` lemma takes the short name, and the `mono` lemma gets the suffix `in`.
(This should all be added to CONTRIBUTING.md I guess)

theories/measure.v
```diff
@@ -527,47 +533,48 @@ Qed.
...
 Proof.
 apply/eqP; rewrite eq_le; apply/andP; split; last first.
   by apply: mu_ext_ge0; exact: measurable0.
 rewrite /mu_ext; apply ereal_inf_lb.
 by exists set0 => //; split => //; exact: measurable0.
 Qed.
 
-Lemma mu_ext_monotone mu : {in [set x | measurable x] &,
-  {homo mu_ext mu : A B / A `<=` B >-> (A <= B)%E}}.
+Lemma mu_ext_monotone :
```
CohenCyril, line 555: `lee_mu_ext`

theories/measure.v
```diff
@@ -0,0 +1,606 @@
...
+End Exports.
+
+End Measure.
+Include Measure.Exports.
+
+Section is_measure_lemmas.
+Variables (R : numFieldType) (T : measurableType).
+Variable mu : {measure set T -> {ereal R}}.
+
+Lemma is_measure0 : mu set0 = 0%:E.
```
CohenCyril, line 223: No `is_` prefix anymore, since it is not about a predicate but a packaged operator.
Same is true everywhere incuding in `outer_measure`.

The exception is `is_measure f := (axioms f)` itself, which is a predicate.

theories/measure.v
```diff
@@ -0,0 +1,606 @@
...
+Proof.
+move=> ndA; rewrite funeqE => x; rewrite propeqE; split.
+  move=> -[n _]; rewrite (homo_bigUB_of _ ndA) bigcup_ord => -[m mn ?].
+  by exists m.
+move=> -[m _] myBAmx; exists m => //; rewrite (homo_bigUB_of _ ndA) bigcup_ord.
+by exists m.
+Qed.
+
+(* 401,p.43 measure is continuous from below *)
+Lemma measure_bigcup (A : (set T) ^nat) :
```
CohenCyril, line 332: I would name it `cvg_mu_inc`.

theories/measure.v
```diff
@@ -0,0 +1,606 @@
...
+  by elim=> [|i ih] //=; apply: measurableD.
+apply: cvg_trans (is_measure_sigma_additive mB Binter).
+apply: (@cvg_trans _ [filter of (fun n => (\sum_(i < n.+1) mu (B_of A i))%E)]); last first.
+  by move=> S [n _] nS; exists n => // m nm; apply/(nS m.+1)/(leq_trans nm).
+rewrite (_ : (fun n => \sum_(i < n.+1) mu (B_of A i))%E = mu \o A) //.
+rewrite funeqE => n; rewrite -is_measure_additive// bigcup_ord.
+by rewrite -bigcup_ord -AE.
+Qed.
+
+Theorem Boole_inequality (A : (set T) ^nat) : (forall i, measurable (A i)) ->
```
CohenCyril, line 352: This name is nice for historical purposes and external eye.
Internally I suspect it will become annoying and we should at least provide an alias: e.g. `le_mu_bigsetU`.
(and the generalized one `\le_mu_bigcup`)

theories/measure.v
```diff
@@ -0,0 +1,606 @@
...
+  move=> n m nm; apply: le_measure => //; try by rewrite inE; apply mB.
+  exact: bigsetU_incl.
+have BA : forall m, (mu (B m) <= lim (fun n : nat => \sum_(i < n) mu (A i)))%E.
+  move=> m; rewrite (le_trans (Boole_inequality mu mA m.+1)) // -/(B m).
+  by apply: (@series_nneg _ (mu \o A)) => n; exact: is_measure_ge0.
+by apply ub_ereal_sup => /= x [n _ <-{x}]; apply BA.
+Qed.
+
+End generalized_boole_inequality.
+
```
CohenCyril, line 455: @affeldt-aist I would vouch for cutting here for this PR.

theories/measure.v
```diff
@@ -0,0 +1,441 @@
...
+elim: n => [|n ih]; first by rewrite !big_ord_recl !big_ord0.
+rewrite big_ord_recr [in RHS]big_ord_recr /= predeqE => x; split=> [[Ax|An1x]|].
+    by rewrite -ih; left.
+  rewrite -ih.
+  have [Anx|Anx] := pselect (A n x); last by right.
+  by left; rewrite big_ord_recr /=; right.
+move=> [summyB|[An1x NAnx]]; by [rewrite ih; left | right].
+Qed.
+
+Lemma sub_bigUB_of (A : (set T) ^nat) n :
```
CohenCyril, line 313: My appologies I got carried away and did not realize the `homo` was not even the conclusion.
In fact the right name would be `eq_bigsetUB_of`.
```suggestion
Lemma eq_bigsetUB_of (A : (set T) ^nat) n :
```

theories/measure.v
```diff
@@ -0,0 +1,441 @@
...
+move=> ndA; elim: n => [|n ih]; rewrite funeqE => x; rewrite propeqE; split.
+- by move=> ?; rewrite big_ord_recl big_ord0; left.
+- by rewrite big_ord_recl big_ord0 setU0.
+- rewrite (UB_of ndA) => -[|/=].
+  by rewrite big_ord_recr /= -ih => Anx; left.
+  by move=> -[An1x Anx]; rewrite big_ord_recr /=; right.
+- rewrite big_ord_recr /= -ih => -[|[]//]; exact: ndA.
+Qed.
+
+Lemma A_B_of (A : (set T) ^nat) : {homo A : n m / (n <= m)%nat >-> n `<=` m} ->
```
CohenCyril, line 326: And here
```suggestion
Lemma eq_bugcupB_of (A : (set T) ^nat) : {homo A : n m / (n <= m)%nat >-> n `<=` m} ->
```

PR #180, opened 2020-04-14
Mathcomp master PseudoMetricNormedZmod

mkerjean:
Further generalisations from normedModTypes to PseudoMetricNormedZmodTypes.

NB(rei): fixes #382 

TODO:
- [x] Several explicit arguments are needed with this PR. E.g : https://github.com/math-comp/analysis/blob/45e488f1059a6744fb538a104300033555eb0046/theories/derive.v#L789 Why ?
- [x] Update Changelog
- [x] Check that no new (post apr. 2019) lemmas are generalizable

PR #230, opened 2020-06-17
Fix inheritance from pseudoMetricNormedZmodType to completeNormedModType

pi8027:


pi8027:
@mkerjean @affeldt-aist Could you rework #180 using this fix?

affeldt-aist:
> @mkerjean @affeldt-aist Could you rework #180 using this fix?

Will do. Thank you!

PR #237, opened 2020-07-20
lemmas about supremum and supremums

affeldt-aist:
Lemmas used in the branch `integral_sketch`.

### DIFF
--- a/theories/classical_sets.v
+++ b/theories/classical_sets.v
@@ -989,18 +989,28 @@ Lemma lbP E x : (forall y, E y -> (x <= y)%O) <-> lbound E x.
 Proof. by []. Qed.
 
 Lemma ub_set1 x y : ubound [set x] y = (x <= y)%O.
-Proof.
-by rewrite propeqE; split => [/ubP/(_ x erefl)//|xy]; apply/ubP => z ->.
-Qed.
+Proof. by rewrite propeqE; split => [/(_ x erefl)//|xy z ->]. Qed.
+
+Lemma lb_set1 x y : lbound [set x] y = (x >= y)%O.
+Proof. by rewrite propeqE; split => [/(_ x erefl)//|xy z ->]. Qed.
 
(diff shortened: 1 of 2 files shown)


theories/classical_sets.v
```diff
@@ -980,3 +980,29 @@ Definition supremum (x0 : T) E :=
...
 
 End UpperLowerTheory.
+
+Section UpperLowerOrderTheory.
+Import Order.TTheory.
+Variables (d : unit) (T : orderType d).
+Implicit Types E : set T.
+
+Lemma lt_supremum_Nmem x0 E t :
+  supremums E !=set0 -> (supremum x0 E < t)%O -> ~ E t.
```
CohenCyril, line 990: I'd rather swap the last two arguments: `E t -> (supremum x0 E >= t)%O` and use a [`contra_notN`](https://github.com/math-comp/math-comp/blob/3ceb153b972cbfc23a33daa740ec31050881bfa2/mathcomp/ssreflect/ssrbool.v#L227) lemma to finish the job.
Maybe we should also envision adding `contra_not_le` etc lemmas

theories/classical_sets.v
```diff
@@ -980,3 +980,29 @@ Definition supremum (x0 : T) E :=
...
+Proof.
+move=> [x Ex]; rewrite /supremum; case: pselect => /= [_|]; last first.
+  by move/set0P/negP; rewrite negbK => /eqP -> _.
+case: xgetP => [t' t'E [uEt' _ t't Et]|].
+- by apply/negP : t't; rewrite -leNgt uEt'.
+- by move/(_ x).
+Qed.
+
+End UpperLowerOrderTheory.
```
CohenCyril, line 999: shouldn't there be the same for "infimum"?

theories/classical_sets.v
affeldt-aist: I recorded the idea as a draft PR to mathcomp (https://github.com/math-comp/math-comp/pull/576).

CohenCyril (approved):
ok to merge after rebase

PR #241, opened 2020-07-26
compatibility with coq 8.12

affeldt-aist:
This is to fix a compatibility problem with Coq 8.12 observed by Karl.

### DIFF
--- a/opam
+++ b/opam
@@ -19,7 +19,7 @@ install: [
   [make "install"]
 ]
 depends: [
-  "coq" { ((>= "8.7" & < "8.12~") | = "dev") }
+  "coq" { ((>= "8.7" & < "8.13~") | = "dev") }
   "coq-mathcomp-field"       {(>= "1.11.0" & < "1.12~")}
   "coq-mathcomp-finmap"      {(>= "1.5.0" & < "1.6~")}
 ]
--- a/opam
+++ b/opam
@@ -19,7 +19,7 @@ install: [
   [make "install"]
 ]
 depends: [
-  "coq" { ((>= "8.7" & < "8.13~") | = "dev") }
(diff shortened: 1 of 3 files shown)


theories/landau.v
```diff
@@ -978,7 +978,7 @@ move->; apply/eqOP; have [k c1 kOg] := bigO _ g. have [k' c2 k'Ok] := bigO _ k.
 near=> c; move: k'Ok kOg; apply: filter_app2; near=> x => lek'c2k.
 rewrite -(@ler_pmul2l _ c2%:num) // mulrA => /(le_trans lek'c2k) /le_trans; apply.
 by rewrite ler_pmul //; near: c; apply: locally_pinfty_ge.
-Unshelve. end_near. Grab Existential Variables. all: end_near. Qed.
```
CohenCyril, line 981: @affeldt-aist I wonder if this still compiles with Coq 8.10 and 8.11

opam
```diff
@@ -19,7 +19,7 @@ install: [
   [make "install"]
 ]
 depends: [
-  "coq" { ((>= "8.7" & < "8.12~") | = "dev") }
+  "coq" { ((>= "8.7" & < "8.13~") | = "dev") }
```
CohenCyril, line 22: AFAIK Coq < 8.10 is not supported anymore

theories/landau.v
affeldt-aist: It went through. But we'd better get rid of it.

CohenCyril: Sorry, I meant: does removing this still make Coq 8.10 pass?

affeldt-aist: I think so: https://travis-ci.org/github/math-comp/analysis/jobs/711965366.

PR #256, opened 2020-09-09
rename/remove contra lemmas

affeldt-aist:
Rename `contra` lemmas following the notation scheme adopted in MathComp and remove duplicate definitions of contra lemmas.

TODO when merging: add an issue to remove `contra_not` when it is available with a future release of MathComp

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -22,8 +22,16 @@
   + `setUDl` -> `setUIl`
   + `setUDr` -> `setUIr`
   + `setIDr` -> `setIUl`
+- in `boolp.v`:
+  + `contrap` -> `contra_not`
+  + `contrapL` -> `contraPnot`
+  + `contrapR` -> `contra_notP`
+  + `contrapLR` -> `contraPP`
 
 ### Removed
+
+- in `boolp.v`:
+  + `contrapNN`, `contrapTN`, `contrapNT`, `contrapTT`
 
 ### Infrastructure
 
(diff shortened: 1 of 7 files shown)


theories/boolp.v
```diff
@@ -269,38 +269,30 @@ by case=> x bPx; apply/asboolP; exists x; apply/asboolP.
 Qed.
 
 (* -------------------------------------------------------------------- *)
-Lemma contrap (Q P : Prop) : (Q -> P) -> ~ P -> ~ Q.
+Lemma contra_not_not (Q P : Prop) : (Q -> P) -> ~ P -> ~ Q.
```
CohenCyril, line 272: It is `contra_not` (https://github.com/math-comp/math-comp/blob/master/mathcomp/ssreflect/ssrbool.v#L215)

affeldt-aist: Then I guess we'll need to make an issue to later remove `contra_not` (when MathComp's new master is released).

CHANGELOG_UNRELEASED.md
```diff
@@ -22,9 +22,20 @@
...
+  + `contrapLR` -> `contraPP`
 
 ### Removed
 
+- in `boolp.v`:
+  + `contrapNN` (use `contra_not_not`)
+  + `contrapTN` (use `contraPnot`)
+  + `contrapNT` (use `contra_notP`)
+  + `contrapTT` (use `contraPP`)
+
```
CohenCyril, line 38: This looks like unneeded changelog duplication, doesn't it?

theories/boolp.v
CohenCyril: yes indeed

PR #272, opened 2020-11-06
add link to Marie's slides at TYPES 2019

affeldt-aist:
@mkerjean It is ok with you?


mkerjean:
@affeldt-aist Yes if you think they can be useful (examples are outdated already)

PR #275, opened 2020-11-10
Compatible with MathComp 1.12

pi8027:
This PR makes MathComp analysis compatible with the master branch of MathComp and math-comp/real-closed#28.

CC: @affeldt-aist @mkerjean

### DIFF
--- a/theories/normedtype.v
+++ b/theories/normedtype.v
@@ -81,19 +81,6 @@ Unset Printing Implicit Defensive.
 Import Order.TTheory GRing.Theory Num.Def Num.Theory.
 
 Local Open Scope ring_scope.
-
-Section add_to_mathcomp.
-
-Lemma ltr_distW (R : realDomainType) (x y e : R) :
-  `|x - y| < e -> y - e < x.
-Proof. by rewrite ltr_distl => /andP[]. Qed.
-
-Lemma ler_distW (R : realDomainType) (x y e : R):
-  `|x - y| <= e -> y - e <= x.
-Proof. by rewrite ler_distl => /andP[]. Qed.
-
-End add_to_mathcomp.
-
 Local Open Scope classical_set_scope.
 
 Lemma nbhsN (R : numFieldType) (x : R^o) :
@@ -2065,9 +2052,8 @@ have /cauchy_ballP /cauchyP /(_ 1) [//|x0 x01] := F_cauchy.
 have D_has_sup : has_sup D; first split.
 - exists (x0 - 1) => A FA.
   near F => x.
-  apply/downP; exists x.
-  by near: x.
-  by rewrite ler_distW 1?distrC 1?ltW ?andbT //; near: x.
+  apply/downP; exists x; first by near: x.
+  by rewrite ler_distl_subl // ltW //; near: x.
 - exists (x0 + 1); apply/ubP => x /(_ _ x01) /downP [y].
   rewrite -[ball _ _ _]/(_ (_ < _)) ltr_distl ltr_subl_addr => /andP[/ltW].
   by move=> /(le_trans _) yx01 _ /yx01.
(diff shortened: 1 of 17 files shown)


mkerjean:
> This PR makes MathComp analysis compatible with the master branch of MathComp and [math-comp/real-closed#28](https://github.com/math-comp/real-closed/pull/28).
> 
> CC: @affeldt-aist @mkerjean

Thanks ! So `inE` should be replaced by `in_ivt` but not textually ? `inE` can't be adapted to include `in_ivt` ?

pi8027:
@mkerjean Performing `rewrite inE` for `x \in i` where `i` is an interval now gives us `[x, x] <= i` where `<=` is the subset relation for intervals. So we have to use `in_itv` instead. https://github.com/math-comp/math-comp/blob/72c13992b8961f288c412414fda206213486e25b/mathcomp/algebra/interval.v#L243

mkerjean:
> @mkerjean Performing `rewrite inE` for `x \in i` where `i` is an interval now gives us `[x, x] <= i` where `<=` is the subset relation for intervals. So we have to use `in_itv` instead. https://github.com/math-comp/math-comp/blob/72c13992b8961f288c412414fda206213486e25b/mathcomp/algebra/interval.v#L243

and isn't the fact that `inE` can't apply to intervals an issue for other developments over MathComp ? Do you think this can be corrected at the level of `interval.v` ?

pi8027:
@mkerjean For concrete intervals whose bounds are known to be one of unbounded, open, or close, the former behavior of `inE` is preferable indeed. However, the new interval library (math-comp/math-comp#458) provides a way to reason about intervals abstractly. Intervals are now canonically a `latticeType` (if the underlying type `T` is a `latticeType`) and it allows us to reduce problems about intervals to problems about a lattice. In this case, I think that the new behavior of `inE` is preferable. So I would say it is not an issue we should fix on the MathComp side...

pi8027:
I think @CohenCyril may have an opinion on this issue.

CohenCyril:
Indeed, there seems to be a tension between generic intervals `i` and concrete intervals (e.g. `[a, b]`), if we want `inE` to be readable even for abstract intervals, preferring the form `` `[x, x] <= i` is unavoidable... however 
1. all are be convertible (e.g. all of ``x \in `[a, b]``, `` `[x, x] <= i`` and `a <= x <= b` are convertible`
2. if you have an assumption of the form ``xab : x \in `[a, b]``, you should avoid using either `inE` or `in_itv` and use `(itvP xab)` instead.

theories/normedtype.v
```diff
@@ -2253,37 +2252,32 @@ rewrite [X in closed X](_ : (eq^~ _) = ~` (xpredC (eq_op^~ y))).
...
-  if b is BOpen _ then true else false.
+Definition bound_side d (T : porderType d) (c : bool) (x : itv_bound T) :=
+  if x is BSide c' _ then c == c' else false.
 
-Definition isBClosed (b : itv_bound R) :=
-  if b is BOpen_if false _ then true else false.
-
-Lemma interval_open a b : ~~ isBClosed a -> ~~ isBClosed b ->
+Lemma interval_open a b : ~~ bound_side true a -> ~~ bound_side false b ->
   open [set x : R^o | x \in Interval a b].
```
pi8027, line 2259: @CohenCyril Thanks for fixing! Splitting `Interval a b` into two intervals using `itv_splitI` would be an easier way to prove this and `interval_closed`. (But we can do it later.)

CohenCyril: Sure, I discovered a problem with the uses of `eq_set` here that I am a bit worried about... but, later as well...

CohenCyril:
I am switching this PR back to  a draft to prevent untimely merges.
The reason is that I believe we should release one last version of math-comp/analysis for coq + mathcomp 1.11 with #298 before we merge this one for good.

affeldt-aist (approved):
I feel that the difference between `inE` in `in_itv` for intervals is worth documenting.

pi8027:
`INSTALL.md` has to be changed.

pi8027:
I'm fine with the changes done by Cyril and Reynald. Shall we merge? (I hope to take a look at #205 tomorrow.)

affeldt-aist:
Let's merge since it is waited for for PR #205.

PR #281, opened 2020-11-13
Making sure `[set x | P]` notations are not expanded eagerly

CohenCyril:
- unless applied, where `/=` will simplify.
- unless one uses `rewrite /mkset`.


affeldt-aist:
>     * unless applied, where `/=` will simplify.
> 
>     * unless one uses `rewrite /mkset`.

To confirm. Looking at the diff it seems that we now have from time to time to
do explicit unfolds and simplifys. Does it mean that this solution is a compromise?

CohenCyril:
> Does it mean that this solution is a compromise?

To me these unfolding and simplify are necessary steps to have enough control, and not a compromise.

PR #297, opened 2020-12-07
generate files using Coq-community Templates

affeldt-aist:


meta.yml
```diff
@@ -0,0 +1,120 @@
...
+- name: Reynald Affeldt
+  initial: true
+- name: Cyril Cohen
+  initial: true
+- name: Assia Mahboubi
+  initial: true
+- name: Damien Rouhling
+  initial: true
+- name: Pierre-Yves Strub
+  initial: true
```
CohenCyril, line 24: Add Marie Kerjean and Kazuhiko Sakaguchi if they agree.

affeldt-aist: @mkerjean @pi8027

meta.yml
```diff
@@ -0,0 +1,120 @@
...
+  ## Disclaimer
+ 
+  This library is still at an experimental stage.  Contents may
+  change, definitions and theorems may be renamed, and inference
+  mechanisms may be replaced at any major version bump.  Use at your
+  own risk.
+
+  ## Documentation
+
+  [![project chat](https://img.shields.io/badge/zulip-join_chat-brightgreen.svg)](https://coq.zulipchat.com/#narrow/stream/237666-math-comp-analysis)
```
CohenCyril, line 88: This seems to suggest that an entry should be added to coq template for associated chats.
CC: @Zimmi48

meta.yml
pi8027: Yes, I agree.

mkerjean: I agree

meta.yml
```diff
@@ -0,0 +1,120 @@
...
+  text: Coq 8.11 to 8.12
+  opam: '{ (>= "8.11" & < "8.13~") | (= "dev") }'
+
+tested_coq_opam_versions:
+- version: '1.11.0-coq-8.11'
+  repo: 'mathcomp/mathcomp'
+- version: '1.11.0-coq-8.12'
+  repo: 'mathcomp/mathcomp'
+
+dependencies:
```
CohenCyril, line 45: We should document the whole list of dependency (`ssreflect`, `mathcomp-algebra`, ... whatever is pulled in by `mathcomp-field`)

meta.yml
Zimmi48: Yes, we have such a badge for coq-community projects but not for outside projects. PRs (or issues) welcome!

affeldt-aist: What about https://github.com/coq-community/templates/pull/82 ?

meta.yml
```diff
@@ -0,0 +1,154 @@
...
+
+publications:
+- pub_url: https://jfr.unibo.it/article/view/8124
+  pub_title:  Formalization Techniques for Asymptotic Reasoning in Classical Analysis
+  pub_doi: 10.6092/issn.1972-5787/8124 
+- pub_url: https://hal.inria.fr/hal-02463336
+  pub_title: Competing inheritance paths in dependent type theory---a case study in functional analysis
+  pub_doi: 10.1007/978-3-030-51054-1_1
+- pub_url: http://www-sop.inria.fr/members/Damien.Rouhling/data/phd/thesis.pdf
+  pub_title: Outils pour la Formalisation en Analyse Classique
```
CohenCyril, line 103: @affeldt-aist 
```suggestion
  pub_title: Formalisation Tools for Classical Analysis
```
@drouhling's thesis is in english, let's not deter our english readers by putting the french "by law" title, and let's not give false hopes to french speakers, because they won't get far in french only :laughing:

CohenCyril:
Great :+1:

PR #283, opened 2020-11-16
Closed ball

mkerjean:
A theory of closed balls.

PR #305, opened 2020-12-11
ax5 added to PseudoMetric.mixin_of

affeldt-aist:
- does not compile
- and there is an admit

### DIFF
--- a/theories/topology.v
+++ b/theories/topology.v
@@ -3560,17 +3560,18 @@ Proof. by []. Qed.
 
 Module PseudoMetric.
 
-Record mixin_of (R : numDomainType) (M : Type) (entourage : set (set (M * M))) := Mixin {
+Record mixin_of (R : numDomainType) (M : topologicalType) (entourage : set (set (M * M))) := Mixin {
   ball : M -> R -> M -> Prop ;
   ax1 : forall x (e : R), 0 < e -> ball x e x ;
   ax2 : forall x y (e : R), ball x e y -> ball y e x ;
   ax3 : forall x y z e1 e2, ball x e1 y -> ball y e2 z -> ball x (e1 + e2) z;
-  ax4 : entourage = entourage_ ball
+  ax4 : entourage = entourage_ ball ;
+  ax5 : forall x (e : R), open (ball x e)
 }.
 
 Record class_of (R : numDomainType) (M : Type) := Class {
   base : Uniform.class_of M;
-  mixin : mixin_of R (Uniform.entourage base)
+  mixin : @mixin_of R (Topological.Pack (Uniform.base base)) (Uniform.entourage base)
 }.
 
 Section ClassDef.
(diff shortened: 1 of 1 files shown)


theories/topology.v
```diff
@@ -3992,8 +4011,13 @@ rewrite predeqE => A; split; last first.
 move=> [P]; rewrite -entourage_ballE => -[_/posnumP[e] sbeP] sPA.
 by exists e%:num => // fg fg_e; apply: sPA => t; apply: sbeP; apply: fg_e.
 Qed.
+
+Lemma fct_open_ball x e : open (fct_ball x e).
+Proof.
+Admitted.
```
mkerjean, line 4017: I am not sure to unnderstand why `fct_open_ball`  should be true.  Isn't there some continuity hypothesis at stake ?

affeldt-aist:
by adding a topological structure to zmodType, I am afraid I start doing things that look like PR #205

theories/topology.v
CohenCyril: Actually, I don't even think continuity is enough. There is no reason for this to be true.
It seems we have to backtrack on our decision to add `ax5` and introduce `open_ball x e = (ball x e)°` instead... WDYT?

CohenCyril:
> by adding a topological structure to zmodType, I am afraid I start doing things that look like PR #205

One should not do that, you are adding a trivial topology where everything is open, indeed you defined singletons to be open, hence any set, as an arbitrary union of singletons is automatically open...

Moreover, if you were to add the discrete topology instead, it would be incompatible with any other topology that we use... If one really need discrete topologies, they need to introduce an alias `discrete T := T` and have canonicals on `discrete` instead...

affeldt-aist:
well, this was really to keep going and uncover more problems, it we are to have a topologicalType as a parameter of the pseudometric mixin then the end of the file needs to be changed for we cannot used `topologyOfEntourageMixin` any more, am I wrong?

CohenCyril:
> well, this was really to keep going and uncover more problems, if we are to have a topologicalType as a parameter of the pseudometric mixin then the end of the file needs to be changed for we cannot used `topologyOfEntourageMixin` anymore, am I wrong?

I'm not sure I understand...

affeldt-aist:
> 
> 
> > well, this was really to keep going and uncover more problems, if we are to have a topologicalType as a parameter of the pseudometric mixin then the end of the file needs to be changed for we cannot used `topologyOfEntourageMixin` anymore, am I wrong?
> 
> I'm not sure I understand...

I think that normedZmodType needs to be given its topological structure ealier

CohenCyril:
> I think that ~normedZ~ zmodType needs to be given its topological structure earlier

I do not understand why you say that, moreover `zmodType` should not have a canonical structure...

affeldt-aist:
> > I think that ~normedZ~ zmodType needs to be given its topological structure earlier
> 
> I do not understand why you say that, moreover `zmodType` should not have a canonical structure...

It was to type check the `open (ball_ Num.Def.normr x e)` statement at line 4364 (though now we know that we can't do with this axiom anyway).

CohenCyril:
> It was to type check the `open (ball_ Num.Def.normr x e)` statement at line 4364 (though now we know that we can't do with this axiom anyway).

Then all you need is the topology of a  normedZmodType structure, and it must and should not be canonical, but local, since we are defining a factory on this line.

affeldt-aist:
We gave up extending the interface of pseudometric spaces. This PR is scheduled to be closed as it is. A related discussion is going on in PR #283 .

PR #308, opened 2020-12-13
remove typeclasses opaque fmap

affeldt-aist:


CohenCyril:
@affeldt-aist I merged too quickly, there is a missing changelog entry.

affeldt-aist:
You mean that fact that we remove `Typeclasses Opaque fmap.` ?

CohenCyril:
> You mean that fact that we remove `Typeclasses Opaque fmap.` ?

Yes, it has side effects: more TC are solved and this should be logged.

> Does the change of definition of `interior` qualify for the changelog?

No, it is almost syntactically equal (even though the input is slightly different) and should be convertible.

PR #314, opened 2021-01-07
more uniform variable naming convention

affeldt-aist:
- enforce the convention of using T, U for names of variables
  in Type (or choiceType an pointedType)
  + as a consequence functions are of type T -> U
  + use I when the type corresponds to an index
- enforce the convention that sets are named A, B, C, D, etc.
  + indexed sets are rather named F
  + use X for set T and Y for set U when it makes sense when
    we are talking about images/preimages
  + keep X for sets of pairs
- small reordering of lemmas in sections (basic_lemmas,
  image_lemmas, bigop_lemmas) to take advantage of Implicit Types

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -25,6 +25,7 @@
 
 - in `ereal.v`, lemmas `esum_fset_ninfty`, `esum_fset_pinfty`, `esum_pinfty`
 - in `classical_sets.v`, lemmas `setDT`, `set0D`, `setD0`
+- in `classical_sets.v`, lemmas `setC_bigcup`, `setC_bigcap`
 
 ### Changed
 
@@ -56,6 +57,7 @@
 
 - in `topology.v`:
   + `ball_le`
+- in `classical_sets.v`, lemma `bigcapCU`
 
 ### Infrastructure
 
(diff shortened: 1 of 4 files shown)


CohenCyril:
I'd rather reuse the convetions from `mathcomp/ssreflect/finset.v`

> * enforce the convention of using T, U for names of variables
>   in Type (or choiceType an pointedType)

Use `T` (if multiple types, use a suffix as in `T'` or `T1` ... `Tn` or a prefix as in `aT` for domain type or  `rT` return type)

>   * use I when the type corresponds to an index

Use `I` or `J` for indexes.

> * enforce the convention that sets are named A, B, C, D, etc.

Yes

>   * indexed sets are rather named F
>   * use X for set T and Y for set U when it makes sense when

Use the same sufix or prefix for the sets as their containing type (e.g. `A1` in `T1`, etc)

>     we are talking about images/preimages
>   * keep X for sets of pairs

or remove the suffix? (e.g. `A : set (T1 * T2)`)

> * small reordering of lemmas in sections (basic_lemmas,
>   image_lemmas, bigop_lemmas) to take advantage of Implicit Types

CohenCyril:
+ document these choices in the header?

CohenCyril (approved):
LGTM

CohenCyril:
Please rebase and then we merge

PR #321, opened 2021-01-14
lemmas to prove inf/sup properties of intervals

affeldt-aist:
used in the branch integral_sketch to prove properties of intervals of real numbers


CohenCyril (approved):

CohenCyril:
LGTM merge when rebased

PR #322, opened 2021-01-14
definition of ceil and some lemmas

affeldt-aist:
Needed that for a proof about enclosing intervals with integer bounds.
This is transitory code waiting for (math-comp/math-comp#682 and math-comp/math-comp#691) to be integrated.

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -18,16 +18,36 @@
 
 - in `reals.v`, lemmas `sup_setU`, `inf_setU`
 - in `boolp.v`, lemmas `iff_notr`, `iff_not2`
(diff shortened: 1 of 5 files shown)


CohenCyril:
This is transitory code waiting for (https://github.com/math-comp/math-comp/pull/682 and https://github.com/math-comp/math-comp/pull/691) to be integrated.
Can you please check that the definitions have the same type here as there, that the lemmas have the same shape and that you are not using the definitions anywhere else? If a lemma is missing from the above PR, you could suggest their insertion.

CohenCyril:
Also, note that `Rtoint` is to disappear in favor of the more general and informative `Num.floor`.

CohenCyril (approved):
OK until mathcomp 1.13 comes out

Issue #325, opened 2021-01-19
`bounded_on` -> `bounded_near`


mkerjean:
In `normedtype.v`, `bounded_on f F` means f is bounded on at least one element on the filter `F`. Renaming it by `bounded_near` would be closer to its semantics, and allow the use of `bounded_on f B` for functions bounded on a bornology B (e.g.: simply bounded, uniformly bounded, ...) . 

If legitimate, should this renaming be performed in PR [183](https://github.com/math-comp/analysis/pull/183) or in a new PR ?

CohenCyril:
Makes sense! Please proceed!

mkerjean:
> Makes sense! Please proceed!

In a new PR or in PR 183 ?

CohenCyril:
Whatever is more practical for you!

affeldt-aist:
Fixed by PR #327

PR #327, opened 2021-01-19
`bounded_on` -> `bounded_near`and doc on bounded_fun

mkerjean:
Solves issue [#325](https://github.com/math-comp/analysis/issues/325). It opens the way for a formalisation of bornologies and bounded functions.


CohenCyril (approved):
LGTM

PR #328, opened 2021-01-26
infinite sum contains one infinite summand

affeldt-aist:
pinfty version of esum_ninfty, whose proof has been split with an intermediate lemma


CohenCyril (approved):
The proofs could be shortened I think, but it's good to have the result in the library!

PR #329, opened 2021-01-26
sigma_finite definition from PR integral_sketch

affeldt-aist:
- lowered to ringOfSetsTypes

PRed because it might be needed earlier than integral.v where it was first defined

Co-authored-by: Cyril Cohen <cohen@crans.org>

Also fixes constructor names for two mixins


CohenCyril (changes requested):
Please adhere to the new conventions of classical sets. (X, Y -> A, B I guess?)

PR #340, opened 2021-03-12
silence warnings

affeldt-aist:
What about merging this for the time being?

We'll set back the flags after issues #344 , #342 , #341 are addressed.


affeldt-aist:
This is partly inspired by recent PRs to MathComp :-) @CohenCyril

PR #349, opened 2021-03-18
more lemmas about `ereal_ball`

affeldt-aist:
- isolate a few generic lemmas about `ereal_ball` from the
  construction of entourages for extended reals

This is an effort to clean the instantiation of pseudometric spaces with extended reals.
Even if the progress is minor, I would be happy to have it merged asap: I will get back
to it later but I am not sure when.


CohenCyril (approved):
LGTM

PR #350, opened 2021-03-23
more lemmas about extended real numbers

affeldt-aist:
- used in `integral_sketch`

### DIFF
--- a/theories/ereal.v
+++ b/theories/ereal.v
@@ -359,6 +359,21 @@ Proof. by []. Qed.
 
 End ERealOrderTheory.
 
+Section finNumPred.
+Context {R : numDomainType}.
+
+Definition fin_num := [qualify a x : er R | (x != -oo) && (x != +oo)].
+Fact fin_num_key : pred_key fin_num. by []. Qed.
+Canonical fin_num_keyd := KeyedQualifier fin_num_key.
+
+Lemma fin_numE x : (x \is a fin_num) = ((x != -oo) && (x != +oo)).
+Proof. by []. Qed.
+
+Lemma fin_numP x : reflect ((x != -oo) /\ (x != +oo)) (x \in fin_num).
+Proof. by apply/(iffP idP) => [/andP//|/andP]. Qed.
+
+End finNumPred.
+
 Section ERealArithTh_numDomainType.
 
 Context {R : numDomainType}.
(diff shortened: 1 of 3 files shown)


theories/ereal.v
```diff
@@ -404,9 +446,25 @@ Qed.
 Lemma is_realD x y : is_real (x + y) = (is_real x) && (is_real y).
 Proof. by move: x y => [x| |] [y| |]. Qed.
 
+Lemma real_of_erD x y : is_real x -> is_real y ->
+  (real_of_er (x + y) = real_of_er x + real_of_er y)%R.
```
CohenCyril, line 450:
```suggestion
Lemma real_of_erD : {in is_real &, {morph real_of_er : x y / x + y}}
```
Does it work?

affeldt-aist: This one goes through:
```
{in is_real &, {morph real_of_er : x y / x + y >-> (x + y)%R}}
```
Its usage suggests to introduce the `... \is real` notation as per the next commit.

CohenCyril: `real` is too overloaded... `finite` would actually be more appropriate but it is also too overloaded...

CohenCyril: `finite_num` sounds good.

affeldt-aist: are a notation `x \isn't oo`?

CohenCyril: mmmh could work...

CohenCyril: but then it would result in
```coq
{in predC oo &, {morph real_of_er : x y / x + y >-> (x + y)%R}}
```
which is not as nice as the fully applied use...

CohenCyril: moreover we might want to have automatic inference on being finite, not being infinite...

CohenCyril: So... no

CohenCyril: In this case it should be `\is a fin_num`

theories/ereal.v
```diff
@@ -334,6 +334,46 @@ Notation "\sum_ ( i 'in' A | P ) F" :=
...
+End ERealOrderTheory.
+
+Section finNumPred.
+Context {R : numDomainType}.
+
+Definition fin_num := [qualify a x : er R | (x != -oo) && (x != +oo)].
+Fact fin_num_key : pred_key fin_num. by []. Qed.
+Canonical fin_num_keyd := KeyedQualifier fin_num_key.
+
+Lemma fin_num_def x : (x \is a fin_num) = ((x != -oo) && (x != +oo)).
```
CohenCyril, line 369:
```suggestion
Lemma fin_numE x : (x \is a fin_num) = ((x != -oo) && (x != +oo)).
```

PR #356, opened 2021-03-26
order of arguments of subset_trans

affeldt-aist:
fix the order of arguments of subset_trans


CohenCyril:
Maybe we could still have a changelog entry?

PR #359, opened 2021-03-26
hull of set0

affeldt-aist:
trivial, but it requires to touch several files

### DIFF
--- a/theories/sequences.v
+++ b/theories/sequences.v
@@ -870,7 +870,7 @@ have [Snoo|Snoo] := pselect (u_ = fun=> -oo%E).
   rewrite /l (_ : S = [set -oo%E]); last first.
     rewrite predeqE => x; split => [-[n _ <-]|->]; first by rewrite Snoo.
     by exists O => //; rewrite Snoo.
-  by rewrite ereal_sup_set1 Snoo; exact: cvg_cst.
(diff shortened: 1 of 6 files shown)


theories/sequences.v
```diff
@@ -867,12 +867,10 @@ have [Spoo|Spoo] := pselect (S +oo%E).
...
-  suff : l = -oo%E by move=> ->; rewrite Snoo; exact: cvg_cst.
-  rewrite /l.
-  suff -> : S = [set -oo%E] by rewrite ereal_sup_set1.
-  rewrite predeqE => x; split => [-[n _ <-]|->].
-  by rewrite Snoo.
-  by exists O => //; rewrite Snoo.
+  rewrite /l (_ : S = [set -oo%E]); last first.
+    rewrite predeqE => x; split => [-[n _ <-]|->]; first by rewrite Snoo.
+    by exists O => //; rewrite Snoo.
+  by rewrite ereal_sup_set1 Snoo; exact: cvg_cst.
```
CohenCyril, line 873: There seems to be a disparity between the conventions:
here you use lemma `ereal_sup_set1`, should it be `ereal_sup1`?

affeldt-aist: That's better. I did this renaming and closely related ones. There are more but that's maybe better to make an issue or another PR about them.

CohenCyril (approved):

PR #366, opened 2021-04-02
change the use of scopes in ereal notations

affeldt-aist:
- to reduce nested `%E`/`%R` in `integral_sketch` in particular


CohenCyril:
Great! Can you also take this opportunity to add `0%E` and `1%E` as definitions for `0%R%:E` and `1%R%:E`?

CohenCyril (approved):

PR #367, opened 2021-04-26
change notation `{ereal R}` to  `\bar R`

affeldt-aist:
as discussed in the 2021-03-26 meeting

### DIFF
--- a/theories/ereal.v
+++ b/theories/ereal.v
@@ -11,13 +11,13 @@ Require Import boolp classical_sets reals posnum topology.
 (******************************************************************************)
 (*                        Extended real numbers                               *)
 (*                                                                            *)
-(* Given a type R for numbers, {ereal R} is the type R extended with symbols  *)
-(* -oo and +oo (notation scope: %E), suitable to represent extended real      *)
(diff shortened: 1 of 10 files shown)


theories/ereal.v
```diff
@@ -108,7 +108,7 @@ Notation "+oo" := (@ERPInf _) : ereal_scope.
 Notation "-oo" := (@ERNInf _) : ereal_scope.
 Notation "x %:E" := (@ERFin _ x) (at level 2, format "x %:E").
 
-Notation "{ 'ereal' R }" := (er R) (format "{ 'ereal'  R }").
+Notation "'bar' x" := (er x) (at level 2, format "'bar'  x") : ereal_scope.
```
CohenCyril, line 111: Never ever do a string notation without guarding it with symbols or you will turn the string into a keyword...

CohenCyril: moreover if there was a scope to put, it should be `type_scope`.

CohenCyril:
I prefer `\bar R` much more

CohenCyril:
BTW, `er` does not make sense, it is so short we should use it instead of `\bar` but it is utterly uninformative.
I think it should be renamed `extended` or `with_infinities`.

PR #376, opened 2021-05-06
missing lemmas about floor and ceil

affeldt-aist:


theories/reals.v
```diff
@@ -623,6 +623,11 @@ Proof.
 by move=> ?; rewrite /floor -(@ltrz0 R) RtointK ?isint_Rfloor // Rfloor_lt0.
 Qed.
 
+Lemma floor_le : {homo @floor R : x y / x <= y}.
```
CohenCyril, line 626: `le_floor`

theories/reals.v
```diff
@@ -662,6 +667,11 @@ Proof. by move=> x0; rewrite -ler_oppl oppr0 floor_ge0 -ler_oppr oppr0. Qed.
 Lemma ler_ceil x : x <= (ceil x)%:~R.
 Proof. by rewrite -RceilE; exact: le_Rceil. Qed.
 
+Lemma ceil_le : {homo @ceil R : x y / x <= y}.
```
CohenCyril, line 670: `le_ceil`

theories/reals.v
```diff
@@ -662,6 +667,11 @@ Proof. by move=> x0; rewrite -ler_oppl oppr0 floor_ge0 -ler_oppr oppr0. Qed.
 Lemma ler_ceil x : x <= (ceil x)%:~R.
```
CohenCyril, line 667: Should be: `ceil_ge`

CohenCyril: And I guess it should come before `ceil_ge0` so that it could use it.

PR #385, opened 2021-06-02
A generic lemma to prove the continuity of inverse functions

ybertot:
and its instantiation to prove that sqrt is continuous

However, there is a doubt about the usefulness of such a lemma.  Most functions for which we need an inverse are actually derivable (x ^ 2, tan, sin, cos, exp), and there is a formula for the derivative of the inverse of a derivable function that might be more useful.

### DIFF
--- a/theories/inverse_continuous.v
+++ b/theories/inverse_continuous.v
@@ -1,0 +1,146 @@
+From mathcomp Require Import all_ssreflect all_algebra.
+Require Import classical_sets posnum topology
+  normedtype landau sequences boolp reals ereal.
+
+Import GRing.Theory Num.Theory Num.ExtraDef.
+Import Order.POrderTheory Order.TotalTheory.
+Import numFieldTopology.Exports numFieldNormedType.Exports.
+
+Set Implicit Arguments.
+Unset Strict Implicit.
+Unset Printing Implicit Defensive.
+
+Open Scope ring_scope.
+
+Section real_inverse_functions.
+
+Variable R : realType.
+
+(* Maybe this belongs in normedtype. *)
+Lemma near_in_interval (a b : R) :
+  {in `]a, b[, forall y, \forall z \near y, z \in `]a, b[}.
+Proof.
+move=> y ayb; rewrite (near_shift 0 y).
+have minlt : 0 < Num.min (y - a) (b - y).
+  have : 0 < y - a by rewrite subr_gt0 (itvP ayb).
+  have : 0 < b - y by rewrite subr_gt0 (itvP ayb).
+  by case: (ltrP (y - a) (b - y)).
+near=> z; rewrite /= subr0.
+rewrite in_itv /= -ltr_subl_addl -ltr_subr_addl ltr_normlW /=; last first.
+  rewrite normrN.
+  by near: z; apply: nbhs0_lt; rewrite (lt_le_trans minlt) // le_minl lexx.
+rewrite -ltr_subr_addr ltr_normlW //.
+near: z; apply: nbhs0_lt; rewrite (lt_le_trans minlt) //.
+by rewrite le_minl lexx orbT.
+Unshelve. all: end_near. Qed.
+
+
+Lemma near_0_in_interval (a b : R) :
+  {in `]a, b[, forall y, \forall z \near 0 : R, (z + y : R) \in `]a, b[}.
+Proof.
+move=> y ayb.
+rewrite (near_shift y 0); near=> z; rewrite /= sub0r subrK; near: z.
+by rewrite near_simpl; apply: near_in_interval.
+Unshelve. all: end_near. Qed.
+
+Lemma inverse_increasing_continuous (a b k : R) (f g : R -> R) :
+  0 < k ->
+  {in (g @` [set x | x \in `]a, b[ ])%classic &,
+       forall x y, x < y -> k * (y - x) < f y - f x} ->
+  {in `]a, b[, cancel g f} ->
+  {in `]a, b[, continuous g}.
+Proof.
+move=> kgt0 incrf gK y ayb.
+apply/cvg_distP=> _ /posnumP[e].
+have main1 : \forall x \near y, g y < g x -> `|g y - g x| < e%:num.
+  rewrite (near_shift 0 y); near=> z; rewrite /=subr0.
+  move=> yltz; rewrite ltr0_norm ?opprB; last by rewrite subr_lt0.
+  move: yltz.
+  have zin : z + y \in `]a, b[.
+    by near:z; rewrite near_simpl; apply: near_0_in_interval.
+  have gzin : g (z + y) \in ([set g u | u in [set x | x \in `]a, b[]])%classic.
+    by rewrite inE /=; exists (z + y).
+  have gyin : g (y) \in ([set g u | u in [set x | x \in `]a, b[]])%classic.
+    by rewrite inE /=; exists y.
+  move/incrf=> /(_ gyin gzin); rewrite !gK //.
+  rewrite addrK -ltr_pdivl_mull //; move/lt_trans; apply.
+  rewrite mulrC lter_pdivr_mulr //; apply: ltr_normlW.
+  by near: z; apply: nbhs0_lt; apply: mulr_gt0.
+have main2 : \forall x \near y, g x < g y -> `|g y - g x| < e%:num.
+  rewrite (near_shift 0 y); near=> z; rewrite /= subr0.
+  move=> zlty; rewrite gtr0_norm; last by rewrite subr_gt0.
+  move: zlty.
+  have zin : z + y \in `]a, b[.
+    by near: z; rewrite near_simpl; apply: near_0_in_interval.
+  have gzin : g (z + y) \in ([set g u | u in [set x | x \in `]a, b[]])%classic.
+    by rewrite inE /=; exists (z + y).
+  have gyin : g (y) \in ([set g u | u in [set x | x \in `]a, b[]])%classic.
+    by rewrite inE /=; exists y.
+  move/incrf=> /(_ gzin gyin); rewrite !gK //.
+  rewrite opprD addrA addrAC addrN sub0r.
+  rewrite -ltr_pdivl_mull //; move/lt_trans; apply.
+  rewrite mulrC lter_pdivr_mulr //; apply: ltr_normlW; rewrite normrN.
+  by near: z; apply: nbhs0_lt; apply: mulr_gt0.
+rewrite !near_simpl /=; near=> x.
+have [ | ] := boolP (g x < g y).
+  near:x; rewrite near_simpl; apply: main2.
+rewrite -leNgt le_eqVlt=> /orP[/eqP -> | gyltgx].
+  by rewrite subrr normr0.
+move: gyltgx; near:x; rewrite near_simpl; apply: main1.
+Unshelve. all: end_near. Qed.
+
+Lemma sqrt_continuous : continuous (@Num.sqrt R).
+Proof.
+move=> x.
+have [xlt0 | ] := boolP(x < 0).
+  apply: (near_cst_continuous 0).
+  rewrite (near_shift 0 x).
+  near=> z; rewrite subr0 /=; apply: ltr0_sqrtr.
+  rewrite -(opprK x) subr_lt0; apply: ltr_normlW.
+  by near: z; apply: nbhs0_lt; rewrite ltr_oppr oppr0.
+have neg_part (y : R) (e : {posnum R}) : y < 0 -> Num.sqrt y < e%:num.
+  by move=> ylt0; rewrite ltr0_sqrtr.
+rewrite -leNgt le_eqVlt eq_sym=> /orP[/eqP -> | ].
+  apply/cvg_distP=> _ /posnumP[e]; rewrite !near_simpl /=.
+  near=> y; rewrite sqrtr0 sub0r normrN.
+  rewrite ger0_norm ?sqrtr_ge0 //.
+  have ylte2 : y < e%:num ^+ 2.
+    rewrite ltr_normlW //; near: y; apply: nbhs0_lt.
+    by rewrite exprn_gt0.
+  have twogt0 : (0 < 2)%N by [].
+  rewrite -(ltr_pexpn2r twogt0) ?inE ?nnegrE ?ltrW ?sqrtr_ge0 //.
+  have [ylt0 | ] := boolP(y < 0).
+    by rewrite ltr0_sqrtr // expr0n /= exprn_gt0.
+  by rewrite -leNgt => yge0; rewrite sqr_sqrtr.
+suff main : forall a b, 0 < a -> {in `]a, b[, continuous (@Num.sqrt R)}.
+  move=> xgt0.
+  have halfxgt0 : 0 < x / 2 by apply: divr_gt0.
+  have xgthalf : x / 2 < x by case: (midf_lt xgt0); rewrite add0r.
+  have xltxp1 : x < x + 1 by rewrite ltr_addl.
+  set a := x / 2; set b := x + 1.
+  by apply: (main a b); rewrite // in_itv //= xgthalf xltxp1.
+move=> a b agt0.
+set k := Num.sqrt a.
+set f := fun u : R => u ^+ 2.
+set g := @Num.sqrt R.
+have sagt0 : 0 < Num.sqrt a by rewrite sqrtr_gt0.
+have incr : {in [set g u | u in [set x1 | x1 \in `]a, b[]]%classic &,
+         forall u y : R, u < y -> k * (y - u) < f y - f u}.
+  move=> u v; rewrite !inE /= => [][u' u'in u'q][v' v'in v'q] ultv.
+  have u'gt0 : 0 < u'.
+    by rewrite (lt_trans agt0) // (itvP u'in).
+  have ugt0 : 0 < u.
+    by rewrite (lt_trans sagt0) // -u'q /g ltr_sqrt // (itvP u'in).
+  have v'gt0 : 0 < v'.
+    by rewrite (lt_trans agt0) // (itvP v'in).
+  have vgt0 : 0 < v.
+    by rewrite (lt_trans sagt0) // -v'q /g ltr_sqrt // (itvP v'in).
+  rewrite /f subr_sqr (mulrC (_ - _)) ltr_pmul2r; last by rewrite subr_gt0.
+  rewrite ltr_spsaddr // -?u'q -?u'v /g ?sqrtr_gt0.
+  by rewrite /k -v'q /g ltr_sqrt //; case/andP: v'in.
+have gK : {in `]a, b[, cancel g f}.
+  by move=> u uin; rewrite /f /g sqr_sqrtr ?ltW ?(lt_trans agt0) ?(itvP uin).
+apply: (inverse_increasing_continuous sagt0 incr gK).
+Unshelve. all: end_near. Qed.
+
+End real_inverse_functions.
--- a/_CoqProject
+++ b/_CoqProject
@@ -26,10 +26,12 @@ theories/forms.v
 theories/derive.v
 theories/measure.v
 theories/summability.v
+theories/inverse_continuous.v
 theories/altreals/xfinmap.v
 theories/altreals/discrete.v
 theories/altreals/realseq.v
 theories/altreals/realsum.v
 theories/altreals/distr.v
+
 
 -R theories mathcomp.analysis


thery:
I am looking for the inverse of `ln` and I need something like : 
```
Lemma continuous_inverse (f g : R -> R) x :  
  (\forall z \near x, g (f z) = z)  ->
  (\forall z \near x, {for z, continuous f})  ->
  {for x, continuous g}.
```
Any idea how this could be proved?

thery:
Great! Do you think it would be possible to prove the generic lemma once and forall and not have
to prove the increasing bit for each injective continuous function?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,329 @@
...
+have xab : x \in `[a, b] by rewrite strict_to_large_itv // in_itv /= aLx.
+have ifab : If f a b.
+  by rewrite /If=> u v uin vin fufv; rewrite -(fK u uin) fufv fK.
+have [incr | abs] := near_injective_monotone ifab ctf; last first.
+  move: (abs a b); rewrite !in_itv /= !lexx ltW // ltW // leNgt aLb.
+  by move=> /(_ isT isT).
+apply/cvg_distP=> _ /posnumP[e].
+have /andP [xmex xxpe] : x - e%:num < x < x + e%:num.
+  by rewrite ltr_subl_addr cpr_add; apply/andP.
+set c := Num.max a (x - e%:num); set d := Num.min (x + e%:num) b.
```
ybertot, line 210: It would be better to have a wlog step 4 lines above, where you show that e can be taken small enough for `` `]x-e, x+e[ `` to be inside `` `]a,b[``

ybertot:
The failures seem to be unrelated  to the content of the PR?  Is it possible to restart the compilation?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  have fK' : {in `[a, b], cancel (-%R \o f)(g \o -%R)}.
+    by move=> x; rewrite /= opprK; exact: fK.
+  suff ct_gopp : {in `](-f a), (-f b)[, continuous (g \o -%R)}.
+    rewrite (_ : g = (g \o -%R) \o -%R); last first.
+      by apply: funext => x; rewrite /= opprK.
+    move=> x xin; apply: continuous_comp.
+      by rewrite forE; apply: opp_continuous.
+    rewrite forE; apply: ct_gopp.
+    by rewrite oppr_itvoo !opprK.
+  move=> x xin; apply: main => //.
```
thery, line 343: ```move=> x xin; apply: main => //; first by rewrite /= ltr_oppr opprK.```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  rewrite -(fK a) ?in_itv /= ?lexx ?(ltW aLb) //.
+  by rewrite abs ?(ltW faLfb) // in_itv /= lexx (ltW faLfb).
+have gK := interval_injective_continuous_bijective aLb ctf incrf fK.
+by apply: monotone_surjective_continuous.
+Qed.
+
+Lemma sqr_continuous : continuous (exprz (R := R) ^~ 2).
+Proof.
+move => x s.
+rewrite (_ : exprz (R := R) ^~ 2 = fun x : R => x * x); last first.
```
thery, line 374:
```
rewrite (_ : (fun x => x ^ 2) = (fun x => x * x)); last first.
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  by  apply: funext=> y; rewrite exprSz expr1z.
+rewrite exprSz expr1z; move: x s.
+change (continuous ((fun x : R * R => x.1 * x.2) \o (fun u => (u, u)))).
```
thery, line 377: simply `by apply: continuousM.`

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+Definition strict_increasing (f : R -> R) :=
+  forall x y, x < y -> f x < f y.
+
+Definition monotone (f : R -> R) :=
+  {mono f : y z / y <= z} \/ {mono f : y z /~ y <= z}.
+
+Let Cf (f : R -> R) a b := {in `[a, b], continuous f}.
+Let If (f : R -> R) a b := {in `[a, b] &, injective f}.
+Let Mf (f : R -> R) a b := {in `[a, b] &, {mono f : x y / x <= y}}.
+
```
thery, line 28: Maybe also add `Let MTf (f : R -> R) a b := 
  {in `[a, b] &, {mono f : x y / x <= y}} \/
  {in `[a, b] &, {mono f : x y /~ x <= y}}.`

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+Lemma near_injective_monotone (f : R -> R) (a b : R) :
+  If f a b -> Cf f a b ->
+  {in `[a, b] &, {mono f : x y / x <= y}} \/
```
thery, line 31: ```MTf f a b```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  apply: IVT => //; first by move=> d dIab; apply/fC/acIab.
+  by case: ltrgtP faLfc; rewrite // faLfb // ltW.
+move=> fdEfb; move: fbLfc.
+have : d <= c <= b by rewrite cLb  (itvP dI).
+rewrite (fI _  _ _ _ fdEfb) ?in_itv /= ?lexx ?(itvP dI) ?aLb //.
+  by case: (ltrgtP b) => //= ->; rewrite ltxx.
+by rewrite (le_trans _ cLb) //  (itvP dI).
+Qed.
+
+Lemma strict_to_large_itv (a b x : R) :
```
thery, line 105: on one line

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  x \in `]a, b[ -> x \in `[a, b].
+Proof.
+by rewrite !in_itv /= => /andP[aLx xLb]; rewrite !le_eqVlt aLx xLb !orbT.
+Qed.
+
+(* This belongs in mathcomp *)
+Lemma oppr_min (x y : R) : - Num.min x y = Num.max (-x) (- y).
+Proof.
```
thery, line 113: Proof on one line

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+by rewrite /Num.max ltr_oppr !opprK; case: (ltrgtP x y).
+Qed.
+
+Lemma oppr_max (x y : R) : - Num.max x y = Num.min (-x) (- y).
+Proof.
```
thery, line 118: Proof on one line

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+move=> y ayb.
+rewrite (near_shift y 0); near=> z; rewrite /= sub0r subrK; near: z.
+by rewrite near_simpl; apply: near_in_interval.
+Grab Existential Variables. all: end_near. Qed.
+
+Lemma inverse_monotone (a b : R) (f g : R -> R) :
+  a < b ->
+  {in `[a, b], continuous f} ->
+  {in `[a, b], cancel f g} ->
+  {in `[(Num.min (f a) (f b)), (Num.max (f a) (f b))] &, {mono g : x y / x <= y}} \/
```
thery, line 152: ``` MTf g (Num.min (f a) (f b)) (Num.max (f a) (f b))```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+ {in `[a, b], cancel f g} ->
+ {in `[(f a), (f b)], cancel g f}.
+Proof.
+move=> aLb ctf monf fK y yin.
+have faLfb : f a <= f b by rewrite monf ?in_itv /= ?lexx ltW.
+have := IVT (ltW aLb) ctf; case: (lerP (f a) (f b)) (faLfb)=> // _ _=> ivt.
+by case: (ivt _ yin)=> x xin <-; rewrite fK.
+Qed.
+  
+Lemma near_opp (a : R) (P : R -> Prop) :
```
thery, line 207: nearN?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  rewrite ltNge -monf //.
+    by rewrite -ltNge gK //.
+  rewrite in_itv /= -(fK a) ?mong // ltW //=.
+    by rewrite (le_trans guLgy) // ltW.
+  by rewrite (le_lt_trans faLfgyme fgymeLu).
+near=> u; apply: below.
+suff : u \in `]f (g y - e%:num), (f b)[ by move=> uin; rewrite (itvP uin).
+near: u; apply: near_in_interval; rewrite in_itv /= ?(itvP yin) andbT.
+rewrite -[X in _ < X](gK y); last by rewrite strict_to_large_itv.
+rewrite ltNge monf //.
```
thery, line 313:
```rewrite ltNge monf -?ltNge //; first by rewrite in_itv /= (ltW aLgy) (ltW gyLb).
```

thery (commented):
seems ok.
Still have to understand how to link it to [this](https://github.com/thery/analysis/blob/1e98b26313131c612074fc3feccdd6c4852c4349/theories/exp.v#L205)

thery:
Maybe an alternative formulation of the ffrst theorem is :
```
{near a, injective f} ->
(\near a, continuous f} ->
{near a, {mono f : x y / x <= y}} \/ {near a, {mono f : x y /~ x <= y}}
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  forall x y, x < y -> f x < f y.
+
+Definition monotone (f : R -> R) :=
+  {mono f : y z / y <= z} \/ {mono f : y z /~ y <= z}.
+
+Let Cf (f : R -> R) a b := {in `[a, b], continuous f}.
+Let If (f : R -> R) a b := {in `[a, b] &, injective f}.
+Let Mf (f : R -> R) a b := {in `[a, b] &, {mono f : x y / x <= y}}.
+
+Lemma near_injective_monotone (f : R -> R) (a b : R) :
```
CohenCyril, line 29: This lemma is misnamed since it does not use `near` at all: maybe `segment_inj_monotone` ?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+  by have := lt_trans fbLfc fcLfa; rewrite ltNge faLfb.
+case: (ltrgtP (f b) (f c))=> //= fbLfc.
+have [d dI]: exists2 d, d \in `[a, c] & f d = f b.
+  apply: IVT => //; first by move=> d dIab; apply/fC/acIab.
+  by case: ltrgtP faLfc; rewrite // faLfb // ltW.
+move=> fdEfb; move: fbLfc.
+have : d <= c <= b by rewrite cLb  (itvP dI).
+rewrite (fI _  _ _ _ fdEfb) ?in_itv /= ?lexx ?(itvP dI) ?aLb //.
+  by case: (ltrgtP b) => //= ->; rewrite ltxx.
+by rewrite (le_trans _ cLb) //  (itvP dI).
```
CohenCyril, line 102: This proof can be compressed, I will suggest something later, but if I do not it should not delay the merge.

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+Qed.
+
+Lemma strict_to_large_itv (a b x : R) :
+  x \in `]a, b[ -> x \in `[a, b].
+Proof.
+by rewrite !in_itv /= => /andP[aLx xLb]; rewrite !le_eqVlt aLx xLb !orbT.
+Qed.
+
+(* This belongs in mathcomp *)
```
CohenCyril, line 111: Please open a PR in mathcomp so that it is merged for 1.13.0 :smile:

theories/inverse_continuous.v
```diff
@@ -0,0 +1,419 @@
...
+by move: zclose; rewrite /ball_ /= -opprD normrN opprK.
+Qed.
+
+Lemma monotone_surjective_continuous (a b : R) (f g : R -> R) :
+  a < b ->
+  {in `[a, b] &, {mono f : x y / x <= y}} ->
+  {in `[(f a), (f b)] &, {mono g : x y / x <= y}} ->
+  {in `[a, b], cancel f g} ->
+  {in `[(f a), (f b)], cancel g f} ->
+  {in `](f a), (f b)[ , continuous g}.
```
CohenCyril, line 222:
```suggestion
  {in `](f a), (f b)[, continuous g}.
```

CohenCyril (changes requested):
Great contribution.
Please add all of this to `normed_type.v`
I think the lemmas with `\/` are difficult to use, we should also provide the two alternatives for each: one with hypothesis `f a <= f b` and one with `f a >= f b`.

CohenCyril:
> Maybe an alternative formulation of the ffrst theorem is :
> 
> ```
> {near a, injective f} ->
> (near a, continuous f} ->
> {near a, {mono f : x y / x <= y}} \/ {near a, {mono f : x y /~ x <= y}}
> ```

No, this lemma is less general, the naming is wrong, not the statement.

CohenCyril:
> > Maybe an alternative formulation of the ffrst theorem is :
> > ```
> > {near a, injective f} ->
> > (near a, continuous f} ->
> > {near a, {mono f : x y / x <= y}} \/ {near a, {mono f : x y /~ x <= y}}
> > ```
> 
> No, this lemma would less general, the naming is wrong, not the statement.

Of course, we could/should also have the "near version", which should be an almost trivial specialization of the one on a segment.

thery:
> I think the lemmas with `\/` are difficult to use, we should also provide the two alternatives for each: one with hypothesis `f a <= f b` and one with `f a >= f b`.

This is the way the actual proof works, It proves f a <= f b and then prove the other taking -f

CohenCyril:
> > I think the lemmas with `\/` are difficult to use, we should also provide the two alternatives for each: one with hypothesis `f a <= f b` and one with `f a >= f b`.
> 
> This is the way the actual proof works, It proves f a <= f b and then prove the other taking -f

I know ;)

thery:
so stating the two alternatives before should simplify the proof.

theories/inverse_continuous.v
```diff
@@ -0,0 +1,481 @@
...
+  move: (aLb); rewrite ltNge -(fK b) ?in_itv /= ?lexx ?(ltW aLb) //.
+  rewrite -(fK a) ?in_itv /= ?lexx ?(ltW aLb) //.
+  by rewrite abs ?(ltW faLfb) // in_itv /= lexx (ltW faLfb).
+have gK := interval_injective_continuous_bijective aLb ctf incrf fK.
+by apply: monotone_surjective_continuous.
+Qed.
+
+Lemma continuous_inverse  (f g : R -> R) x :
+  (\forall z \near x, g (f z) = z) ->
+  (\forall z \near x, {for z, continuous f}) ->
```
CohenCyril, line 373:
```suggestion
  {near x, cancel f g} ->
  {near x, continuous f} ->
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,481 @@
...
+  {for f x, continuous g}.
```
CohenCyril, line 374: I guess the conclusion could be strengthened to `{near f x, continuous g}`?

theories/inverse_continuous.v
ybertot: When looking for the place where this should be included, I found the following lemma
`real_oppr_min`  which expresses the required property in a more elegant way.  However, it feels strange that we need to prove the `_ \is real` property, while all elements of a realType are real (actually, all elements of a realDomainType).
Should the theorem `num_real` be included in the hint database for // (is there such a hint database)?

ybertot: Anyway, I just check oppr_min and oppr_max already exist.  I did not search for them in the right way (it did not come to my mind that they were morphism lemmas).

theories/inverse_continuous.v
```diff
@@ -0,0 +1,409 @@
...
+  by rewrite /If=> u v uin vin fufv; rewrite -(fK u uin) fufv fK.
+have [incrf | abs] := segment_injective_monotone ifab ctf; last first.
+  move: (abs a b); rewrite !in_itv /= !lexx ltW // ltW // leNgt aLb.
+  by move=> /(_ isT isT).
+have incrg := left_inverse_increasing aLb faLfb' ctf fK.
+have gK := interval_injective_continuous_bijective aLb ctf incrf fK.
+by apply: monotone_surjective_continuous.
+Qed.
+
+Lemma continuous_inverse  (f g : R -> R) (x : R) :
```
thery, line 306: Maybe one could use `segment_injective_monotone` instead of `wlog`

```
Lemma continuous_inverse  (f g : R -> R) (x : R) :
  {near x, cancel f g} -> {near x, continuous f} -> {near (f x), continuous g}.
Proof.
move=> [e egt0 fK] [e' e'gt0 ctf].
set e'' := Num.min e e' / 2.
have e''gt0 : 0 < e''.
  by apply: divr_gt0; [rewrite /Num.min; case: ifP | rewrite ltr0Sn].
have [e''lte e''lte'] : e'' < e /\ e'' < e'.
  have e''lt2 : e'' < e'' + e'' by rewrite ltr_addl //.
  rewrite !(lt_le_trans e''lt2) // /e'' -mulrDl (le_trans (midf_le _).2) //;
    by case: (lerP e e') => // strict; rewrite ltW.
have fK' : {in `[x - e'', x + e''], cancel f g}.
  move=> y; rewrite in_itv /= => /andP [yin1 yin2]; apply: fK=> /=.
  rewrite -opprB normrN real_ltr_distl ?num_real // (lt_le_trans _ yin1) /=.
    by rewrite (le_lt_trans yin2) // ltr_add2.
  by rewrite ltr_add2 ltr_oppl opprK.
have ctf' : {in `[x - e'', x + e''], continuous f}.
  move=> y; rewrite in_itv /= => /andP [yin1 yin2]; apply: ctf => /=.
  rewrite -opprB normrN real_ltr_distl ?num_real // (lt_le_trans _ yin1) /=.
    by rewrite (le_lt_trans yin2) // ltr_add2.
  by rewrite ltr_add2 ltr_oppl opprK.
have cmp1 : x - e'' < x by rewrite ltr_subl_addr ltr_addl.
have cmp2 : x < x + e'' by rewrite ltr_addl.
have cmp : x - e'' < x + e'' by rewrite (lt_trans cmp1).
have ifx : If f (x - e'') (x + e'').
  by move=> v w vin win fq; rewrite -(fK' v) // fq fK'.
have [mfx |mfx] := segment_injective_monotone ifx ctf'.
  near=> z; apply: inverse_increasing_continuous ctf' _ _ _ => //.
    by rewrite mfx ?ltW // in_itv /= ?lexx ltW.
  near: z; rewrite near_simpl; apply: near_in_interval.
  by rewrite in_itv /= !(ltNge, mfx, in_itv) /= -?ltNge ?(lexx, cmp1, ltW).
near=> y; rewrite -[y]opprK (_ : g = (g \o -%R \o -%R)); last first.
  by apply: funext=> u; rewrite /= opprK.
apply: continuous_comp; rewrite opprK; first by apply: continuousN.
have ctNf: {in `[(x - e''), (x + e'')], continuous (- f)}.
  by move=> z zI; apply/continuousN/ctf'.
apply: inverse_increasing_continuous ctNf _ _ _ => //.
- by rewrite lter_oppr opprK mfx ?ltW // in_itv /= ?lexx ltW.
- by move=> z zI; rewrite /= opprK fK'.
rewrite oppr_itvoo !opprK.
near: y; rewrite near_simpl; apply: near_in_interval.
by rewrite in_itv /= !(ltNge, mfx, in_itv) /= -?ltNge ?(lexx, cmp2, ltW).
Grab Existential Variables. all:end_near. Qed.
```

thery:
Here is the other proof I need

```
Lemma inverse_swap_continuous  (f g : R -> R) (x : R) :
  {near x, cancel f g} -> {near x, continuous f} -> {near (f x), cancel g f}.
Proof.
move=> [e egt0 fK] [e' e'gt0 ctf].
set e'' := Num.min e e' / 2.
have e''gt0 : 0 < e''.
  by apply: divr_gt0; [rewrite /Num.min; case: ifP | rewrite ltr0Sn].
have [e''lte e''lte'] : e'' < e /\ e'' < e'.
  have e''lt2 : e'' < e'' + e'' by rewrite ltr_addl //.
  rewrite !(lt_le_trans e''lt2) // /e'' -mulrDl (le_trans (midf_le _).2) //;
    by case: (lerP e e') => // strict; rewrite ltW.
have fK' : {in `[x - e'', x + e''], cancel f g}.
  move=> y; rewrite in_itv /= => /andP [yin1 yin2]; apply: fK=> /=.
  rewrite -opprB normrN real_ltr_distl ?num_real // (lt_le_trans _ yin1) /=.
    by rewrite (le_lt_trans yin2) // ltr_add2.
  by rewrite ltr_add2 ltr_oppl opprK.
have ctf' : {in `[x - e'', x + e''], continuous f}.
  move=> y; rewrite in_itv /= => /andP [yin1 yin2]; apply: ctf => /=.
  rewrite -opprB normrN real_ltr_distl ?num_real // (lt_le_trans _ yin1) /=.
    by rewrite (le_lt_trans yin2) // ltr_add2.
  by rewrite ltr_add2 ltr_oppl opprK.
have cmp1 : x - e'' < x by rewrite ltr_subl_addr ltr_addl.
have cmp2 : x < x + e'' by rewrite ltr_addl.
have cmp : x - e'' < x + e'' by rewrite (lt_trans cmp1).
have ifx : If f (x - e'') (x + e'').
  by move=> v w vin win fq; rewrite -(fK' v) // fq fK'.
have [mfx |mfx] := segment_injective_monotone ifx ctf'.
  near=> z.
  apply: interval_injective_continuous_bijective _ ctf' _ _ _ _ => //.
  apply: strict_to_large_itv.
  near: z; rewrite near_simpl; apply: near_in_interval.
  by rewrite in_itv /= !(ltNge, mfx, in_itv) /= -?ltNge ?(lexx, cmp1, ltW).
near=> y; suff : ((f \o -%R) \o -g) y = y by rewrite /= opprK.
have ctNf' : {in `[ (- x) - e'', (-x) + e''], continuous (f \o -%R)}.
  move=> z /itvP zI; apply: continuous_comp; first by apply: continuousN.
  apply: ctf'.
  by rewrite in_itv /= lter_oppr opprD opprK zI lter_oppl opprD zI.
apply: interval_injective_continuous_bijective _ ctNf' _ _ _ _.
- by rewrite -opprD lter_oppl opprD opprK.
- move=> z1 z2 /itvP z1I /itvP z2I.
  by rewrite mfx // 1?lter_oppl ?opprK // in_itv /= lter_oppl lter_oppr !opprD
             opprK ?(z1I, z2I).
- move=> z /itvP zI; rewrite -[RHS](opprK z); congr (- _).
  apply: fK'.
  by rewrite in_itv /= lter_oppr opprD opprK zI lter_oppl opprD zI.
apply: strict_to_large_itv.
near: y; rewrite near_simpl; apply: near_in_interval.
rewrite in_itv /= !(opprD, opprK).
by rewrite !(ltNge, mfx, in_itv) /= -?ltNge ?(lexx, cmp2) // ltW // ltW.
Grab Existential Variables. all:end_near. Qed.
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,452 @@
...
+  apply: triplet_injective_increasing
+     => [|x1 x2 /itvP x1I /itvP x2I |x1 /itvP x1I|] //.
+  - by apply: fI; rewrite in_itv /= (le_trans (_ : a <= y)) !(yI, x1I, x2I).
+  - by apply: fC; rewrite in_itv /= (le_trans (_ : a <= y)) !(yI, x1I).
+  by rewrite yLx xI.
+have : f x == f y -> x == y.
+  by move=> /eqP/fI-> //; rewrite in_itv /= !(xI, yI).
+by case: (ltrgtP x y); case: (ltrgtP (f x) (f y)) => // _ _ H1 H2 H3;
+     (case: (H1 isT) || case: (H2 isT) || case: (H3 isT)).
+Qed.
```
affeldt-aist, line 90: What about using `total_homo_mono_in`?

```coq
Lemma segment_injective_increasing' (f : R -> R) (a b : R) :
  f a <= f b -> If f a b -> Cf f a b -> {in `[a, b] &, {homo f : x y / x < y}}.
Proof.
move=> fab fI fC x y xI yI; rewrite lt_neqAle => /andP[xy yx].
rewrite lt_neqAle; apply/andP; split; first by apply: contra xy => /eqP/fI ->.
have /andP[fax fxb] : f a <= f x <= f b by exact: triplet_injective_increasing.
suff /andP[-> //] : f x <= f y <= f b.
apply: triplet_injective_increasing => [|u v /itvP uI /itvP vI |z /itvP uI|] //.
- by apply: fI; rewrite in_itv /= (le_trans (_ : a <= x)) !(itvP xI, uI, vI).
- by apply: fC; rewrite in_itv /= (le_trans (_ : a <= x)) !(itvP xI, uI).
- by rewrite yx (itvP yI).
Qed.

Lemma segment_injective_increasing (f : R -> R) (a b : R) :
  f a <= f b -> If f a b -> Cf f a b -> Mf f a b.
Proof.
move=> fab fI fC.
by apply: total_homo_mono_in (segment_injective_increasing' fab fI fC) => //;
  [exact: lt_neqAle | exact: lt_neqAle | exact: le_anti | exact: le_total].
Qed.
```

affeldt-aist (commented):
It removes one redundancy (and warnings but this is unrelated) but I am not sure about the naming of the intermediate lemma.

affeldt-aist:
> It removes one redundancy (and warnings but this is unrelated) but I am not sure about the naming of the intermediate lemma.

Actually, `segment_injective_increasing'` might be `segment_injective_nondecreasing`
like in `sequences.v` where there is a `nondecreasing_seq` notation for the `homo` notation.

theories/inverse_continuous.v
```diff
@@ -0,0 +1,458 @@
...
+right => x y xI yI.
+suff : (- f) y <= (- f) x = (y <= x) by rewrite ler_oppl opprK.
+apply: segment_injective_increasing xI => // [|x1 x2 x1I y1I U| x1 x1I].
+- by rewrite ler_oppl opprK.
+- by apply: fI => //; rewrite -[LHS]opprK [- f _]U opprK.
+by apply/continuousN/fC.
+Qed.
+
+Lemma strict_to_large_itv (a b x : R) : x \in `]a, b[ -> x \in `[a, b].
+Proof. by rewrite !in_itv /= => /andP[A B]; rewrite !le_eqVlt A B !orbT. Qed.
```
affeldt-aist, line 108: What about using the `subset` notation?

```coq
Lemma strict_to_large_itv (a b : R) : {subset `]a, b[ <= `[a, b]}.
Proof. by move=> x; rewrite !in_itv /= => /andP[/ltW -> /ltW ->]. Qed.
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,458 @@
...
+  by have := lt_trans fbLfc fcLfa; rewrite ltNge faLfb.
+case: (ltrgtP (f b) (f c))=> //= fbLfc.
+have [d dI]: exists2 d, d \in `[a, c] & f d = f b.
+  apply: IVT => //; first by move=> d dIab; apply/fC/acIab.
+  by case: ltrgtP faLfc; rewrite // faLfb // ltW.
+move=> fdEfb; move: fbLfc.
+have : d <= c <= b by rewrite cLb  (itvP dI).
+rewrite (fI _  _ _ _ fdEfb) ?in_itv /= ?lexx ?(itvP dI) ?aLb //.
+  by case: (ltrgtP b) => //= ->; rewrite ltxx.
+by rewrite (le_trans _ cLb) //  (itvP dI).
```
affeldt-aist, line 57: I tried to use `subset_prop_in1` and make it shorter but the gain is not obvious. I let it here for the record. 

```coq
move=> faLfb fI fC /andP[aLc cLb].
have [faLfc|fcLfa|faEfc] /= := ltrgtP (f a) (f c).
- rewrite leNgt; apply/negP => fbLfc.
  move: (fbLfc); suff /eqP -> : c == b by rewrite ltxx.
  rewrite eq_le cLb /=.
  have [d /itvP dI fdEfb] : exists2 d, d \in `[a, c] & f d = f b.
    apply: IVT => //; last by case: ltrgtP faLfc; rewrite // faLfb // ltW.
    apply: subset_prop_in1 fC.
    by move=> x; rewrite !in_itv => /andP[->] /le_trans; exact.
  suff <- : d = b by rewrite dI.
  by apply: fI fdEfb; rewrite in_itv/= ?(lexx,dI,le_trans aLc,le_trans _ cLb).
- have [fbLfc|fcLfb|fbEfc] /= := ltrgtP (f b) (f c).
  + by have := lt_trans fbLfc fcLfa; rewrite ltNge faLfb.
  + move: (fcLfa); suff /eqP -> : c == a by rewrite ltxx.
    rewrite eq_le aLc andbT.
    have [d /itvP dI fdEfa] : exists2 d, d \in `[c, b] & f d = f a.
      apply: IVT => //; last by case: ltrgtP fcLfb => // _ _; rewrite ltW.
      apply: subset_prop_in1 fC.
      by move=> x; rewrite !in_itv => /andP[/(le_trans aLc) /= ->].
    suff <- : d = a by rewrite dI.
    by apply: fI fdEfa; rewrite in_itv/= ?(lexx,dI,le_trans aLc).
  + by move: fcLfa; rewrite -fbEfc ltNge faLfb.
- by rewrite -(fI _  _ _ _ faEfc) // in_itv /= ?lexx (le_trans aLc).
```

theories/inverse_continuous.v
CohenCyril: I suggest to inline the former inside the latter.
If there is an additional lemma to add it should be:
```coq
forall (f : R -> R) (a b : R), f a <= f b -> If f a b -> Cf f a b -> {in `[a, b] &, {mono f : x y / x < y}}.
```
which is a oneliner from the latter.

theories/inverse_continuous.v
CohenCyril: Normally this kind of goal used to be closed by `apply/subsetP; rewrite !ltW.` but it looks like that inequalities on bounds do not unfold anymore with mathcomp 1.12 ... @pi8027 do you have a solution for that?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,458 @@
...
+
+Let Cf (f : R -> R) a b := {in `[a, b], continuous f}.
+Let If (f : R -> R) a b := {in `[a, b] &, injective f}.
+Let Mf (f : R -> R) a b := {in `[a, b] &, {mono f : x y / x <= y}}.
+Let MTf (f : R -> R) a b :=
+  {in `[a, b] &, {mono f : x y / x <= y}} \/
+  {in `[a, b] &, {mono f : x y /~ x <= y}}.
+
+Lemma triplet_injective_increasing (f : R -> R) (a c b : R) :
+  f a <= f b -> If f a b -> Cf f a b -> a <= c <= b -> f a <= f c <= f b.
```
CohenCyril, line 27: I believe this is an internal lemma to `segment_injective_increasing` and should be inlined, because it is subsumed by it.

theories/inverse_continuous.v
ybertot: An alternative proof goes through `subitvP` : 

    Proof. by apply: subitvP; rewrite subitvE !bound_lexx. Qed.

I am scared by how difficult it was for me to find `bound_lexx`.  This is yet another example of the need to improve the Search command drastically.

By the way, How do you make that your code is colorized?

affeldt-aist: By starting with 
```
```coq
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,458 @@
...
+  apply: IVT => //; first by move=> d dIab; apply/fC/acIab.
+  by case: ltrgtP faLfc; rewrite // faLfb // ltW.
+move=> fdEfb; move: fbLfc.
+have : d <= c <= b by rewrite cLb  (itvP dI).
+rewrite (fI _  _ _ _ fdEfb) ?in_itv /= ?lexx ?(itvP dI) ?aLb //.
+  by case: (ltrgtP b) => //= ->; rewrite ltxx.
+by rewrite (le_trans _ cLb) //  (itvP dI).
+Qed.
+
+Lemma segment_injective_increasing (f : R -> R) (a b : R) :
```
CohenCyril, line 60: What about `continuous_inj_le_itvcc` (cf [`inj_eq`](https://github.com/math-comp/math-comp/blob/1c4a6d3ca17820f6c60f194ffa845f8c007c9d8f/mathcomp/ssreflect/eqtype.v#L423-L424))
I think for the sake of completeness, the theorem should actually be provided on any interval...

CohenCyril:
I think all the shorthands that appear in statements should be removed from statements: I find `If`, `Cf`, and `Mf` unreadable. Please replace them by the body of the let in...

theories/inverse_continuous.v
ybertot: Thanks, I am integrating this variant in the proof, together with more uses of `subitvP` `subitvE`

theories/inverse_continuous.v
ybertot: Just did the inlining.

theories/inverse_continuous.v
ybertot: Is this a naming suggestion?

ybertot: Anyway, I am not sure it holds for open intervals.

CohenCyril: > Anyway, I am not sure it holds for open intervals.

I'm quite sure it does. The problem might be the condition `f a <= f b`, which you do not want to consider.
I think this is actually a flaw of the current phrasing. Maybe a precondition of the form `(exists x y, [x \in I, y \in I, x < y & f x <= f y])` would actually be more flexible and desirable than the specialization to `a` and `b` which does not make sense when a bound is excluded or infinite)

￼

CohenCyril: > I think the lemmas with `\/` are difficult to use, we should also provide the two alternatives for each: one with hypothesis `f a <= f b` and one with `f a >= f b`.

Also when I said this, I did not mean to have you remove the statements with the `\/`, just derive them from the other ones.

ybertot:
I did the replacement of If Cf and Mf, but only in main statements.  Inside proofs, I kept them.  Is that ok?

theories/inverse_continuous.v
```diff
@@ -0,0 +1,556 @@
...
+- move=> _ _ _ x y; rewrite in_itv /= => /andP[aLx xLb]; move: bLa.
+  by rewrite ltNge (le_trans aLx).
+move=> _ _ _ x y; rewrite !in_itv /= -ab=> /le_anti -> /le_anti ->.
+by rewrite !lexx.
+Qed.
+
+Lemma segment_continuous_inj_monotone (f : R -> R) (a b : R) :
+  {in `[a, b], continuous f} -> {in `[a, b] &, injective f} ->
+  {in `[a, b] &, {mono f : x y / x <= y}} \/
+  {in `[a, b] &, {mono f : x y /~ x <= y}}.
```
CohenCyril, line 196: This can also be generalized to any interval.

ybertot: done, but I have difficulties with name choice.  Also should we keep the ones with f a <= f b where a and b are extremities of a closed interval, or should this lemma be removed?

theories/inverse_continuous.v
```diff
@@ -20,61 +20,45 @@ Let Cf (f : R -> R) a b := {in `[a, b], continuous f}.
...
-have /subitvP : (`[a, b] <= I)%O by rewrite itvcc_le ?aI ?(le_trans aLc cLb).
-by apply; rewrite in_itv/= ?aLc.
+move=> aI bI; rewrite !le_eqVlt.
+move=> /andP[] /orP[/eqP <- // | aLc] /orP [/eqP -> // | cLb].
+by rewrite (interval_is_interval aI bI) ?aLc.
 Qed.
 
 Lemma continuous_inj_le_itv (f : R -> R) (I : interval R) :
-  (exists x y, x \in I /\ y \in I /\ x < y /\ f x <= f y) ->
+  (exists x y, (x \in I) && (y \in I) && (x < y) && (f x <= f y)) ->
```
CohenCyril, line 59: Can you rarher use a quaternary conjuction? `[/\ ..., ..., & ...]`

theories/inverse_continuous.v
```diff
@@ -61,16 +61,17 @@ Lemma continuous_inj_le_itv (f : R -> R) (I : interval R) :
   {in I &, {mono f : x y / x <= y}}.
 Proof.
 move: f.
+  have fxy (f : R -> R) : {in I &, injective f} -> forall x y,
+    x\in I -> y\in I -> x < y -> f x != f y.
```
CohenCyril, line 65:
```suggestion
    x \in I -> y \in I -> x < y -> f x != f y.
```

theories/inverse_continuous.v
```diff
@@ -91,77 +91,61 @@ have stepper_main (f : R -> R) (a c b : R) :
         have cLb' : c <= b by rewrite ltW.
         apply: IVT => //; last by case: ltrgtP fcLfb; rewrite // !ltW.
         apply: sub_in1 fC => y; rewrite in_itv /= => /andP[?].
```
CohenCyril, line 93: do not `?` unless the goal is closed in the same line.

theories/inverse_continuous.v
```diff
@@ -0,0 +1,551 @@
...
+                     ((b \in `[a, b]) = (a <= b)) *
+                     ((a \in `[a, b[) = (a < b)) *
+                     ((b \in `]a, b]) = (a < b)) *
+                     (a \in `[a, +oo[ ) *
+                     (a \in `]-oo, a]).
+Proof. by rewrite !(boundr_in_itv, boundl_in_itv). Qed.
+
+(* All uses of this lemma have been removed, replacing them with
+  uses of interval_is_interval, but it would be better if
+  is_interval used large comparisons instead of strict ones. *)
```
CohenCyril, line 36: https://github.com/math-comp/analysis/blob/804a5e060da46853ea14e29a720edab76ea79fa4/theories/normedtype.v#L2968-L2972

Gives the two characterization, one with strict inequalities, one with large. Both are useful depending on whether you want to prove or use.

theories/inverse_continuous.v
```diff
@@ -0,0 +1,551 @@
...
+  x <= y -> (`[x, y] <= I)%O = ((x \in I) && (y \in I)).
+Proof.
+by move=> C; case: I => [[[] l| []] [[|] u| [|]]];
+  rewrite !subitvE ?in_itv -?andbA //=; apply/idP/idP;
+  repeat (move/andP=>[] ?); move=> ?; repeat (apply/andP; split=> //);
+  rewrite ?(le_trans C) ?(le_trans _ C) ?(lt_le_trans _ C) ?(le_lt_trans C).
+Qed.
+
+(* This is just an example showing how to use interval_is_interval
+  in the case of large comparisons. the lemma is not used otherwise.
```
CohenCyril, line 47: Use `is_intervalPle` instead

theories/inverse_continuous.v
```diff
@@ -0,0 +1,551 @@
...
+- have fxLfu : f x < f u by have[] := high x u v fuLfv xI vI xLu ulv.
+  case: (ltrgtP y u) => [yLu | uLy | -> //]; first by case: (whole x y u).
+  by case: (low x u y).
+move: xLy; rewrite -xu => uLy.
+case: (ltrgtP y v)=> [yLv | vLy | -> //]; first by case: (whole u y v).
+by case: (low u v y).
+Qed.
+
+Lemma continuous_inj_ge_itv (f : R -> R) (I : interval R) :
+  (exists x y, (x \in I) && (y \in I) && (x < y) && (f y <= f x)) ->
```
CohenCyril, line 152: Use quaternary conjuction instead
```suggestion
  (exists x y, [/\ x \in I, y \in I, x < y & f y <= f x]) ->
```

theories/inverse_continuous.v
```diff
@@ -0,0 +1,551 @@
...
+(* This is an example that appears in the following proof.  I think
+  there could be a use case for tactic that leads the user to proving
+  a conjunction by addressing the second component with the assumption
+  that the first one already holds. *)
+Fact itv_example (x e  : R):
+  0 < e -> [&& x < x + e, x - e < x & x - e < x + e].
+Proof.
+move=> e0; staged_split;[rewrite ltr_addl // | rewrite ltr_subl_addr //| ].
+(* Unfortunately, the name of the hypotheses could be useful here. *)
+eapply lt_trans; eassumption.
```
CohenCyril, line 435: (untested)
```suggestion
move=> e0 [:xlt] [:xgt]; case/and3P; split.
- by abstract: xlt; rewrite ltr_addl.
- by abstract: xgt; rewrite ltr_subl_addr.
exact: lt_trans xgt.
```

theories/inverse_continuous.v
ybertot: I agree they are both useful, but by choosing to use the one with the "strongest" hypothesis by default, we make usage harder: is_intervalPle is a clumsy tool: implicit arguments don't do the right thing in this use case.

thery:
is it ready to merge?

theories/normedtype.v
```diff
@@ -3908,6 +3952,523 @@ Qed.
...
+  by rewrite ltNge (le_trans aLx).
+move=> _ _ _ x y; rewrite !in_itv /= -ab=> /le_anti -> /le_anti ->.
+by rewrite !lexx.
+Qed.
+  
+Let Cf (f : R -> R) a b := {in `[a, b], continuous f}.
+Let If (f : R -> R) a b := {in `[a, b] &, injective f}.
+Let Mf (f : R -> R) a b := {in `[a, b] &, {mono f : x y / x <= y}}.
+
+Lemma segment_left_inverse_increasing (a b : R) (f g : R -> R) :
```
thery, line 4129: I was wondering why there was not the assumption f a <= f b then I understood.
maybe a comment could be added.

thery: should we have the variant `segment_left_inverse_decreasing`
with `[f b , f a]`?

ybertot: This is delaying the merge of the main theorem for the sake of being complete with respect to less important by-products.
Anyway, I am adding the comment and the extra lemma in the next commit.

ybertot:
@CohenCyril you requested changes long ago, and I think they were made, can you check?

theories/normedtype.v
```diff
@@ -3801,6 +3801,50 @@ rewrite !near_simpl near_withinE near_simpl => Pf; near=> y.
...
+Proof.
+rewrite propeqE; split=> /= /nbhs_normP [_/posnumP[e] ye];
+apply/nbhs_normP; exists e%:num => // t; rewrite -ball_normE /= => et.
+  apply: ye; rewrite -ball_normE /= !opprD addrA addrACA subrr add0r.
+  by rewrite opprK addrC.
+have /= := ye (t - (x - y)); rewrite addrNK; apply.
+by rewrite -ball_normE /= opprB addrCA opprD addrA subrr add0r opprB.
+Qed.
+
+Lemma cvg_shift {T : Type} {K : numDomainType} {R : normedModType K}
```
CohenCyril, line 3840:
```suggestion
Lemma cvg_comp_shift {T : Type} {K : numDomainType} {R : normedModType K}
```

theories/normedtype.v
```diff
@@ -3936,6 +3980,558 @@ Qed.
...
+  This does not expose the order between interval bounds. *)
+Lemma bound_itvE1 (a b : R): ((a \in `[a, b]) = (a <= b)) *
+                     ((b \in `[a, b]) = (a <= b)) *
+                     ((a \in `[a, b[) = (a < b)) *
+                     ((b \in `]a, b]) = (a < b)) *
+                     (a \in `[a, +oo[ ) *
+                     (a \in `]-oo, a]).
+Proof. by rewrite !(boundr_in_itv, boundl_in_itv). Qed.
+
+Definition bound_itvE := (bound_itvE1, boundr_in_itv, boundl_in_itv).
```
CohenCyril, line 3998: Why? Doesn't `bound_itvE1` subsume them all?

theories/normedtype.v
```diff
@@ -3936,6 +3980,558 @@ Qed.
...
+
+End numDomainType.
+
+Lemma nearN {R : numFieldType} (a : R) (P : R -> Prop) :
+  (\forall x \near a,  P (- x)) <-> (\forall x \near (- a), P x).
```
CohenCyril, line 4003:
```suggestion
Lemma nearN {R : numFieldType} (a : R) (P : R -> Prop) :
  (\forall x \near a,  P (- x)) <-> (\forall x \near (- a), P x).
```

CohenCyril:
@ybertot there are several naming problems, ordering of lemmas and too strong hypotheses here and there. I take the lock on this PR, please do not touch anything yet.

thery:
@ybertot  @CohenCyril news?

ybertot:
I had a direct discussion with @CohenCyril.  I won't be able to work on this PR before October.

affeldt-aist:
> including backports to mathcomp

ltr_distlC`, `ler_distlC`, `subset_itv_oo_cc` backported to MathComp
(PR: https://github.com/math-comp/math-comp/pull/791)

theories/normedtype.v
```diff
@@ -3936,6 +3990,600 @@ Qed.
...
+Proof.
+move=> aLb fC fK x y xfbfa yfbfa; rewrite -ler_opp2.
+apply: (@segment_can_le (- b) (- a) (f \o -%R) (- g));
+    rewrite /= ?ler_opp2 ?opprK//.
+  move=> z zab; apply: continuous_comp; first exact: continuousN.
+  by apply: fC; rewrite oppr_itvcc.
+by move=> z zab; rewrite -[- g]/(@GRing.opp _ \o g)/= fK ?opprK// oppr_itvcc.
+Qed.
+
+Notation "f @`[ a , b ]" := (`[minr (f a) (f b), maxr (f a) (f b)])
```
affeldt-aist, line 4280: These notations are intended to become global?

CohenCyril: Would be nice.

affeldt-aist:
> I will stop there for now... It needs a second pass to create `realfun.v`, dispatch various lemmas (including backports to mathcomp) where they belong and double check their names. There still are some generalizations to make on some auxiliary lemmas. Also the changelog must be update. @ybertot if you feel like doing it...

I did some work in this direction but I have to stop for now.

PR #386, opened 2021-06-04
divergence of the harmonic series

affeldt-aist:
- incidentally remove a lemma that made its way to MathComp


CohenCyril (approved):
really nice proof!

PR #394, opened 2021-06-17
fixes #365

### LINKED ISSUE #365: adde_def or adde_undef

affeldt-aist:
fixes #365 

since we observed that we rather use `~~ adde_undef blah`


CohenCyril (approved):
LGTM

PR #371, opened 2021-05-03
Lebesgue measure and integral

affeldt-aist:
A formalization of the Lebesgue measure and integral.

Fixes #368 .

PR #395, opened 2021-06-17
extension theorem

affeldt-aist:
- build an outer measure out of a measure over a ring of sets

Extracted from PR #371 

TODO: issue about the epsilon trick lemma, see below

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -2,8 +2,8 @@
 From Coq Require Import ssreflect ssrfun ssrbool.
 From mathcomp Require Import ssrnat eqtype choice seq fintype order bigop.
 From mathcomp Require Import ssralg ssrnum finmap.
-Require Import boolp classical_sets reals ereal posnum topology normedtype.
-Require Import sequences.
+Require Import boolp classical_sets reals ereal posnum nngnum topology.
+Require Import normedtype sequences cardinality csum.
 From HB Require Import structures.
 
 (******************************************************************************)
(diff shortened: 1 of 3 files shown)


theories/measure.v
```diff
@@ -1064,3 +1078,167 @@ by move/(le_outer_measure mu); rewrite muB0 => ->.
 Qed.
 
 End caratheodory_measure.
+
+Lemma epsilon_trick (R : realType) (A : (\bar R)^nat) (e : {nonneg R})
+  (P : pred nat) : (forall n, 0 <= A n) ->
+  \sum_(i <oo | P i) (A i + (e%:nngnum / (2 ^ i)%:R)%:E) <=
+  \sum_(i <oo | P i) A i + (2 * e%:nngnum)%:E.
```
CohenCyril, line 1085: Would
```suggestion
  \sum_(i <oo | P i) (A i + (e%:nngnum / (2 ^ i.+1)%:R)%:E) <=
  \sum_(i <oo | P i) A i + e%:nngnum%:E.
```
be more practical by any chance?

CohenCyril: possibly with  `e : \bar R` such that `0 <= e`?

affeldt-aist: > be more practical by any chance?

It seems to shorten a bit the proofs where it is used (one occurrence in measure.v and another in measure_wip.v not commited yet).

affeldt-aist: > possibly with `e : \bar R` such that `0 <= e`?

Is it ok to issue and wait for PR #403 to check this one?

PR #399, opened 2021-06-22
add the structure of algebra of sets

affeldt-aist:
- rename factory isRingOfSets to isAlgebraOfSets
- generalize `measurableC`


CohenCyril (approved):
LGTM modulo rebase

theories/measure.v
```diff
@@ -161,6 +169,14 @@ Definition T_isRingOfSets : RingOfSets_from_semiRingOfSets T :=
 
 HB.instance T T_isRingOfSets.
 
+Lemma algebraOfSets_measurableT : measurable (@setT T).
```
CohenCyril, line 172: no need to qualify with `algebraOfSets_` it's in a module anyway.
(but you do not need to fix it)

PR #401, opened 2021-06-28
fixes #400

### LINKED ISSUE #400: bigsetU naming
There are several lemmas dealing with `\big[setU/set0]` (e.g., `subset_bigsetU` in `classical_sets.v` or `bigsetU_measurable` in `measure.v`) but `bigsetU` is also the name of the identifier behind the `\bigcup_` notation which does not use `setU` explicitly. What about renaming `bigsetU` to `bigcup` and `bigsetI` to `bigcap`?

(by the way, `bigUB_of` in `measure.v` and `trivIset_bigUI` in `classical_sets.v` should maybe be renamed)

affeldt-aist:
fixes #400 

and rename `B_of` into `seqD` for "sequence of `setD`'s"

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -25,6 +25,8 @@ From HB Require Import structures.
 (*                   that this function maps set0 to 0, is non-negative over  *)
 (*                   measurable sets and is semi-sigma-additive               *)
 (*                                                                            *)
+(*           seqD F == the sequence F_0, F_1 \ F_0, F_2 \ F_1,...             *)
+(*                                                                            *)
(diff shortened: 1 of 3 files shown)


theories/measure.v
```diff
@@ -544,11 +544,11 @@ Qed.
 Section trivIfy.
 Variables (T : Type).
 
-Definition B_of (A : (set T) ^nat) :=
+Definition seqD (A : (set T) ^nat) :=
```
CohenCyril, line 547: If `seqD` is meant to be reused outside (Caratheodory extension theorem?) please document it in the header of the file.
Otherwise, document locally.

PR #417, opened 2021-08-21
add seqDU with a few theorems

affeldt-aist:
- slight generalization of `seqD`

`seqDU` is used in the formalization of Dynkin systems in the Lebesgue measure development

also uses `nondecreasing_seq` instead of unfolded `homo` definitions

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -6,6 +6,9 @@
 
 - in `boolp.v`:
   + lemmas `orA`, `andA`
+- in `measure.v`:
+  + definition `seqDU`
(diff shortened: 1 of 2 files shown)


CohenCyril:
@affeldt-aist is `seqD` still useful in the presence of `seqDU`, doesn't the latter subsume the former?

affeldt-aist:
> @affeldt-aist is `seqD` still useful in the presence of `seqDU`, doesn't the latter subsume the former?

They are the same for non-decreasing sequences wrt inclusion which indeed suggest this simplification
but I could not figure out quickly how to clean seqD proofs, and since both sequences are anyway used
in paper proofs I thought that for the time being it could be enough to put them in the same file, keep an eye
on them and come back to factorization when revising the proofs using seqDU, which should happen soon.

PR #434, opened 2021-09-21
density of rationals in reals

affeldt-aist:
TODO: if it can be merged, create an issue to connect `density_rat` to `dense`


theories/reals.v
```diff
@@ -782,3 +782,87 @@ by move=> ltFxl; exists x=> //; move/lbP : (inf_lower_bound hs) => -> //; exists
 Qed.
 
 End Sup.
+
+Lemma archimedean (R : realType) (x y : R) : 0 < x -> exists n, y < n%:R * x.
```
CohenCyril, line 786: A `realType` is defined as archimedean... https://github.com/math-comp/analysis/blob/61df13069bee88cd5fbd35c22705d8f4488f515d/theories/reals.v#L151

CohenCyril: That's why you have supremums in the first place...

affeldt-aist: This was to have the lemma in this form, which is a form that often appears in books. True this can be shortened with lemmas from archimedean, the next commit does that.

PR #445, opened 2021-10-04
fixes #439

### LINKED ISSUE #439: cvg_seq_bounded
https://github.com/math-comp/analysis/blob/92a7e98255032717faab48882960f0b4c2b30e34/theories/normedtype.v#L2740

shouldn't it be `is_cvg_seq_bounded`?

affeldt-aist:
fixes #439


CohenCyril:
Indeed `is_cvg` is the naming convention for `cvg _` while `cvg` is for `lim _`, but I think we can make an exception for preconditions, since `lim` is not a proposition, there is no ambiguity.

PR #447, opened 2021-10-04
fixes #446

### LINKED ISSUE #446: naming
https://github.com/math-comp/analysis/blob/f8eff34ded1ff4bd51883b56d7563e7cb408c181/theories/normedtype.v#L509

there are remaining uses of `minfty` in the names of lemmas

affeldt-aist:
TODO: issue to question the usefulness of `nbhs_pinfty_gt_pos`?

### DIFF
--- a/theories/normedtype.v
+++ b/theories/normedtype.v
@@ -407,6 +407,7 @@ Notation "-oo" := (ninfty_nbhs _) : ring_scope.
 Section infty_nbhs_instances.
 Context {R : numFieldType}.
 Let R_topologicalType := [topologicalType of R].
+Implicit Types r : R.
 
 Global Instance proper_pinfty_nbhs : ProperFilter (pinfty_nbhs R).
 Proof.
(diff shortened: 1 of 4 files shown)


theories/normedtype.v
```diff
@@ -498,25 +499,22 @@ Proof. by exists c%:num; split => // ; rewrite realE posnum_ge0. Qed.
...
-Proof. by exists c; split => //; apply: ltW. Qed.
+Lemma nbhs_pinfty_ge_real r : r \is Num.real -> \forall x \near +oo, r <= x.
+Proof. by exists r; split => //; apply: ltW. Qed.
 
-Lemma nbhs_minfty_lt (c : {posnum R}) : \forall x \near -oo, c%:num > x.
+Lemma nbhs_ninfty_lt (c : {posnum R}) : \forall x \near -oo, c%:num > x.
 Proof. by exists c%:num; split => // ; rewrite realE posnum_ge0. Qed.
 
-Lemma nbhs_minfty_le (c : {posnum R}) : \forall x \near -oo, c%:num >= x.
+Lemma nbhs_ninfty_le (c : {posnum R}) : \forall x \near -oo, c%:num >= x.
```
CohenCyril, line 511: While you're at it, I'd rephrase those with a precondition rather than an explicit `%:num`

theories/normedtype.v
```diff
@@ -508,35 +509,29 @@ move=> [M [Mreal AM]]; exists (M * 2); split.
...
 
-Lemma nbhs_pinfty_ge (c : {posnum R}) : \forall x \near +oo, c%:num <= x.
-Proof. by exists c%:num; rewrite realE posnum_ge0; split => //; apply: ltW. Qed.
+Lemma nbhs_pinfty_ge_real r : r \is Num.real -> \forall x \near +oo, r <= x.
+Proof. by exists r; split => //; apply: ltW. Qed.
 
-Lemma nbhs_pinfty_gt_real (c : R) : c \is Num.real ->
-  \forall x \near +oo, c < x.
-Proof. by exists c. Qed.
+Lemma nbhs_pinfty_gt r : 0 < r -> \forall x \near +oo, r < x.
```
CohenCyril, line 518:
```suggestion
Lemma nbhs_pinfty_gt_pos r : 0 < r -> \forall x \near +oo, r < x.
```

CohenCyril: Is it necessary though? Isn't it subsumed by the `real` one? which could be the "main" one instead.

affeldt-aist: Both are actually used but it is true that there is such a small difference that this indicates a potential factorization. Should you put an issue about that?

CohenCyril (approved):

PR #458, opened 2021-10-22
doc generated with alectryon

affeldt-aist:
~~- I have only seen part of it, it seems to take forever from topology.v~~

but I'm first-time user, I may very well have gotten it wrong @proux01 @thery

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -124,3 +124,16 @@ doc: __always__ Makefile.coq
 
 doc-clean:
 	rm -rf _build_doc/
+
+alectryon:
+	mkdir -p _build_alectryon/
+	cp -r $(COQFILES) -t _build_alectryon/ --parents
+	cp _CoqProject Makefile* _build_alectryon
+	mkdir -p _build_alectryon/htmldoc
+	+cd _build_alectryon && $(COQMAKE)
+	cd _build_alectryon && alectryon -R classical mathcomp.classical -R theories mathcomp.analysis --frontend coqdoc --backend webpage --output-directory htmldoc \
+		-R theories mathcomp.analysis \
+		$(COQFILES)
+
+alectryon-clean:
+	rm -rf _build_alectryon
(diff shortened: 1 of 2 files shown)


affeldt-aist:
> I can have a look, thanks for the ping. What commands do I need to run to compile that file?

Thanks for your time (and for the tool, the rendering is nice).
`make alectryon`. This is a target in `Makefile.common`.

affeldt-aist:
I tried again today with Alectryon 1.4.0, coq-serapi 8.15.0+0.15.0, coq 8.15.2 and
it seems that it could generate html versions for all files,
see https://math-comp.github.io/analysis/htmldoc_master_alectryon/.

However there were a number of warnings (`Dynlink error... The offending chunk is deliminated by...`) and
a critical error (`[sertop] Critical Dynlink` critical).
I guess that part of the problem is that alectryon does not find `.vo` files despite the `-R` options
that I maybe got wrong.

`--html-minification` does not seem to have a visible effect.

Is there a way to hide the `Proof. ... Qed.` blocks?

cpitclaudel:
Great work!

> Is there a way to hide the Proof. ... Qed. blocks?

Not built-in; it's relatively easy to add it as a custom script, but the exact process depends on what you need.  Do you want them removed completely, or dynamically shown/hidden with a click?

> However there were a number of warnings (Dynlink error... The offending chunk is deliminated by...) and
a critical error ([sertop] Critical Dynlink critical).

How annoying.  I didn't have time to work on Alectryon while I was at AWS, but I'm planning to resume now that I've moved.
Is there a simple recipe I could try out?

> `--html-minification` does not seem to have a visible effect.

Maybe I can try to help with that too.

affeldt-aist:
>

affeldt-aist:
> Do you want them removed completely, or dynamically shown/hidden with a click?

Dynamic would be great but we were only thinking about hiding as with coqdoc.
In fact, so far, we were mostly thinking of using alectryon as a replacement for coqdoc
to generate the documentation of MathComp-Analysis as in:
https://math-comp.github.io/analysis/

> Is there a simple recipe I could try out?

Maybe clone this branch https://github.com/math-comp/analysis/tree/alectryon
and try `make alectryon`. This takes some time.

PR #404, opened 2021-07-12
Lebesgue integral

affeldt-aist:
~~based on PR #422~~ (merged) (~~which is based on PR #403~~ (merged))

~~duplicates lemmas from PR #421 and PR #430~~ (merged)

~~also uses elements of PR #371 as axioms (proved in PR #371) (should be rebased on top of PR #371)~~

based on PR #371

PR #473, opened 2021-11-12
lemmas about sequences of ereals

affeldt-aist:
- from PR #404


affeldt-aist:
Can we consider the proof simplifications as an approval?

PR #487, opened 2021-12-04
change inequality in ereal_{d,}nbhs and {p,n}infty_{d,}nbhs

affeldt-aist:
- from strict to large
- also adds a few lemmas to `ereal.v`

closed #122 @CohenCyril 

This is globally getting a bit bigger (only a few proofs are shortened).
I have maybe not been careful enough when updating.

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -31,6 +31,14 @@
   + lemmas `cvgr_expR`, `cvgn_expR`
 
 ### Changed
+  
+
+- in `ereal.v`:
+  + definitions `ereal_dnbhs` and `ereal_nbhs` changed to use large inequality instead
+    of strict inequality
+- in `normedtype.v`:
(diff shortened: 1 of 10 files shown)


affeldt-aist:
IIRC, the problem with this PR is that it does not bring the expected code size reduction. But there is now code size augmentation (except with one lemma). Since we all seem to agree that large inequalities fell better suited, we can maybe merge this PR for now. What do you think?

CohenCyril:
I think this PR essentially exibits a lack of abstraction. We should actually understand how to make the code smaller, and I think it might not be about changing the def.

affeldt-aist:
NB: after a rebase on MathComp 1.3.1, the proofs of the lemmas `open_ereal_lt_ereal` and `open_ereal_gt_ereal` in `normedtype.v` rely on internals (to fix)

affeldt-aist:
@CohenCyril This "experiment" has not been conclusive. Changing from strict to large inequality does not seem to have any impact on proof size. Shall we just close it instead of postponing it forever?

PR #488, opened 2021-12-06
mem_set expressed with reflect

affeldt-aist:
I think this makes the lemma `mem_set` a bit more useful (it changes the statements of `forall_sig` and `restrict_depE` though).


CohenCyril:
no please, no.

CohenCyril:
The equivalence is already dealt with by `inE`, these functions are meant to be adapter lemmas.
If you really need a `reflect` version add a new lemma called `mem_setP`

PR #489, opened 2021-12-06
express sup with supremum

### LINKED ISSUE #497: Doc of reals.v is missing
The docs of reals is lacking (should explain the reasons for the file to exist and the axiomatic of reals)
https://github.com/math-comp/analysis/blob/91644432eec13f644278a14ab25053aad8266c77/theories/reals.v#L6-L9

affeldt-aist:
The goal of this PR is to replace the definition of `sup` in `reals.v` by a definition using the more generic `supremum` from `classical_sets.v`. As a consequence, for example `sup_out` becomes a consequence of the properites of `supremum` and need not be axiomatized.

fixes #497

### DIFF
--- a/theories/reals.v
+++ b/theories/reals.v
@@ -1,12 +1,39 @@
+(* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
 (* -------------------------------------------------------------------- *)
 (* Copyright (c) - 2015--2016 - IMDEA Software Institute                *)
 (* Copyright (c) - 2015--2018 - Inria                                   *)
 (* Copyright (c) - 2016--2018 - Polytechnique                           *)
+(* -------------------------------------------------------------------- *)
 
-(* -------------------------------------------------------------------- *)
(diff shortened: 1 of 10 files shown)


theories/reals.v
```diff
@@ -67,6 +88,9 @@ split=> [[x ENx]|[x Ex]]; exists (- x); last by rewrite -memNE.
 by rewrite memNE opprK.
 Qed.
 
+Lemma emptyN E : (-%R @` E) = set0 <-> E = set0.
```
CohenCyril, line 91: maybe:
```suggestion
Lemma opp_set_eq0 E : (-%R @` E) = set0 <-> E = set0.
```
?

affeldt-aist:
Could we merge this one on the grounds that it simplifies code a bit and provides documentation?
(It is my understanding that the blocking point is that lemmas in classical_sets.v could be better organized
to highlight the different side-conditions. Since it take time, we can maybe issue about that.)

CohenCyril (approved):
ok, sorry it took so much time

PR #511, opened 2022-01-09
rewriting posnum

CohenCyril:
Generalizing posnum and nonneg and linking the two
TODO:
- [x] add support for inequalities in extended reals (done in https://github.com/math-comp/analysis/pull/375 )
- [ ] discuss construction of orderType on signs (c.f., commented Section Order)
- [ ] add a comment to signed_intro
- [ ] fix two problems in the application of le_num
- [ ] add a nonzero structure to detect that `` `|x| > 0`` when `x != 0`


CohenCyril:
CC @affeldt-aist @proux01

proux01:
I had a quick look, this looks great.
I'd like to do a more careful review, but this will probably have to wait for a few more days.

theories/sequences.v
```diff
@@ -216,6 +216,8 @@ Qed.
 
 End sequences_patched.
 
+Inductive box T := Box of T.
+
```
CohenCyril, line 220: Helper lemma for using the lemma `near`, which is unused here, should be moved to an other
```suggestion
```

theories/landau.v
```diff
@@ -1070,16 +1069,12 @@ Lemma mulo_numClosedFieldType (F : filter_on pT) (h1 h2 f g : pT -> R^o) :
...
- have [/= h] := littleo; apply.
- by rewrite sqrtC_gt0 posnum_gt0.
-by rewrite nnegrE; apply (ltW [gt0 of e%:num]).
-by rewrite nnegrE; apply (ltW [gt0 of e%:num]).
+rewrite [`|_|]normrM -(sqrCK (ltW (gt0 e))) expr2 sqrtCM ?qualifE//.
+rewrite (@normrM _ (h1 x) (h2 x)) mulrACA ler_pmul //; near: x;
+by have [/= h] := littleo; apply.
 Unshelve. all: by end_near. Qed.
 
+
```
proux01, line 1077:
```suggestion

```
spurious new line?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
+(* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
+From Coq Require Import ssreflect ssrfun ssrbool.
+From mathcomp Require Import ssrnat eqtype choice order ssralg ssrnum ssrint.
+Require Import boolp.
+
+Reserved Notation "'{compare' x0 & nz & cond }"
+  (at level 0, x0 at level 200, nz at level 200, format "'{compare'  x0  &  nz  &  cond }").
```
proux01, line 7:
```suggestion
Reserved Notation "{ 'compare' x0 & nz & cond }"
  (at level 0, x0 at level 200, nz at level 200, format "{ 'compare'  x0  &  nz  &  cond }").
```
this is what seems to be done everywhere in mathcomp (but not in analysis), don't know what's best

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Local Coercion is_real r := if r is Real _ then true else false.
+Definition reality_cond (n : reality) (x : T) :=
+  match n with
+  | Real (Sign NonNeg) => x >= x0
+  | Real (Sign NonPos) => x <= x0
+  | Real AnySign       => (x0 <= x) || (x <= x0)
+  | Arbitary           => true
+  end.
+Local Coercion reality_cond : reality >-> Funclass.
```
proux01, line 97: This is used only once in the section (as a coercion), not sure it is worth the effort to declare it a local coercion

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+(* Record ty := Ty { *)
+(*   rty :> T; *)
+(*   nz : nullity; *)
+(*   cond : reality; *)
+(*   Pty : (nz ==> (rty != x0)) && cond rty *)
+(* }. *)
+
+(* Definition ty_def (x : ty) : def (nz x) (cond x) := Def (Pty x). *)
+(* Definition def_ty n c (x : def n c) : ty := Ty (P x). *)
+
```
proux01, line 113: leftover?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Notation "{ ?= x0 }" := (def x0 MaybeZero Arbitrary) : type_scope.
+Notation "x %:sgn" := (from (Phantom _ x)) : ring_scope.
+Notation "[ 'sgn' 'of' x ]" := (fromP (Phantom _ x)) : ring_scope.
+Notation num := r.
+Notation "x %:num" := (r x) : ring_scope.
+Definition posnum (R : numDomainType) of phant R := {> 0%R : R}.
+Notation "{posnum R }" := (@posnum _ (Phant R))  : type_scope.
+Definition nonneg (R : numDomainType) of phant R := {>= 0%R : R}.
+Notation "{nonneg R }" := (@nonneg _ (Phant R))  : type_scope.
+Notation "2" := 2%:R : ring_scope.
```
proux01, line 168: I'm not fond of this notation for 2 (why not 3 or 4 then? and why here?)

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+(* Lemma signed_le_total : totalPOrderMixin [porderType of sT]. *)
+(* Proof. by move=> x y; apply: le_total. Qed. *)
+
+(* Canonical signed_latticeType := LatticeType sT signed_le_total. *)
+(* Canonical signed_distrLatticeType := DistrLatticeType sT signed_le_total. *)
+(* Canonical signed_orderType := OrderType sT signed_le_total. *)
+
+(* End Order. *)
+
```
proux01, line 198: leftover?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Section Theory.
+Context {d : unit} {T : porderType d} {x0 : T}
+  {nz : nullity} {cond : reality}.
+Local Notation sT := {compare x0 & nz & cond}.
+Implicit Type x : sT.
+
+Lemma signed_intro {x} : x%:num = x%:num :> T. Proof. by []. Qed.
```
proux01, line 205: What is this `signed_intro` for? (it seems unused)

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Proof. by move=> ?; rewrite le_gtF//; apply: ge0. Qed.
+
+Lemma le0 x : unify cond NonPos -> x0 >= x%:num :> T.
+Proof. by move=> cond_eq; case: x; rewrite cond_eq/= => ? /andP[]. Qed.
+
+Lemma gt0F x : unify cond NonPos -> x0 < x%:num :> T = false.
+Proof. by move=> ?; rewrite le_gtF//; apply: le0. Qed.
+
+Lemma cmp0 {r : real} x :
+  unify cond (Real r) -> (x0 >=< x%:num).
```
proux01, line 242: Is it common to write `0 >=< x` rather than `x >=< 0`? ssrnum seems to have both and I may find the second one a tad more natural (we have `x != 0` and not `0 != x` below)

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Proof.
+move=> cond_eq;  rewrite /(_ >=< _); move: x; rewrite {}cond_eq.
+by case: r => [[|]|] [/= x /andP[_ ->]]//; rewrite orbT.
+Qed.
+
+Lemma neq0 x : unify nz NonZero -> x%:num != x0 :> T.
+Proof. by move=> nz_eq; case: x; rewrite nz_eq/= => ? /andP[]. Qed.
+
+Lemma eq0 x : unify nz NonZero -> x%:num == x0 :> T = false.
```
proux01, line 251:
```suggestion

Lemma eq0F x : unify nz NonZero -> x%:num == x0 :> T = false.
```

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+ Proof. by move=> /neq0-/(_ x)/negPf->. Qed.
+
+End Theory.
+
+Arguments gt0 {d T x0 nz cond} _ {_ _}.
+Arguments ge0 {d T x0 nz cond} _ {_}.
+Arguments lt0 {d T x0 nz cond} _ {_ _}.
+Arguments le0 {d T x0 nz cond} _ {_}.
+Arguments cmp0 {d T x0 nz cond r} _ {_}.
+Arguments neq0 {d T x0 nz cond} _ {_}.
```
proux01, line 261: Missing `Arguments` for `le0F`, `lt0F`, `gt0F` and `eq0`?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Notation "[ 'gt0' 'of' x ]" := (ltac:(refine (gt0 x%:sgn))).
+Notation "[ 'lt0' 'of' x ]" := (ltac:(refine (lt0 x%:sgn))).
+Notation "[ 'ge0' 'of' x ]" := (ltac:(refine (ge0 x%:sgn))).
+Notation "[ 'le0' 'of' x ]" := (ltac:(refine (le0 x%:sgn))).
+Notation "[ 'cmp0' 'of' x ]" := (ltac:(refine (cmp0 x%:sgn))).
+Notation "[ 'neq0' 'of' x ]" := (ltac:(refine (neq0 x%:sgn))).
```
proux01, line 268: Should those have a `Reserved Notation` on top of the file?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Canonical one_snum := Signed.mk one_snum_subproof.
+
+Definition opp_reality_subdef (xnz : nullity) (xr : reality) :
+    reality :=
+  match xr with
+  | Real (Sign >=0) => <=0
+  | Real (Sign <=0) => >=0
+  | Real AnySign    => >=<0
+  | Arbitrary              => >?<0
+  end.
```
proux01, line 313:
```suggestion
  match xr with
  | Real (Sign >=0) => <=0
  | Real (Sign <=0) => >=0
  | Real AnySign => >=<0
  | Arbitrary => >?<0
  end.
```

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Canonical opp_snum (xnz : nullity) (xr : reality) (x : {num R & xnz & xr}) :=
+  Signed.mk (opp_snum_subproof x).
+
+Definition add_samesign_subdef (xnz ynz : nullity) (xr yr : reality) :=
+  match xr, yr with
+      | Real (Sign >=0), Real (Sign >=0)
+      | Real (Sign <=0), Real (Sign <=0) => true
+      | _, _ => false
+  end.
```
proux01, line 331: Why having `xnz` and `ynz` as arguments here? (and same for many others below, I guess it's for the sake of uniformity?)
Also indentation:
```suggestion
Definition add_samesign_subdef (xnz ynz : nullity) (xr yr : reality) :=
  match xr, yr with
  | Real (Sign >=0), Real (Sign >=0)
  | Real (Sign <=0), Real (Sign <=0) => true
  | _, _ => false
  end.
```

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Definition add_nonzero_subdef (xnz ynz : nullity) (xr yr : reality) :=
+  nz_of_bool (xr && add_samesign_subdef xnz ynz xr yr && (xnz || ynz)).
```
proux01, line 334: Isn't the `xr &&` redundant with `add_samesign_subdef`?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Canonical add_snum (xnz ynz : nullity) (xr yr : reality)
+  (x : {num R & xnz & xr}) (y : {num R & ynz & yr}) :=
+  Signed.mk (add_snum_subproof x y).
+
+Definition mul_nonzero_subdef (xnz ynz : nullity) (xr yr : reality) :=
+  nz_of_bool match xnz, ynz with
+                    | NonZero, NonZero => NonZero
+                    | _      , _       => MaybeZero
+                    end.
```
proux01, line 369: Why have `xr` and `yr` as arguments? (and same for many others below, I guess it's for the sake of uniformity?)
Also
```suggestion
Definition mul_nonzero_subdef (xnz ynz : nullity) (xr yr : reality) :=
  nz_of_bool (xnz && ynz).
```

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Canonical mul_snum (xnz ynz : nullity) (xr yr : reality)
+    (x : {num R & xnz & xr}) (y : {num R & ynz & yr}) :=
+   Signed.mk (mul_snum_subproof x y).
+
+Lemma natmul_snum_subproof (xnz nnz : nullity) (xr nr : reality)
+  (x : {num R & xnz & xr}) (n : {compare 0%N & nnz & nr})
+    (rnz := mul_nonzero_subdef xnz nnz xr nr)
+    (rrl := mul_reality_subdef xnz nnz xr nr) :
+  Signed.spec 0 rnz rrl (x%:num *+ n%:num).
```
proux01, line 406: Couldn't we be more precise, replacing `nr` with `>=0`? in which case `mul_reality_subdef xnz ynz xr yr` just becomes `xr`

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Canonical natmul_snum (xnz nnz : nullity) (xr nr : reality)
+    (x : {num R & xnz & xr}) (n : {compare 0%N & nnz & nr}) :=
+   Signed.mk (natmul_snum_subproof x n).
+
+Lemma intmul_snum_subproof (xnz nnz : nullity) (xr nr : reality)
+  (x : {num R & xnz & xr}) (n : {num int & nnz & nr})
+    (rnz := mul_nonzero_subdef xnz nnz xr nr)
+    (rrl := mul_reality_subdef xnz nnz xr nr) :
+  Signed.spec 0 rnz rrl (x%:num *~ n%:num).
```
proux01, line 425: Couldn't we be more precise, knowing that `nr` is `Real _` (even if we don't know the sign)?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+by rewrite {}/rr; case: xnz xr x => [] [[[]|]|]//= r/=;
+   rewrite ?expf_eq0 ?eq0 ?andbF//=;
+   rewrite -?[_ || _]/(_ \is Num.real) ?realX ?exprn_ge0.
+Qed.
+
+Canonical exprn_snum (nz : nullity) (r : reality) (x : {num R & nz & r}) n :=
+  Signed.mk (exprn_snum_subproof x n).
+
+Definition both_nonzero_subdef (xnz ynz : nullity) (xr yr : reality) :=
+  nz_of_bool (xnz && ynz).
```
proux01, line 470: This is exactly `mul_nonzero_subdef`

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Arguments both_nonzero_subdef /.
+
+Lemma min_snum_subproof (xnz ynz : nullity) (xr yr : reality)
+    (x : {num R & xnz & xr}) (y : {num R & ynz & yr}) 
+    (rnz := both_nonzero_subdef xnz ynz xr yr)
+    (rrl := both_reality_subdef xnz ynz xr yr) :
+  Signed.spec 0 rnz rrl (Num.min x%:num y%:num).
```
proux01, line 477: We could be more precise, for instance min(<0, >=0) != 0 or min(<=0, >=0) <= 0

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Definition max_nonzero_subdef (xnz ynz : nullity) (xr yr : reality) : nullity :=
+  nz_of_bool ((xnz && ynz)
+  || match xr, yr with NonNeg, NonNeg => xnz || ynz | _, _ => false end).
+Arguments max_nonzero_subdef /.
+
+Lemma max_snum_subproof (xnz ynz : nullity) (xr yr : reality)
+    (x : {num R & xnz & xr}) (y : {num R & ynz & yr}) 
+    (rnz := max_nonzero_subdef xnz ynz xr yr)
+    (rrl := both_reality_subdef xnz ynz xr yr) :
+  Signed.spec 0 rnz rrl (Num.max x%:num y%:num).
```
proux01, line 500: This could probably also be more precise

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Section NumClosedStability.
+Context {R : numClosedFieldType}.
+
+Definition sqrtC_reality_subdef (xnz : nullity) (xr : reality) :=
+  if xr is Real (Sign >=0) then xr else Arbitrary.
+Arguments sqrtC_reality_subdef /.
+
+Lemma sqrtC_snum_subproof xnz xr (x : {num R & xnz & xr})
+  (nz := sqrt_nonzero_subdef xnz xr) (r := sqrtC_reality_subdef xnz xr):
+  Signed.spec 0  nz r (sqrtC x%:num).
```
proux01, line 557: Could we be more precise on `nz`?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Proof. by  move: nz r x => [] [[[]|]|] [[|n]]. Qed.
+
+Canonical double_snum nz r x :=
+  Signed.mk (@double_snum_subproof nz r x).
+
+Lemma addn_snum_subproof (xnz ynz : nullity) (xr yr : reality)
+    (x : {compare 0 & xnz & xr}) (y : {compare 0 & ynz & yr})
+    (rnz := add_nonzero_subdef xnz ynz xr yr)
+    (rrl := both_reality_subdef xnz ynz xr yr) :
+  Signed.spec 0 rnz rrl (x%:num + y%:num).
```
proux01, line 593: Could we directly put `>=0` for the reality? (same for all nat ops)

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+by move=> [[|x]//= _] [[|y]//= _]; rewrite muln0.
+Qed.
+
+Canonical muln_snum (xnz ynz : nullity) (xr yr : reality)
+    (x : {compare 0 & xnz & xr}) (y : {compare 0 & ynz & yr}) :=
+   Signed.mk (muln_snum_subproof x y).
+
+Lemma expn_snum_subproof (xnz : nullity) (xr : reality)
+   (x : {compare 0 & xnz & xr}) (rr := exprn_reality_subdef xnz xr) n :
+  Signed.spec 0 xnz rr (x%:num ^ n).
```
proux01, line 620: Bikeshedding: we could be a little more precise on `0 ^ 0 = 1` is `!=0`

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Canonical minn_snum (xnz ynz : nullity) (xr yr : reality)
+    (x : {compare 0 & xnz & xr}) (y : {compare 0 & ynz & yr}) :=
+  Signed.mk (minn_snum_subproof x y).
+
+Lemma maxn_snum_subproof (xnz ynz : nullity) (xr yr : reality)
+    (x : {compare 0 & xnz & xr}) (y : {compare 0 & ynz & yr}) 
+    (rnz := max_nonzero_subdef xnz ynz xr yr)
+    (rrl := both_reality_subdef xnz ynz xr yr) :
+  Signed.spec 0 rnz rrl (maxn x%:num y%:num).
```
proux01, line 648: `rnz` could probably just be `nz_of_bool (xnz || ynz)`

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+End NatStability.
+
+Section Morph0.
+Context {R : numDomainType} {cond : reality}.
+Local Notation nR := {num R & ?=0 & cond}.
+Implicit Types x y : nR.
+Local Notation num := (@num _ _ (0 : R) ?=0 cond).
+
+Lemma num_eq0 x : (x%:num == 0) = (x == (0%:sgn : nR)).
```
proux01, line 667: What is this for? (seems unused)

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+  Num.min x%:num y%:num < a = (x%:num < a) || (y%:num < a).
+Proof. by rewrite -comparable_lt_minl// real_comparable. Qed.
+
+End MorphReal.
+
+Section MorphGe0.
+Context {R : numDomainType} {nz : nullity}.
+Local Notation nR := {num R & ?=0 & >=0}.
+Implicit Type x y : nR.
+Local Notation num := (@num _ _ (0 : R) ?=0 >=0).
```
proux01, line 735: Why is the nullity fixed to `?=0`? (it seems to me everything in this section would also work fine for `!=0`)

proux01 (commented):
Sorry for the delay, here are some comments after a careful reading
of the code. I still need to
* do the doc
* do some testing
* do the extended reals

so more comments will come later.

Main comment: in abstract interpretation, we would say that the abstract domain (>=0, <=0) doesn't have the property of the best abstraction, which causes a bunch of issues (should 0 be considered >=0 or <=0 ? both abstractions are not comparable), the usual fix is to add a 0 in the lattice (as (meet <=0 >=0) so that 0 now has a best abstraction). I could offer a commit with that change if you want.

Comments on entire files
* measure.v and normedtype.v: replace all `%:nngnum` by `%:num`?
* nngnum.v and posnum.v: shouldn't we just remove these files?

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+
+Section NumDomainStability.
+Context {R : numDomainType}.
+
+Lemma zero_snum_subproof cond : Signed.spec 0 MaybeZero cond (0 : R).
+Proof. by case: cond => // -[[]|]//=; rewrite lexx. Qed.
+
+Canonical zero_snum cond := Signed.mk (zero_snum_subproof cond).
+
+Lemma one_snum_subproof : Signed.spec 0 NonZero NonNeg (1 : R).
```
proux01, line 301: I would expect
```suggestion

Lemma one_snum_subproof nz cond : Signed.spec 0 nz cond (1 : R).
```
otherwise
```Coq
Check (1%R : R)%:nng.
```
fails with
```Coq
The term "(1 : R)%:sgn" has type "{compare 0 & !=0 & >=0}"
while it is expected to have type "{nonneg R}"
```

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+        by rewrite mulrz_ge0_le0|by rewrite mulrz_le0|
+        by rewrite -realE rpredMz//; apply: cmp0].
+Qed.
+
+Canonical intmul_snum (xnz nnz : nullity) (xr nr : reality)
+    (x : {num R & xnz & xr}) (n : {num int & nnz & nr}) :=
+   Signed.mk (intmul_snum_subproof x n).
+
+Lemma inv_snum_subproof (xnz : nullity) (xr : reality) (x : {num R & xnz & xr}) :
+  Signed.spec 0 xnz xr (x%:num^-1 : R).
```
proux01, line 441: I'm a bit worried by the fact that the following fails but I have no idea of a good fix:
```Coq
Variable x : {posnum R}.
Check (x%:num^-1)%:pos.  (* OK *)
Check (x%:num^-1)%:nng.  (* fails *)
```

proux01:
Previously, a posnum was also a nonneg thanks to that Canonical instance in posnum.v
```Coq
Canonical posnum_nngnum x := Nonneg.NngNum x%:num (posnum_ge0 x).
```
it seems we have a regression here (c.f. comments above on one and inv).

CohenCyril:
Thank you for your thorough review @proux01, this is much appreciated. I will try to address all your comments.
My main objective on this pass was to be able to cover everything `posnum` and `nnegnum` were covering before, some holes are voluntary in this first past others not.

> Main comment: in abstract interpretation, we would say that the abstract domain (>=0, <=0) doesn't have the property of the best abstraction, which causes a bunch of issues (should 0 be considered >=0 or <=0 ? both abstractions are not comparable), the usual fix is to add a 0 in the lattice (as (meet <=0 >=0) so that 0 now has a best abstraction). I could offer a commit with that change if you want.

Yes, I noticed this very late and this additional would make sense.
However in practical cases, it is very easy to get around this tricky corner case, e.g. by using `addr0` or `mulr0` to remove `0` altogether from the guilty goal, and I guess that's why it was not a problem at all.
Naturally, I'm not opposed to adding this case of course!

> Comments on entire files
> 
> * measure.v and normedtype.v: replace all `%:nngnum` by `%:num`?
> * nngnum.v and posnum.v: shouldn't we just remove these files?

Yes, I kept these two files as a stub in order to transition smoothly between the new and old code, especially to anticipate a compat with unmerged PRs.

theories/landau.v
CohenCyril: most certainly

theories/signed.v
CohenCyril: Yes you are right, gluing `{` to the rest is dangerous in terms of lexing.

theories/signed.v
CohenCyril: Yes, that's because there are at least two ways to get signs totally ordered :-/ and this is not the one which is used most of the time. We have to discuss to make them live with each other, because right now, I don't know how...

theories/signed.v
CohenCyril: It is made for debugging purposes mostly, it forces canonical structure inference by rewriting with it.

theories/signed.v
CohenCyril: I agree with you, except that [`Num.real` is convertible to `0 >=< x`](https://github.com/math-comp/math-comp/blob/80ccf365b8a5f35e0ce329b37b68aa0339c17c77/mathcomp/algebra/ssrnum.v#L387) rather than the other, which makes a more things work better with `0 >=< x` than with `x >=< 0`...

theories/signed.v
proux01: Ok, might be worth a comment?

theories/signed.v
CohenCyril: Maybe, I think since they are unlikely to be combined like the others, it is not important, but at least it would be uniform.

theories/signed.v
CohenCyril: Uniformity indeed, just in case there is a more precise interpretation taking xnz and ynz into account.

theories/signed.v
CohenCyril: Mmhh, looks like it indeed.

theories/signed.v
CohenCyril: We also can expect natural numbers to always be `>=0`, however we could replace it in `rnz` and `rrl` to simplify computations, or just in case it fails for some weird reason.

theories/signed.v
CohenCyril: Same, no in `n`, but yes in `rnz` and `rrl`.

theories/signed.v
CohenCyril: Mmh, true, maybe it we could eliminate the redundancy...

theories/signed.v
CohenCyril: Oh yeah, this is an unfortunate copy-paste of the real sqrt I think.

theories/signed.v
CohenCyril: Cf my answer for `natmul`.

theories/signed.v
CohenCyril: probably...

theories/signed.v
CohenCyril: For the sake of completeness? (For the sake of which we should probably add a couple more things)

theories/signed.v
CohenCyril: In order to for the inference to work properly, the arguments have to variable as nullity and reality and the output must depend solely on this variable. Letting `nz` and `cond` be free means that during inference they stay variable and are not computed by the various abstract interpretation functions.

I think the way to preserve the behavior you mention is to have a coercion operator (using the same technique as all the other operators `add`, `mul `, etc).

theories/signed.v
CohenCyril: Cf my previous comment about coercion operator.

theories/signed.v
proux01: I don't see how to do it with coercions, but I have an idea encoding as a boolean function the order `more_precise` of the abstract lattice (for instance `more precise (!0, >=0) (?0, >=0) = true`), thus enabling to have a cast lemma (for instance from posnum to nonneg) taking a proof of `more_precise = true` as input, that could then be fed with `erefl` in a notation. Let me try that.

theories/signed.v
```diff
@@ -0,0 +1,798 @@
...
+Class infer (P : Prop) := Infer : P.
+Hint Mode infer ! : typeclass_instances.
+Hint Extern 0 (infer _) => (exact) : typeclass_instances.
+Lemma inferP (P : Prop) : P -> infer P. Proof. by []. Qed.
+
+Class unify {T} (x y : T) := Unify : x = y.
+Hint Mode unify - + - : typeclass_instances.
+Class unify' {T} (x y : T) := Unify' : x = y.
+Instance unify'P {T} (x y : T) : unify' x y -> unify x y := id.
+Hint Extern 0 (unify' _ _) => vm_compute; reflexivity : typeclass_instances.
```
proux01, line 70: BTW, this seems unused

CohenCyril: Oh no this is totally used in `eq0` etc

proux01: Hmm, looks like there is some magic I still don't perfectly understand, I'll have a closer look.

theories/signed.v
proux01: In fact, here is a much simpler fix: https://github.com/proux01/analysis/commit/54f21a9a3cb5132aec9b50d234ecf10e81ff5baa (feel free to cherry-pick)

CohenCyril: Does not seem generic enough, I'll try the "coercion operator"

proux01: @CohenCyril this is more generic: https://github.com/proux01/analysis/commit/422bb1ba398e1730d238f3809ec58ecbb7fe0815 and https://github.com/proux01/analysis/commit/4162ac3865f896961316c5dea388ae62d9737507 is that what you had in mind?

proux01: That was broken (mostly because of the canonical instance for 0 that had a variable as input ;-) ), here is a fix: https://github.com/proux01/analysis/commit/b9a7732916bc4c82c7dd8e0571c81cb2d2a56485

proux01:
> Yes, I noticed this very late and this additional would make sense. However in practical cases, it is very easy to get around this tricky corner case, e.g. by using `addr0` or `mulr0` to remove `0` altogether from the guilty goal, and I guess that's why it was not a problem at all.

I'm not worried about `0 + _` or `0 * _` but I'm less confident about things like `min 0 _` or `max 0 _`.

> Naturally, I'm not opposed to adding this case of course!

Here it is: https://github.com/proux01/analysis/commits/posnum_wip (this branch also adresses all my comments above we agreed upon)

theories/signed.v
CohenCyril: Really nice! I'm a bit worried about [this](https://github.com/proux01/analysis/commit/b9a7732916bc4c82c7dd8e0571c81cb2d2a56485#diff-ec3855f35a085739515162cb2620b5daea6488b737c11df25d618370adc5e240R1942-R1943) and [that](https://github.com/proux01/analysis/commit/b9a7732916bc4c82c7dd8e0571c81cb2d2a56485#diff-ec3855f35a085739515162cb2620b5daea6488b737c11df25d618370adc5e240R1950-R1951) though. Will it go back to oneliners once merged with [the new `=0`](https://github.com/proux01/analysis/commit/8b03eb6ba821d6a3e27ada88c033fd544158d99c)?

proux01: Sorry for the delay, I managed to improve things quite a bit, but that's still not perfect: https://github.com/proux01/analysis/commit/aeb0cfedb71289d0c2869085194eeef01a9fc2a3
Note the following rewrite
```Coq
rewrite [X in _ <= X](_ : 0 = 0%:nng%:num)//
```
that I would have expected to write
```Coq
rewrite [0]/(0%:nng%:num)
```
but that fails with an anomaly, I opened an issue https://github.com/coq/coq/issues/15598
(more generally, that might explain the difficulties here).

Maybe we could have elimination lemmas here like `nngE` and `posE`?

proux01: That works, we are back to one liners: https://github.com/proux01/analysis/commit/64090854e9c8dd3d1c5b23e437696d90ffab2dea

CohenCyril: All of this looks good to me. I expected that `rewrite -[0]/(0%:nng%:num)` would stop working, and I'm ok with elimination lemmas !

CohenCyril: Feel free to stack up on my branch

proux01: Looks like I don't have write access:
```
remote: Permission to math-comp/analysis.git denied to proux01.
fatal: unable to access 'https://github.com/math-comp/analysis/': The requested URL returned error: 403
```

CohenCyril: This should be fixed

proux01: Thanks!

proux01:
I added the doc I promised, so AFAICT, the remaining TODOs are:
* should notations such as `%:pos` and `%:nng` be in ring_scope? (that would allow for similar notations in `ereal_scope` in ereal.v)
* discuss construction of orderType on signs (c.f., commented Section Order)
* add a comment to signed_intro
* fix two problems in the application of le_num (don't know which ones)
* add a nonzero structure to detect that `|x| > 0 when x != 0 (not sure to grasp the details)

Extended reals are handled separately in #375

proux01:
@CohenCyril I generalized (c.f. last commit) what was already done for nat to infer some sign information based on the type when no canonical instance matches the head symbol. In particular, I added a "top" catch-all that states that everything is in `{compare x0 & ?=0 & >?<0 }`. This enables things like `abse x` (in #375 ) or `max 0 x` to infer `>=0` even when nothing is known about `x`.

The main drawback is that canonical structure inference no longer unfolds anything (since the catch-all triggers before).
Thus a few more unfolding are needed. This seems an acceptable trade-off to me, but you may not share my opinion (see last commit for example of impact).

Apart from the few points above (and adding still more canonical instances, but that's endless), I consider this and #375 ready.

CohenCyril:
> The main drawback is that canonical structure inference no longer unfolds anything (since the catch-all triggers before).
> Thus a few more unfolding are needed. This seems an acceptable trade-off to me, but you may not share my opinion (see last commit for example of impact).

Yes indeed, I think that's a good trade-off.

theories/signed.v
```diff
@@ -0,0 +1,1165 @@
...
+  Signed.spec x0 rnz rrl (Order.min x%:num y%:num).
+Proof.
+rewrite {}/rnz {}/rrl; apply/andP; split.
+  move: xr yr xnz ynz x y => [[[]|]|] [[[]|]|] [] []//= x y; rewrite /Order.min;
+    do ?[by case: (bottom x)|by case: (bottom y)
+        |by case: ifP; rewrite ?eq0F// => xlty;
+            have := !! lt_trans xlty (lt0 y); rewrite lt_neqAle => /andP[]
+        |by rewrite ifT ?eq0F//; apply: lt_le_trans (ge0 y); exact: lt0
+        |by have := !! le0 y;
+            rewrite le_eqVlt => /orP[/eqP ->|]; rewrite ?lt0 ?eq0F//;
```
affeldt-aist, line 581: `/orP[/eqP ->|]` -> `/predU1P[->|]` (4 occurrences)

theories/signed.v
```diff
@@ -0,0 +1,1165 @@
...
+
+(* Enrico's trick for tc resolution in have *)
+Notation "!! x" := (ltac:(refine x)) (only parsing).
+(* infer class to help typeclass inference on the fly *)
+Class infer (P : Prop) := Infer : P.
+Hint Mode infer ! : typeclass_instances.
+Hint Extern 0 (infer _) => (exact) : typeclass_instances.
+Lemma inferP (P : Prop) : P -> infer P. Proof. by []. Qed.
+
+Lemma splitr (R : numFieldType) (x : R) : x = x / 2%:R + x / 2%:R.
```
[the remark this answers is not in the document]
proux01: Looks like it's already there :laughing: (will remove here when rebasing)

proux01:
Since there was no opposition, I rebased (thanks @affeldt-aist for the initial rebase!).
A few points remain (c.f. initial PR description on top), but none of them is blocking.
I would argue in favor of merging (and adding issues to remember those points). @CohenCyril WDYT?

affeldt-aist (commented):
I committed a few simplifications that I observed when doing the rebase for PR #559 .

CohenCyril:
"Pull request authors cannot approve their own pull request", still I'm in favor.

proux01:
Looks everyone agrees, let's merge then.

PR #513, opened 2022-01-11
fixes #512

### LINKED ISSUE #512: should be easier to use with exists2 instead of exists
https://github.com/math-comp/analysis/blob/306b48277c78a1872d83971c74f76dd587d83cdb/theories/ereal.v#L2374

https://github.com/math-comp/analysis/blob/306b48277c78a1872d83971c74f76dd587d83cdb/theories/ereal.v#L2381

affeldt-aist:
fixes #512


proux01 (approved):
Indeed, thanks for fixing it.

PR #543, opened 2022-02-17
Refactoring lebesgue measure and integral

CohenCyril:
Lebesgue measure and integral (was: "refactor of intervals")
CC @affeldt-aist 

closes #371 
closes #492


affeldt-aist:
The errors are syntax errors with Coq 8.13 only.
I fixed some in `lebesgue_measure.v` (by adding parentheses to disambiguate the parsing of intervals) but didn't have time to fix the ones in `lebesgue_integral.v":
```
# File "./theories/lebesgue_integral.v", line 1428, characters 53-54:
# Error: Syntax error: 'nnsfun' 'of' expected after '[' (in [term]).
 ```

affeldt-aist:
The only errors are now "Nix CI for bundle" errors, so they are maybe not relevant.

CohenCyril:
> The only errors are now "Nix CI for bundle" errors, so they are maybe not relevant.

@affeldt-aist I fixed something in `boolp.v` doc, added some (working) nix CI, and we should be good to go.

PR #558, opened 2022-02-28
new file to factor out lebesgue_integral.v

affeldt-aist:
summary:
- `function_space` section and indicator functions moved to `functions.v`
- fimfun moved to `cardinality.v` (with the zmod and ring structure)
- `^\+`, `^\-`, nnfun, and mfun moved to a new file `numfun.v`
- this leaves only the theories of sfun and nnsfun in `lebesgue_integral.v`


CohenCyril:
Great! But why are the `fct_zmodType` etc stuff in `functions` rather than `numfun`?

affeldt-aist:
> Great! But why are the `fct_zmodType` etc stuff in `functions` rather than `numfun`?

It seemed to me that some files (e.g., topology.v, derive.v) didn't need most of `numfun.v`
and that doing it this way is more conservative (it's easier to move things out of `functions.v`
than the other way around).

affeldt-aist:
Heading for merge?

CohenCyril:
I did not have time to so a full review yet... but I saw measurability results in `numfun`, I don't think it should depend on `measure`

theories/functions.v
```diff
@@ -2392,3 +2401,103 @@ Notation "''injpPfun_' dflt" := (injpPfun_ dflt) : form_scope.
...
+
+End function_space_lemmas.
+
+Section fun_cmul.
+Local Open Scope ring_scope.
+
+Definition fun_cmul {U : Type} {R : ringType} (k : R) (f : U -> R) x := k * f x.
+
+End fun_cmul.
+Notation "k *\ f" := (fun_cmul k f) (at level 40, format "k  *\  f") : ring_scope.
```
CohenCyril, line 2503: Actually,  (my mistake), it exists and it is called `\o*`

PR #565, opened 2022-03-03
measurable_fun_mulrn

hoheinzollern:

This adds the lemma `measurable_fun_mulrn`


CohenCyril:
Hi, first, thank you for your contribution, it is greatly appreciated.
In order for your change to be integrated it will need to be slightly changed.
The proof should be compressed and the name should be updated to `measurable_fun_exprn`.
The proof can be done by composition with between a function and a continuous function (the lemma about the measurability of $$\mathbb{R}$$ function is not yet in the library AFAIK)
`measurable_exprn : measurable setT (fun x => x ^+ n)`

CohenCyril:
additionally, I strongly advise you *not to* PR from your fork's master branch.

affeldt-aist:
> (the lemma about the measurability of $$\mathbb{R}$$ function is not yet in the library AFAIK)

Here is a proposal PR #571

PR #590, opened 2022-03-09
fix hb dependency version

affeldt-aist:

fix hb dependency version (only 1.2.0 and 1.2.1 seem ok)
problem revealed by https://github.com/coq/opam-coq-archive/pull/2127

##### Things done/to do

~- [ ] added corresponding entries in `CHANGELOG_UNRELEASED.md`~

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,7 +5,7 @@
 - [The Coq Proof Assistant version ≥ 8.13](https://coq.inria.fr)
 - [Mathematical Components version ≥ 1.13.0](https://github.com/math-comp/math-comp)
 - [Finmap library version ≥ 1.5.1](https://github.com/math-comp/finmap)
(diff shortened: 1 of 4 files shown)


affeldt-aist:
I guess this is ok to merge but I suspect that if we do so further PR will be red.

default.nix
```diff
@@ -1,12 +1,36 @@
...
+
+with coqPackages;
+
+pkgs.stdenv.mkDerivation {
+
+  name = "analysis";
+
+  propagatedBuildInputs = [
+    coq
+    true
```
CohenCyril, line 23: What is that?

affeldt-aist: I don't know. Is it possible that templates/generate.sh messed up with the Nix configuration?

affeldt-aist: It indeed looks like templates/generate.sh overwrote the Nix configuration.

CohenCyril: meta.yml should disable the nix config

PR #611, opened 2022-03-24
Measurable fun exprn

hoheinzollern:
@affeldt-aist and I have reworked the lemma `measurable_fun_mulrn`, now renamed to `measurable_fun_exprn`, following the suggestion by @CohenCyril to use the fact that continuous implies measurable.
This leaves us with an additional assumption that the function `f` from the lemma is continuous, which was not required before.

### DIFF
--- a/theories/lebesgue_measure.v
+++ b/theories/lebesgue_measure.v
@@ -1392,6 +1392,21 @@ rewrite -(addrA (f x * g x *+ 2)) -opprB opprK (addrC (g x ^+ 2)) addrK.
 by rewrite -(mulr_natr (f x * g x)) -(mulrC 2) mulrA mulVr ?mul1r// unitfE.
 Qed.
 
+Lemma measurable_fun_mulrn n D (f : T -> R) :
+  measurable_fun D f -> measurable_fun D (fun x => f x ^+ n).
+Proof.
+elim: n.
+move => _.
+apply: (@eq_measurable_fun _ _ _ (cst (1 : R))) => //.
+exact: measurable_fun_cst.
+move => n ih mf.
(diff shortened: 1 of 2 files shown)


CohenCyril:
But ... why? It's the `fun x => x ^ n` which is continuous, `f` does not have to be. Use composition.

hoheinzollern:
Ok thanks for the suggestion, this actually fixes that.

theories/lebesgue_measure.v
```diff
@@ -1376,6 +1347,20 @@ Proof.
...
 Qed.
 
+Lemma measurable_fun_exprn D n f :
+  measurable_fun D f -> measurable_fun D (fun x => f x ^+ n).
+Proof.
+move => mf.
+rewrite (_ : (fun x => f x ^+ n) = (fun x => x ^+ n) \o f).
+apply: measurable_fun_comp => //.
+apply: continuous_measurable_fun; apply: exp_continuous.
+by apply /funext => y.
```
CohenCyril, line 1357:
```suggestion
apply: measurable_fun_comp ((@GRing.exp R)^~ n) _ _ _.
by apply: continuous_measurable_fun; apply: exp_continuous.
```

theories/lebesgue_measure.v
```diff
@@ -1376,6 +1347,20 @@ Proof.
...
+Qed.
+
+Lemma measurable_fun_sqr D f :
+  measurable_fun D f -> measurable_fun D (fun x => f x ^+ 2).
+Proof. by apply: measurable_fun_exprn. Qed.
```
CohenCyril, line 1362:
```suggestion
Proof. exact: measurable_fun_exprn. Qed.
```

PR #620, opened 2022-03-29
Measure series

affeldt-aist:
##### Motivation for this change

Sum of measures. Based on PR #614 (merged).
There are a few more lemmas about sum of measures and integration but they need lemmas in PR #616 (merged).

fixes #659 @proux01

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [x] added corresponding documentation in the headers

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and put a milestone if possible.

PR #628, opened 2022-04-04
Integral count

### LINKED ISSUE #630: generalize to any choiceType
https://github.com/math-comp/analysis/blob/7b88928f524b16b2bbd29940bfd0ce4e69cb61d0/theories/mathcomp_extra.v#L162

affeldt-aist:

equivalence between series and integral over the counting measure

based on PR #598 

~~assumes PR #614, PR #616, PR #620~~ merged

fixes #630 

@t6s 

##### Things done/to do

### DIFF
--- a/theories/esum.v
+++ b/theories/esum.v
@@ -1,7 +1,7 @@
 (* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
 From mathcomp Require Import all_ssreflect ssralg ssrnum finmap.
 Require Import boolp reals mathcomp_extra ereal classical_sets signed topology.
-Require Import sequences functions cardinality.
+Require Import sequences functions cardinality normedtype numfun.
 
 (******************************************************************************)
 (*                      Summation over classical sets                         *)
@@ -15,6 +15,7 @@ Require Import sequences functions cardinality.
 (*                       function whose codomain is included in the extended  *)
 (*                       reals; it is 0 if I = set0 and sup(\sum_A a) where A *)
 (*                       is a finite set included in I o.w.                   *)
+(*       summable D f := \esum_(x in D) `| f x | < +oo                        *)
 (*                                                                            *)
 (******************************************************************************)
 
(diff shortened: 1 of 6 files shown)


theories/esum.v
```diff
@@ -1,7 +1,7 @@
 (* mathcomp analysis (c) 2017 Inria and AIST. License: CeCILL-C.              *)
 From mathcomp Require Import all_ssreflect ssralg ssrnum finmap.
 Require Import boolp reals mathcomp_extra ereal classical_sets signed topology.
-Require Import sequences functions cardinality.
+Require Import sequences functions cardinality normedtype numfun.
 
 (******************************************************************************)
 (*                      Summation over classical sets                         *)
```
t6s, line 7: The e in the notation esum does not express very well that it is actually an unordered summation(, or the summation over classical sets). But I do not strongly insist on changing that.

theories/lebesgue_integral.v
```diff
@@ -2558,9 +2750,91 @@ move=> [mf]; apply: le_lt_trans; apply ge0_le_integral => //.
...
+Section sequence_measure.
+Local Open Scope ereal_scope.
+Variables (T : measurableType) (R : realType) (m_ : {measure set T -> \bar R}^nat).
+Let m := measure_series m_ O.
+
+Lemma integral_measure_series (D : set T) (mD : measurable D) (f : T -> \bar R) :
+  (forall n, integrable (m_ n) D f) ->
+  measurable_fun D f ->
+  \sum_(n <oo) `|\int[m_ n]_(x in D) f^\- x | < +oo%E ->
+  \sum_(n <oo) `|\int[m_ n]_(x in D) f^\+ x | < +oo%E ->
```
t6s, line 2789: If I remember correctly these two lines are the new findings while you were trying to improve this lemma. Isn't it worth leaving a few comments on these lines?

t6s (approved):
Looks good to me

theories/esum.v
affeldt-aist: It used to be `csum`. Do you think that `usum` would be more appropriate?

theories/lebesgue_integral.v
affeldt-aist: Were they "new findings"? I do not remember very well. Were they oversights in a reference we used? Maybe but now they rather look to me as technical consequences of the `esumB` lemma which was maybe our "finding" in the sense that we did not understand immediately how to organize ourselves to produce a satisfactory version of it. We started with the PR https://github.com/math-comp/analysis/pull/596 that redefined `esum` as `\nnesum_(i in I) f^\+ i - \nnesum_(i in I) f^\- i` so that we could prove:
```
Lemma esumB D f g : esummable D f -> esummable D g ->
  (forall i, D i -> 0 <= f i) -> (forall i, D i -> 0 <= g i) ->
  \esum_(i in D) (f i - g i) = \esum_(i in D) f i - \esum_(i in D) g i.
```
but that led to reproduce lemmas looking too much like already-proved integration lemmas.
Whereas in this PR we can get along without redefining `esum` and prove instead:
```
Lemma esumB D f g : summable D f -> summable D g ->
  (forall i, D i -> 0 <= f i) -> (forall i, D i -> 0 <= g i) ->
  \esum_(i in D) (f \- g)^\+ i - \esum_(i in D) (f \- g)^\- i =
  \esum_(i in D) f i - \esum_(i in D) g i.
```
to limit duplication of similar-looking facts. I am not sure what should be commented on.

affeldt-aist:
> Looks good to me

Is it an approval? Note that the present PR is based on another one which is not directly approved.

affeldt-aist:
We have a need for `ge0_integral_measure_series` in another project.

PR #633, opened 2022-04-10
frechet filter

affeldt-aist:

Just the easy observation that cardinality.v makes it possible to define Frechet filters.

##### Things done/to do

### DIFF
--- a/theories/topology.v
+++ b/theories/topology.v
@@ -3,6 +3,7 @@ From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat eqtype choice div.
 From mathcomp Require Import seq fintype bigop order interval ssralg ssrnum rat.
 From mathcomp Require Import matrix finmap.
 Require Import mathcomp_extra boolp reals classical_sets signed functions.
+Require Import cardinality.
 
 (******************************************************************************)
 (*                  Filters and basic topological notions                     *)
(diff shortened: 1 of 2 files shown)


theories/topology.v
```diff
@@ -1341,18 +1342,33 @@ Definition globally {T : Type} (A : set T) : set (set T) :=
...
 Global Instance globally_properfilter {T : Type} (A : set T) a :
-   infer (A a) -> ProperFilter (globally A).
+  infer (A a) -> ProperFilter (globally A).
 Proof. by move=> Aa; apply: Build_ProperFilter' => /(_ a). Qed.
 
 (** ** Specific filters *)
 
+Section frechet_filter.
+
+Definition frechet_filter := [set S : set nat | finite_set (~` S)].
```
zstone1, line 1359: Can we generalize this to any to use any type `T` instead of `nat`? There's a handy lemma that says 
> ultrafilters either extend the frechet filter, or are principle

which I'd like to prove at some point.

Then the frechet_properfilter would ideally say `infinite_set [set: T] -> ProperFilter frechet_filter`. It looks like that would be possible, but I'm not sure. I could imagine still creating a `Global Instance` specifically for the `nat` case, as that's helpful separately.

affeldt-aist: See https://github.com/math-comp/analysis/pull/633/commits/e885fee421768b6c08e691a69f25a36cdd2f7cd3 for a tentative generalization.

zstone1: Looks good.

PR #646, opened 2022-04-18
use HB to define measures

### LINKED ISSUE #642: measures might be better defined using HB
https://github.com/math-comp/analysis/blob/3b719520482481b6d45b114247fd2cb131312fc6/theories/measure.v#L1098-L1132

### LINKED ISSUE #660: `dirac` instead of `dirac_measure`
https://github.com/math-comp/analysis/blob/77b5614fe8a149e8ae32d797cbe7a8bba54cd1af/theories/measure.v#L1419

This should better be `dirac` instead of `dirac_measure`. Should be fixed once PR #646 is rebased.

affeldt-aist:

fixes #642 

fixes #660 

there is some overlap with PR #598 that can be taken care of later

~~the way `Restr`, `Product_measure1`, `Product_measure2` is certainly not the best one~~
definition of `mrestr`, `mproduct_measure1`, `mproduct_measure2` still wip (fixed)
##### Things done/to do


theories/lebesgue_integral.v
```diff
@@ -3811,37 +3825,49 @@ Variables (m1 : {measure set T1 -> \bar R}) (m2 : {measure set T2 -> \bar R}).
...
-Definition product_measure2 : {measure set (T1 * T2) -> \bar R} :=
-  Measure.Pack _ (Measure.Axioms m0 m_ge0 m2_sigma_additive).
+HB.instance Definition product_measure2_build := isMeasure.Build
+  _ _ _ pm20 pm2_ge0 pm2_sigma_additive.
+
+Definition Product_measure2 :
+  {measure set [the semiRingOfSetsType of (T1 * T2)%type] -> \bar R}.
+apply: Measure.Pack.
+apply: Measure.Class.
+apply: isMeasure0.Axioms_.
```
CohenCyril, line 3866: This exposes the supposedly blackbox implementation of HB defined structures.
Can you explain the purpose of the definition so that we try to rephrase it?

theories/measure.v
```diff
@@ -1095,57 +1107,32 @@ apply/idP/idP => [|/imfsetP[r /=]].
...
-Variables (R : numFieldType) (T : semiRingOfSetsType).
-Record axioms (mu : set T -> \bar R) := Axioms {
-  _ : mu set0 = 0 ;
-  _ : forall x, 0 <= mu x ;
-  _ : semi_additive mu }.
+HB.mixin Record isAdditiveMeasure
+    (R : numFieldType) (T : semiRingOfSetsType) (mu : set T -> \bar R) := {
+  measure0 : mu set0 = 0 ;
+  measure_ge0 : forall x, 0 <= mu x ;
+  measure_semi_additive : semi_additive mu }.
```
proux01, line 1114: Isn't `measure0` implied by `measure_semi_additive` (or alternatively, couldn't `semi_additive` be replaced by `semi_additive2`?)

theories/lebesgue_integral.v
affeldt-aist: This definition is used a few times in `lebesgue_integral.v` (e.g., https://github.com/math-comp/analysis/blob/fixes_642/theories/lebesgue_integral.v#L3535). It seemed necessary to have a constructor to which one can tell explicitly that the domain is measurable because this fact didn't seem inferrable automatically with a standard use of HB packed classes.

theories/measure.v
affeldt-aist: > Isn't measure0 implied by measure_semi_additive

Indeed.

affeldt-aist: > couldn't semi_additive be replaced by semi_additive2?

Under the hypothesis that `mu set0 = 0`, using lemmas `additive2P` and `semi_additive2E`.

CohenCyril: If I recall correcly, `semi_additive2` and `semi_additive` are not equivalent on a semiring.

theories/lebesgue_integral.v
CohenCyril: OK I see! The solution is to create a wrapper around a function, which takes as a phantom argument the measurablility of the domain : `Definition mrestr D f (mD : measurable D):= f` and provide an HB instance for this one.

affeldt-aist: This indeed provides a way not to relay on the internals of HB but I suspect this commit https://github.com/math-comp/analysis/pull/646/commits/8e0001fbeac1edd2925c85a188075112f6ae06d6 is not exactly it.

theories/measure.v
proux01: Indeed, my bad. So this is just `measure0` being redundant.

theories/measure.v
```diff
@@ -1250,61 +1242,48 @@ End additive_measure_on_ring_of_sets.
...
+    format "{ 'measure'  'set'  T  '->'  '\bar'  R }") : ring_scope.
 
-Structure map (phUV : phant (set T -> \bar R)) :=
-  Pack {apply : set T -> \bar R ; _ : axioms apply}.
-Local Coercion apply : map >-> Funclass.
+HB.factory Record isMeasure
+    (R : realFieldType) (T : semiRingOfSetsType) (mu : set T -> \bar R) := {
+  measure0 : mu set0 = 0 ;
+  measure_ge0 : forall x, 0 <= mu x ;
+  measure_semi_sigma_additive : semi_sigma_additive mu }.
```
proux01, line 1263: Isn't `measure0` redundant here too?

theories/measure.v
```diff
@@ -1097,57 +1109,36 @@ apply/idP/idP => [|/imfsetP[r /=]].
...
-  _ : forall x, 0 <= mu x ;
-  _ : semi_additive mu }.
-
-Structure map (phUV : phant (set T -> \bar R)) :=
-  Pack {apply : set T -> \bar R ; _ : axioms apply}.
-Local Coercion apply : map >-> Funclass.
+HB.mixin Record isAdditiveMeasure
+    (R : numFieldType) (T : semiRingOfSetsType) (mu : set T -> \bar R) := {
+  measure_ge0 : forall x, 0 <= mu x ;
+  measure_semi_additive : semi_additive mu }.
```
proux01, line 1115: An alternative to `measure_ge0` would be to ask for `mu : set T -> {nonneg \bar R}` don't know what's best?

CohenCyril: I'm 99% sure it is better to have `measure_ge0`.

theories/lebesgue_integral.v
CohenCyril: > This indeed provides a way not to relay on the internals of HB but I suspect this commit [8e0001f](https://github.com/math-comp/analysis/commit/8e0001fbeac1edd2925c85a188075112f6ae06d6) is not exactly it.

Sure I guess the definition of `mrestr` should be `restr` and not the identity.

theories/measure.v
proux01: Ok, would you mind explaining? it's not obvious to me.

CohenCyril: It's better to deliver the raw type, without having to add an extra `%:num` almost everywhere and have a canonical instance to declare the output is always non-negative.
The idea is pervasive : if possible give the simplest possible type (no structure) and recover any structure by resolution.

CohenCyril: Morever, if there is more structure on the output, since the head symbol is now `%:num` instead of `Measure.sort` -> the head symbol is wrong.

theories/measure.v
affeldt-aist: Maybe not because otherwise the function that associates +oo to any set could be a measure.

theories/measure.v
proux01: So we could add the following
```Coq
Section additive_measure_signed.
Context (R : numFieldType) (T : semiRingOfSetsType).

Variable mu : {additive_measure set T -> \bar R}.

Lemma additive_measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
Proof. exact: measure_ge0. Qed.

Canonical additive_measure_snum S :=
  Signed.mk (additive_measure_snum_subproof S).

End additive_measure_signed.

Section	measure_signed.
Variables (R : numFieldType) (T : semiRingOfSetsType).
         
Variable mu : {measure set T -> \bar R}.

Lemma measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
Proof. exact: measure_ge0. Qed.

Canonical measure_snum S := Signed.mk (measure_snum_subproof S).

End measure_signed.
```
respectively after the definitions of additive_measure and measure.

theories/measure.v
```diff
@@ -1097,57 +1106,36 @@ apply/idP/idP => [|/imfsetP[r /=]].
...
 
-End AdditiveMeasure.
-Include AdditiveMeasure.Exports.
+Arguments measure_ge0 {R T} _.
 
 Section additive_measure_on_semiring_of_sets.
 Variables (R : realFieldType) (T : semiRingOfSetsType)
   (mu : {additive_measure set T -> \bar R}).
 
 Lemma measure0 : mu set0 = 0.
```
proux01, line 1129: This seems to only require R to be a numFieldType rather than a realFieldType.

theories/measure.v
CohenCyril: One day HB might finally handle that automatically ...

affeldt-aist: Are you suggesting to add this right now or will it be part of a later PR?

CohenCyril: > Are you suggesting to add this right now or will it be part of a later PR?

I'm fine with either. (And the HB stuff is pure fiction for now)

proux01: You can add it now (I checked it doesn't require any adaptation).

theories/lebesgue_integral.v
```diff
@@ -1088,7 +1089,7 @@ Let sintegral_measure_zero (f : T -> R) : sintegral mzero f = 0.
 Proof. by rewrite sintegralE big1// => r _ /=; rewrite /mzero mule0. Qed.
 
 Lemma integral_measure_zero (D : set T) (f : T -> \bar R) :
-  \int_ D f x 'd (@measure_zero T R)[x] = 0.
+  \int_ D f x 'd [the measure _ _ of mzero][x] = 0.
```
CohenCyril, line 1092: doesn't 
```suggestion
  \int_ D f x 'd mzero[x] = 0.
```
work?

theories/lebesgue_integral.v
```diff
@@ -3812,58 +3808,63 @@ Qed.
...
   by rewrite inE.
-suff /cvg_ex[l cl] : cvg (fun n => (\sum_(0 <= i < n) m (F i))%E).
+suff /cvg_ex[l cl] : cvg (fun n => \sum_(0 <= i < n) pm1 (F i)).
   by move: (cl) => /cvg_lim; rewrite -lim_mkord => ->.
 by apply: is_cvg_ereal_nneg_natsum => n _; exact: integral_ge0.
 Qed.
 
-Definition product_measure1 : {measure set (T1 * T2) -> \bar R} :=
-  Measure.Pack _ (Measure.Axioms m0 m_ge0 m_sigma_additive).
+HB.instance Definition product_measure1_build := isMeasure.Build _ _
```
CohenCyril, line 3851: No need to name instances.
```suggestion
HB.instance Definition _ := isMeasure.Build _ _
```

theories/lebesgue_integral.v
affeldt-aist: No, it does not work.

t6s (approved):
Looks good. The new code is cleaner.

affeldt-aist:
After rebasing, I get this error message:
```coq
Let lebesgue_measure_itvoc (a b : R) :
  (lebesgue_measure `]a, b] = hlength `]a, b])%classic.
Error: Stack overflow.
```
What could cause this error? Any hint is welcomed.
By the way, I can avoid the error by writing `]a, b] : set R`.

theories/lebesgue_integral.v
```diff
@@ -1085,7 +1086,7 @@ Let sintegral_measure_zero (f : T -> R) : sintegral mzero f = 0.
 Proof. by rewrite sintegralE big1// => r _ /=; rewrite /mzero mule0. Qed.
 
 Lemma integral_measure_zero (D : set T) (f : T -> \bar R) :
-  \int[@measure_zero T R]_(x in D) f x = 0.
+  \int[[the measure _ _ of mzero]]_(x in D) f x = 0.
```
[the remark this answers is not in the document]
affeldt-aist:
```
The term "mzero" has type "set ?T0 -> \bar ?R0" while it is expected to have type
 "{measure set ?T -> \bar ?R}".
 ```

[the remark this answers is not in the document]
proux01: Couldn't we add a phantom or something in the `\int` notation to avoid that?

affeldt-aist:
In the section below [1], we haved added a type constraint `: set R` around
intervals to avoid a Stack Overflow. I do not remember what change caused that
and if I already told you but since this PR uses an experimental(?) feature of HB I thought you might
want to know @gares @CohenCyril .
--
[1] https://github.com/math-comp/analysis/blob/fixes_642/theories/lebesgue_measure.v#L826-L978

gares:
I'm a bit out of context, but if you tell me how to reproduce the stack overflow I'll look at it.

affeldt-aist:
> I'm a bit out of context, but if you tell me how to reproduce the stack overflow I'll look at it.

You can just remote the type constraint here
https://github.com/math-comp/analysis/blob/fixes_642/theories/lebesgue_measure.v#L830
and compile.

PR #716, opened 2022-07-29
add lemma set_unit

affeldt-aist:

trivial lemma that turns out to be useful

shouldn't it rather be named `setT_unit` (and similarly `set_bool` -> `setT_bool`)?

##### Things done/to do


zstone1 (approved):
I agree with your point on the names. But I guess we don't need to solve the general problem in this PR.

PR #732, opened 2022-08-22
generalization of continuous_measurable_fun

affeldt-aist:

This is a generalization of `continuous_measurable_fun` useful to prove the measurability of e.g. inverse.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -49,7 +49,11 @@
   + lemma `continuous_inP`
 - in `lebesgue_measure.v`:
   + lemma `open_measurable_subspace`
+  + lemma ``subspace_continuous_measurable_fun``
   + corollary `open_continuous_measurable_fun`
+  + lemma `continuous_inP`
+- in `topology.v`:
+  + lemmas `open_setIS`, `open_setSI`, `closed_setIS`, `closed_setSI`
 
 ### Changed
 
@@ -66,10 +70,6 @@
   + change implicits of `integralM_indic`
 - in `derive.v`:
   + generalized `is_diff_scalel`
-- in `topology.v`:
-  + generalize `continuousP` and rename to `continuous_inP`
-- in `lebesgue_measure.v`:
-  + generalize `continuous_measurable_fun`
 
 ### Renamed
 
@@ -94,8 +94,6 @@
   + `nnfun_mulem_ge0` -> `nnsfun_mulemu_ge0`
 - in `esum.v`:
   + `esum0` -> `esum1`
-- in `lebesgue_measure.v`:
-  + `continuous_measurable_fun` -> `continuousT_measurable_fun`
 
 ### Removed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -51,7 +51,6 @@
   + lemma `open_measurable_subspace`
   + lemma ``subspace_continuous_measurable_fun``
   + corollary `open_continuous_measurable_fun`
-  + lemma `continuous_inP`
 - in `topology.v`:
   + lemmas `open_setIS`, `open_setSI`, `closed_setIS`, `closed_setSI`
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -61,6 +61,10 @@
   + change implicits of `integralM_indic`
 - in `derive.v`:
   + generalized `is_diff_scalel`
+- in `topology.v`:
+  + generalize `continuousP`
+- in `lebesgue_measure.v`:
+  + generalize `continuous_measurable_fun`
 
 ### Renamed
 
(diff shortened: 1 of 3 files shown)


theories/lebesgue_measure.v
```diff
@@ -1470,14 +1470,20 @@ move=> q; case: ifPn => // qfab; apply: is_interval_measurable => //.
 exact: is_interval_bigcup_ointsub.
 Qed.
```
zstone1, line 1472: I think we should generalize this further by using the subspace topology. I'm worried that requiring `open D` is too strict because continuous functions from `[0,1]` are important. If we add something here to prove that measurable subspaces have a sensible notion of open...
```suggestion
Lemma open_measurable_subspace (D : set R) (U : set (subspace D)) : 
  measurable D -> open U -> measurable (D `&` U).
Proof.
move=> mD /open_subspaceP [V [oV] VD]; rewrite setIC -VD. 
by apply: measurableI => //; exact: open_measurable.
Qed.
```

theories/lebesgue_measure.v
```diff
@@ -1470,14 +1470,20 @@ move=> q; case: ifPn => // qfab; apply: is_interval_measurable => //.
...
 
-Lemma continuous_measurable_fun (f : R -> R) : continuous f ->
-  measurable_fun setT f.
+Lemma continuous_measurable_fun (f : R -> R) D :
```
zstone1, line 1473: Then we can define this in terms of subspace continuity, which weakens the `open D` to just `measurable D`. Note also that when `D` is open, continuous_open_subspace applies and we can recover the original statement. 
```suggestion
Lemma continuous_measurable_fun (D : set R) (f : subspace D -> R) : 
  measurable D -> continuous f -> measurable_fun D f.
Proof.
move=> mD /continuousP cf; apply: (measurability (RGenOpens.measurableE R)).
move=> _ [_ [a [b ->] <-]]; apply: open_measurable_subspace => //.
by exact/cf/interval_open.
Qed.
```

zstone1 (commented):
Happy to see you generalize this statement. I think we can use the existing subspace machinery to push the generalization further though. Let me know if you think this is worthwhile.

affeldt-aist:
Perfect! I was reluctant to PR because I didn't know how to generalize. :-)

theories/topology.v
```diff
@@ -1731,12 +1731,25 @@ Notation "A ^°" := (interior A) : classical_set_scope.
 
 Notation continuous f := (forall x, f%function @ x --> f%function x).
 
-Lemma continuousP (S T : topologicalType) (f : S -> T) :
-  continuous f <-> forall A, open A -> open (f @^-1` A).
+Lemma continuousP (S T : topologicalType) (f : S -> T) (D : set S) :
```
CohenCyril, line 1734: I'd switch the name and call this one `continuous_inP`

theories/lebesgue_measure.v
affeldt-aist: I'd like to keep the specialized version `open_continuous_measurable_fun` as a corollary.

theories/topology.v
```diff
@@ -1731,12 +1731,25 @@ Notation "A ^°" := (interior A) : classical_set_scope.
 
 Notation continuous f := (forall x, f%function @ x --> f%function x).
 
+Lemma continuous_inP (S T : topologicalType) (f : S -> T) (D : set S) :
```
CohenCyril, line 1734: This is a trivial consequence of continuousP for the subspace toploogy.

theories/lebesgue_measure.v
```diff
@@ -1487,12 +1487,32 @@ move=> q; case: ifPn => // qfab; apply: is_interval_measurable => //.
...
+  measurable D -> open U -> measurable (D `&` U).
 Proof.
-move=> /continuousP cf; apply: (measurability (RGenOpens.measurableE R)).
-move=> _ [_ [a [b ->] <-]]; rewrite setTI.
-by apply: open_measurable; exact/cf/interval_open.
+move=> mD /open_subspaceP [V [oV] VD]; rewrite setIC -VD.
+by apply: measurableI => //; exact: open_measurable.
+Qed.
+
+Lemma continuous_measurable_fun (D : set R) (f : subspace D -> R) :
```
CohenCyril, line 1497:
```suggestion
Lemma subspace_continuous_measurable_fun (D : set R) (f : subspace D -> R) :
```
?
(to preserve the previous names, also subspace topologies are maybe not the most commonly used, so longer name for them)

CHANGELOG_UNRELEASED.md
```diff
@@ -45,6 +45,15 @@
 
 - in `derive.v`:
   + lemma `diff_derivable`
+- in `topology.v`:
+  + lemma `continuous_inP`
+- in `lebesgue_measure.v`:
+  + lemma `open_measurable_subspace`
+  + lemma ``subspace_continuous_measurable_fun``
+  + corollary `open_continuous_measurable_fun`
+  + lemma `continuous_inP`
```
CohenCyril, line 54:
```suggestion
```

PR #757, opened 2022-10-02
sum_fine

affeldt-aist:

To deal with sums of ereals that are actually reals.

##### Things done/to do

### DIFF
--- a/theories/constructive_ereal.v
+++ b/theories/constructive_ereal.v
@@ -785,6 +785,16 @@ Proof. by move: x y => [x| |] [y | |]. Qed.
 Lemma fineK x : x \is a fin_num -> (fine x)%:E = x.
 Proof. by case: x. Qed.
 
+Lemma sum_fine (I : Type) s (P : pred I) (F : I -> \bar R) :
+  (forall i, P i -> F i \is a fin_num) ->
+  (\sum_(i <- s | P i) fine (F i) = fine (\sum_(i <- s | P i) F i))%R.
+Proof.
+move=> h; apply: EFin_inj; rewrite -sumEFin fineK.
+  by apply eq_bigr => ? ?; rewrite fineK// h.
+rewrite sum_fin_num; apply/allP => x; elim: s => //= a b ih.
+by case: ifPn => // /h ? /[!inE] /predU1P[->//|]; exact: ih.
+Qed.
+
 Lemma telescope_sume n m (f : nat -> \bar R) :
   (forall i, (n <= i <= m)%N -> f i \is a fin_num) -> (n <= m)%N ->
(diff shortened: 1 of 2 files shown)


theories/constructive_ereal.v
```diff
@@ -785,6 +785,15 @@ Proof. by move: x y => [x| |] [y | |]. Qed.
 Lemma fineK x : x \is a fin_num -> (fine x)%:E = x.
 Proof. by case: x. Qed.
 
+Lemma sum_fine (I : eqType) s (P : pred I) (F : I -> \bar R) :
+  (forall i, (i \in s) || P i -> F i \is a fin_num) ->
```
zstone1, line 789: The `i \in s` can be safely removed with no change to the proof. In fact, it should be possible to assume `(forall i, (i \in s) && P i -> F i \is a fin_num)` instead, although that will require some change.

zstone1: Oh wow, I have no idea how to get an `i \in s` in scope since `eq_bigr` doesn't do it naturally. I'm happy to just leave it as
```suggestion
  (forall i, P i -> F i \is a fin_num) ->
```

theories/constructive_ereal.v
```diff
@@ -785,6 +785,16 @@ Proof. by move: x y => [x| |] [y | |]. Qed.
 Lemma fineK x : x \is a fin_num -> (fine x)%:E = x.
 Proof. by case: x. Qed.
 
+Lemma sum_fine (I : eqType) s (P : pred I) (F : I -> \bar R) :
+  (forall i, i \in s -> P i -> F i \is a fin_num) ->
+  (\sum_(i <- s | P i) fine (F i) = fine (\sum_(i <- s | P i) F i))%R.
```
CohenCyril, line 790:
```suggestion
Lemma sum_fine (I : Type) s (P : pred I) (F : I -> \bar R) :
  (forall i, P i -> F i \is a fin_num) ->
  (\sum_(i <- s | P i) fine (F i) = fine (\sum_(i <- s | P i) F i))%R.
```
and use `big_seq` or `big_seq_cond` to bridge the gap.

affeldt-aist: I didn't see how to use big_seq{,cond} with Type but that's provable anyway.

CohenCyril: What I meant was that if someone needs to rely on `forall i, i \in s -> P i -> F i \is a fin_num` rather than `forall i, P i -> F i \is a fin_num` as a precondition, they can combine the new `sum_fine` together with `big_seq{,cond}`

affeldt-aist: Ah, ok, this I got.

CohenCyril (approved):

PR #772, opened 2022-10-13
inv_cvg

affeldt-aist:

needed to develop measure theory

##### Things done/to do

### DIFF
--- a/theories/sequences.v
+++ b/theories/sequences.v
@@ -501,6 +501,20 @@ Proof. by move/is_cvgN/cvg_has_sup; rewrite -has_inf_supN image_comp. Qed.
 
 End sequences_R_lemmas_realFieldType.
 
+Lemma inv_cvg (R : realFieldType) (u : R ^nat) :
+  (forall n, 0 < u n) -> (fun n => (u n)^-1) --> 0 -> u --> +oo.
+Proof.
+move=> u0 uV0; apply/cvgPpinfty => M.
+move/cvg_distP : uV0 => /(_ (`|M| + 1)^-1%R).
+rewrite invr_gt0 ltr_paddl// => /(_ erefl); rewrite !near_map.
+apply: filterS => n.
+rewrite sub0r normrN normrV ?unitfE ?gt_eqF// ger0_norm; last exact: ltW.
+rewrite ltr_pinv.
+- by move/ltW; apply: le_trans; rewrite (le_trans (ler_norm _))// ler_addl.
+- by rewrite inE unitfE u0 andbT gt_eqF.
+- by rewrite inE unitfE ltr_paddl// andbT gt_eqF.
+Qed.
+
 Section partial_sum.
 Variables (V : zmodType) (u_ : V ^nat).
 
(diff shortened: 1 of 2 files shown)


zstone1 (approved):

CohenCyril (changes requested):
@affeldt-aist the name was clearly too general `inv_cvg` was far too ambiguous (note that `cvgV` also exist). I gave a suggestion.
I added a few missing helper lemmas, however like many lemmas here, they could and should actually be generalized to filters rather than mere sequences.

CohenCyril:
Also the condition `u i > 0` could be restricted to "near `\oo`"

theories/sequences.v
```diff
@@ -1791,11 +1793,11 @@ move=> f0 g0; rewrite /adde_def !negb_and; apply/andP; split; apply/orP.
 Qed.
 
 Lemma ereal_cvgPpinfty (R : realFieldType) (u_ : (\bar R)^nat) :
-  u_ --> +oo <-> (forall A, (0 < A)%R -> \forall n \near \oo, A%:E <= u_ n).
+  u_ --> +oo <-> (forall A, \forall n \near \oo, A%:E <= u_ n).
```
CohenCyril, line 1796: This is actually neither a generalization nor a weakening. The following are all equivalent:
1. `u --> +oo`
2. `forall A, \forall n \near \oo, A%:E <= u_ n`
3. `\forall A \near +oo, \forall n \near \oo, A%:E <= u_ n`

When faced with 1.:
- as a goal it's actually better to turn it into 3., then you can use `near=> A` and assume `A` to be as large as you need.
- but as an hypothesis, it's better to get 2., so you can apply it to anything without any constraint on `A`

CohenCyril: Maybe we should actually write it
```suggestion
  [<-> u_ --> +oo,
         forall A, \forall n \near \oo, A%:E <= u_ n &
         \forall A \near +oo, \forall n \near \oo, A%:E <= u_ n ]
```
?

theories/sequences.v
```diff
@@ -499,8 +499,85 @@ Qed.
 Lemma cvg_has_inf u_ : cvg u_ -> has_inf (u_ @` setT).
 Proof. by move/is_cvgN/cvg_has_sup; rewrite -has_inf_supN image_comp. Qed.
 
+Lemma cvgPpinfty_lt (u_ : R ^nat) :
+  u_ --> +oo%R <-> forall A, \forall n \near \oo, (A < u_ n)%R.
+Proof.
+split=> [uoo A|uB]; last by apply/cvgPpinfty => A; near=> n; apply/ltW; near: n.
+near=> n; rewrite (@lt_le_trans _ _ (A + 1)) ?ltr_addl//.
+by near: n; apply: (cvgPpinfty _).1.
+Unshelve. all: by end_near. Qed.
```
CohenCyril, line 508: Rewrite with `[<-> ..., ... ]` and merge with `cvgPpinfty_lt_near` ?

CohenCyril: And the same for everything here?

CohenCyril:
It seems that I hijacked this PR.

CohenCyril:
@affeldt-aist I will move my changes to another PR. I'll truncate this one at e61b452d0b7f2b099911c823dfac40c2585adb65 and merge

CohenCyril:
Opened https://github.com/math-comp/analysis/pull/780

CohenCyril:
Rebased

CohenCyril (approved):

CohenCyril:
Rebase, merge when CI passes

PR #788, opened 2022-10-28
measurable_fun_if

affeldt-aist:

This lemma has found a use in PR #749 and in PR #769 so it is maybe worth integrating master.

@t6s @hoheinzollern 

##### Things done/to do

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -60,8 +60,10 @@ From HB Require Import structures.
 (*                            of measurableType                               *)
 (*  G.-sigma.-measurable A == A is measurable for the sigma-algebra <<s G >>  *)
 (*                                                                            *)
-(* discrete_measurable_nat == the measurableType corresponding to             *)
-(*                            [set: set nat]                                  *)
+(* discrete_measurable_bool == the measurableType corresponding to            *)
+(*                             [set: set bool]                                *)
+(*  discrete_measurable_nat == the measurableType corresponding to            *)
+(*                             [set: set nat]                                 *)
 (*          salgebraType G == the measurableType corresponding to <<s G >>    *)
 (*                                                                            *)
 (*      measurable_fun D f == the function f with domain D is measurable      *)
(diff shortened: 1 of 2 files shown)


theories/measure.v
```diff
@@ -996,6 +1020,42 @@ move=> mD mE DE; split => mf _ /= X mX.
   by rewrite setIA (setIidl DE) setIUr setICr set0U.
 Qed.
 
+Lemma measurable_fun_if (g h : T1 -> T2) D (mD : measurable D)
+    (f : T1 -> bool) (mf : measurable_fun setT f) :
```
zstone1, line 1024:
```suggestion
    (f : T1 -> bool) (mf : measurable_fun D f) :
```

theories/measure.v
```diff
@@ -996,6 +1020,42 @@ move=> mD mE DE; split => mf _ /= X mX.
...
+  measurable_fun (D `&` (f @^-1` [set true])) g ->
+  measurable_fun (D `&` (f @^-1` [set false])) h ->
+  measurable_fun D (fun t => if f t then g t else h t).
+Proof.
+move=> mx my /= _ B mB.
+have mDf : measurable (D `&` [set b | f b]).
+  apply: measurableI => //.
+  by have := mf measurableT [set true]; rewrite setTI; exact.
```
zstone1, line 1032:
```suggestion
have mDf : measurable (D `&` [set b | f b]) by exact: mf.
```

theories/measure.v
```diff
@@ -996,6 +1020,42 @@ move=> mD mE DE; split => mf _ /= X mX.
...
+have := mx mDf _ mB.
+have mDNf : measurable (D `&` f @^-1` [set false]).
+  apply: measurableI => //.
+  by have := mf measurableT [set false]; rewrite setTI; exact.
```
zstone1, line 1036:
```suggestion
have mDNf : measurable (D `&` f @^-1` [set false]) by exact: mf.
```

zstone1 (changes requested):
The condition `measurable_fun setT f` is too strong. The suggestions should weaken it to `measurable_fun D f`, which is somewhat stronger.

theories/measure.v
affeldt-aist: Indeed. I push it in a next commit with a shorter proof.

zstone1 (approved):

PR #792, opened 2022-11-02
Basic fact about T1

affeldt-aist:

basic fact about T1

##### Things done/to do

### DIFF
--- a/classical/fsbigop.v
+++ b/classical/fsbigop.v
@@ -484,6 +484,13 @@ Lemma fsbig_setU {T} {I : choiceType} (A : set I) (F : I -> set T) :
   \big[setU/set0]_(i \in A) F i = \bigcup_(i in A) F i.
 Proof. by move=> Afin; rewrite fsbig_finite// bigcup_fset_set. Qed.
 
+Lemma fsbig_setU_set1 {T : choiceType} (A : set T) :
+  finite_set A -> \big[setU/set0]_(x \in A) [set x] = A.
+Proof.
+move=> fA; rewrite fsbig_setU//.
+by apply/seteqP; split=> [t [x Ax ->]//|t At]; exists t.
+Qed.
+
 Lemma pair_fsbig (R : Type) (idx : R) (op : Monoid.com_law idx)
     (I J : choiceType) (P : set I) (Q : set J) (F : I -> J -> R) :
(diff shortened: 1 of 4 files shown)


classical/fsbigop.v
```diff
@@ -484,6 +484,13 @@ Lemma fsbig_setU {T} {I : choiceType} (A : set I) (F : I -> set T) :
   \big[setU/set0]_(i \in A) F i = \bigcup_(i in A) F i.
 Proof. by move=> Afin; rewrite fsbig_finite// bigcup_fset_set. Qed.
 
+Lemma finite_set_fsbig_set1 {T : choiceType} (A : set T) :
+  finite_set A -> \big[setU/set0]_(x \in A) [set x] = A.
+Proof.
+move=> fA; rewrite fsbig_setU//.
+by apply/seteqP; split=> [t [x Ax ->]//|t At]; exists t.
+Qed.
+
```
CohenCyril, line 493: No need to prefix with `finite_set`, lemmas about `fsbig` are supposed to be valid only for finite sets anyways!

CohenCyril: However it matters that it's a big op on `setU` so it it chould be in the name, e.g. `fsbig_setU_set1`.

CohenCyril:
> My only note is on the addition of another variety of bigop in `topology.v`. Most of the theorems in this file use `{fset I}` + `\bigcup` (like `filter_bigI`) instead of `{seq I}` + `\big[setU/set0]`. I can see a benefit to separating the "definitely finite bigops" (like `filter_bigI`) from "maybe infinite bigops" (like bigcup_open). Among the "definitely finite bigops", is there a reason to have both the `{fset I}` and `{seq I}` representations?

If the `seq` one is true there is no reason to have the `fset` one (since the `seq` one is more general). But I'm not sure how it relates to this PR... which adds a lemma about a third category of bigop lemmas, the "valid-only-if-finite bigop, otherwise defaults to neutral"

zstone1:
Ah, you're right. This is more general. I think there might be some benefit to replace some of the existing fset style lemmas with finite support versions. It might actually simplify things a bit. I'll add an issue for trying it. This looks good then.

zstone1 (approved):

CohenCyril (approved):

PR #794, opened 2022-11-04
 Introduce convex function on real numbers

yoshihiro503:
Co-authored-by: Reynald Affeldt <reynald.affeldt@aist.go.jp>

##### Things done/to do

### DIFF
--- a/theories/convex.v
+++ b/theories/convex.v
@@ -115,9 +115,126 @@ have -> : ((x - a) / (b - a) = p).
 done.
 Qed.
 
+Variable pderivable : (R -> R) -> (R -> Prop) -> Prop.
+Definition derive_pt (f: R -> R) (x: R) : R := f^`() x.
+Let I := fun x0 => (a <= x0 <= b).
+Hypothesis HDf : pderivable f I.
+
+Lemma MVT_cor1_pderivable
+     : forall (a b : R) (f : R -> R) (pr : pderivable f (fun x : R => a <= x <= b)),
+    a < b ->
+    exists (c : R) (Hc : a <= c <= b),
+      f b - f a = derive_pt f c * (b - a) /\ a < c < b.
+Proof.
+Admitted.
+(*use MVT in derive.v *)
+
+Hypothesis DfE : forall (x : R) (Hx : I x), Df x = derive_pt f x.
+Hypothesis DDfE : forall x (Hx : I x), DDf x = derive_pt Df x.
+
+Lemma unitfB: forall (x y:R), x < y -> y - x \is a GRing.unit.
+Proof.
+  by move=> x y xy; rewrite unitfE subr_eq0 (gt_eqF xy).
+Qed.
+
 Lemma second_derivative_convexf_pt : forall t : prob,
     f (conv a b t) <= conv (f a) (f b) t.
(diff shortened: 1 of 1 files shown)


theories/convex.v
```diff
@@ -0,0 +1,254 @@
...
+End Exports.
+End Prob.
+Export Prob.Exports.
+Notation prob := Prob.t.
+Notation "q %:pr" := (@Prob.mk _ q (@Prob.O1 _ _)).
+Coercion Prob.p : prob >-> Num.NumDomain.sort.
+(*  ^^^ probability ^^^ *)
+
+Definition derivable_interval {R : numFieldType}(f : R -> R) (a b: R) :=
+  forall x, x \in `[a, b] -> derivable f x 1.
```
zstone1, line 65: This definition misbehaves at the endpoints. In particular, it requires `f` to be be "differentiable from the left at `a`". That makes this definition rely on values of `f` outside `[a,b]`. One big consequence is characteristic functions like ``if x \in `[a,b] then 1 else 0`` fails this definition, as its not derivable at `a` or `b`. For what it's worth, I had an analogous issue with `realfun.v`, see #752 for how I addressed it there. You'll need something slightly different here, though.

I've seen textbooks take these approaches:
1. Demanding "continuity up the endpoints" is one approach
```
forall x, x \in `]a, b[ -> derivable f x 1.
/\
cvg (at_right a f^`()) /\ cvg (at_right b f^`())
```
or some variation of this. However, this implies some continuity of the derivative, which is maybe not ideal.
2. We could change the derivative to use a different topology. One way to do that is add an analog to the ``{within `[a,b], continuous f}`` notation of  ``{within `[a,b], derivable f}``. This would take limits only on the subspace topology. 
3. Similarly, but specialized to working in `R`, it should be fine to define it like
```
forall x, x \in `]a, b[ -> derivable f x 1
/\
"differentiable from the right at b" /\ "differentiable from the left at a"
```
For some sensible notion of "left" and "right" derivative.

theories/convex.v
```diff
@@ -0,0 +1,254 @@
...
+Proof.
+rewrite/derivable_interval => b'b deriv x xin. apply/deriv.
+by move: xin; apply in_IntervalW.
+Qed.
+
+(* ref: http://www.math.wisc.edu/~nagel/convexity.pdf *)
+Section twice_derivable_convex.
+
+Variable R : realType.
+Variables (f : R -> R) (a b : R).
```
zstone1, line 94: Where do we need the assumption that the range of `f` is `R`? If it's painless, generalizing to a vector space would be quite helpful for me. I'll need some stuff about convexity in both `R` and `R^n` for my work on space-filling curves, so I'm happy to have these results either way.

zstone1 (changes requested):
Nice results. I've made comments about a couple generalizations, one regarding annoying endpoint issues, the other regarding generalizing to vector space ranges. I care more about the endpoints one mostly because it's much harder to fix later. Happy to work with you on that if that'd be helpful.

affeldt-aist:
I opened PR #873 and moved the comments by @zstone1 there because it looked like maintainers do not have write access (not 100% sure though). I hope you don't mind moving the conversation there.

PR #804, opened 2022-11-24
Adapt to coq/coq#16798 (cast in pattern is an error by default)

SkySkimmer:
A pattern `x : y` is equivalent to just `x`.

Some of the casts were bogus and so were removed.

Other casts come from notations (from constructive_ereals around line 260), for these the warning was ignored as it seems ignoring them was not a problem.


SkySkimmer:
(if you think making this warning error by default this is the right time to say so, the right place is the coq side PR)

affeldt-aist (approved):
I am ok with these changes.

CohenCyril:
I'm not so happy about them...

affeldt-aist:
Do you mean that you are unhappy with the fact that they are ineffective?

CohenCyril:
No I'm unhappy about the verbosity

PR #813, opened 2022-12-21
minor generalization in esum.v

affeldt-aist:
- generalize esum_esum
- rename esum_fset to esum_fsbig
- remove fsbig_esum (redundant)
- fix implicit arguments of eq_fsbigr

##### Things done/to do

### DIFF
--- a/theories/esum.v
+++ b/theories/esum.v
@@ -97,10 +97,6 @@ Proof.
 by move=> ?; rewrite esum_fset// ?fset_set1// ?fsbig_set1// => t' /[!inE] ->.
 Qed.
 
-Lemma fsbig_esum (A : set T) a : finite_set A -> (forall x, 0 <= a x) ->
-  \sum_(x \in A) (a x) = \esum_(x in A) a x.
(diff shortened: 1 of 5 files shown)


theories/esum.v
```diff
@@ -81,7 +81,7 @@ move=> a0; apply: ereal_sup_ub.
 by exists set0; [exact: fsets_set0|rewrite fsbig_set0].
 Qed.
 
-Lemma esum_fset (F : set T) a : finite_set F ->
+Lemma esum_fsbig (F : set T) a : finite_set F ->
```
CohenCyril, line 84: I'm not sure about this renaming...

affeldt-aist: It's based on the fact that the removed lemma was `fsbig_esum` and it was the `esym` version of this lemma.
What about removing `enum_fset` and keep `fsbig_esum` instead?

CohenCyril: I prefer to keep this one (`fsbig_esum` was a weird name in the first place...) and keep the name as such...

CohenCyril (approved):

PR #777, opened 2022-10-18
Hahn decomposition theorem

affeldt-aist:
- introduces signed measures

Co-authored-by: IshiguroYoshihiro

~~based on PR #836~~ (has been merged into master) 

##### Things done/to do

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [x] added corresponding documentation in the headers

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and put a milestone if possible.

PR #818, opened 2023-01-03
tentative proof of Radon-Nikodym

affeldt-aist:

based on PR #777 

##### Things done/to do

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -1303,7 +1303,7 @@ Qed.
 
 HB.mixin Record isAdditiveMeasure d
     (R : numFieldType) (T : semiRingOfSetsType d) (mu : set T -> \bar R) := {
-  measure_ge0 : forall x, 0 <= mu x ;
+  measure_ge0 : forall x, measurable x -> 0 <= mu x ;
   measure_semi_additive : semi_additive mu }.
 
 HB.structure Definition AdditiveMeasure d
@@ -1322,11 +1322,11 @@ Context d (R : numFieldType) (T : semiRingOfSetsType d).
 
 Variable mu : {additive_measure set T -> \bar R}.
 
-Lemma additive_measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
+(*Lemma additive_measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
 Proof. exact: measure_ge0. Qed.
 
 Canonical additive_measure_snum S :=
-  Signed.mk (additive_measure_snum_subproof S).
+  Signed.mk (additive_measure_snum_subproof S).*)
 
 End additive_measure_signed.
 
@@ -1474,17 +1474,17 @@ Context d (R : numFieldType) (T : semiRingOfSetsType d).
 
 Variable mu : {measure set T -> \bar R}.
 
-Lemma measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
+(*Lemma measure_snum_subproof S : Signed.spec 0 ?=0 >=0 (mu S).
 Proof. exact: measure_ge0. Qed.
 
-Canonical measure_snum S := Signed.mk (measure_snum_subproof S).
+Canonical measure_snum S := Signed.mk (measure_snum_subproof S).}*)
 
 End measure_signed.
 
 HB.factory Record isMeasure d
     (R : realFieldType) (T : semiRingOfSetsType d) (mu : set T -> \bar R) := {
   measure0 : mu set0 = 0 ;
-  measure_ge0 : forall x, 0 <= mu x ;
+  measure_ge0 : forall x, measurable x -> 0 <= mu x ;
   measure_semi_sigma_additive : semi_sigma_additive mu }.
 
 HB.builders Context d (R : realFieldType) (T : semiRingOfSetsType d)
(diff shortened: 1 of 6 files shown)


affeldt-aist:
ping @zstone1 (the remaining errors in the CI are not from this PR)

zstone1:
Sorry for the delay, I expect to have some time for a review this weekend.

theories/charge.v
```diff
@@ -698,3 +766,774 @@ split.
...
+      apply: ge0_le_integral => //.
+      * apply: measurableT_comp => //; apply: emeasurable_funD => //.
+        exact: measurable_funS measurable_fun_fRN.
+      * apply: emeasurable_funD => //; apply: measurableT_comp => //.
+        exact: measurable_funS measurable_fun_fRN.
+      * by move=> x _; exact: lee_abs_add.
+    apply: lte_add_pinfty.
+      rewrite (le_lt_trans _ int_fRN_lty)// subset_integral//.
+      by apply: measurableT_comp => //; exact: measurable_fun_fRN.
+    by rewrite integral_cst//= lte_mul_pinfty// ltey_eq fin_num_measure.
```
zstone1, line 1310:
```suggestion
  + apply: integrableD => //. 
      (apply: integrableS; first exact: measurableT) => //; apply /integrableP. 
      by split; [exact: measurable_fun_fRN | exact: int_fRN_lty].
    apply/integrableP; split; rewrite // integral_cst // lte_mul_pinfty //. 
    by rewrite ltey_eq fin_num_measure.
```

zstone1 (approved):
It seems like a pretty faithful formalization of the textbook proof. I'm happy with the overall structure, and putting all these one-time use definitions into a module is a good idea for sure. I found one moderate simplification regarding `fRN + eps`. Basically shuffling the `int fRN x  + eps` to `int fRN x + int eps` before the `integrableP` saves some headache. There are probably a few other similar proof improvements in the proof of `radon_nikodym_finite`, but I didn't find any trivial ones.

theories/measure.v
```diff
@@ -4140,3 +4171,18 @@ End partial_measurable_fun.
   solve [apply: measurable_pair1] : core.
 #[global] Hint Extern 0 (measurable_fun _ (pair^~ _)) =>
   solve [apply: measurable_pair2] : core.
+
+Section absolute_continuity.
+Context d (T : measurableType d) (R : realType).
+Implicit Types m : set T -> \bar R.
+
+Definition dominates m1 m2 :=
```
CohenCyril, line 4179: Maybe we should have a more specific name for this?
```suggestion
Definition measure_dominates m1 m2 :=
```

theories/charge.v
```diff
@@ -698,3 +766,774 @@ split.
...
+- by rewrite -setI_bigcupr bigcupE setIT.
+- by move=> i _; exact: measurableI.
+- exact: trivIset_setIl.
+Qed.
+
+Theorem Radon_Nikodym
+  (mu : {sigma_finite_measure set T -> \bar R}) (nu : {charge set T -> \bar R}) :
+  nu `<< mu ->
+  exists2 f : T -> \bar R, mu.-integrable [set: T] f &
+    forall E, measurable E -> nu E = \int[mu]_(x in E) f x.
```
CohenCyril, line 1523: Should we extract this existential and have a function `Randon_Nikodym mu nu` that is `mu.-integrable` and such that:
 ```coq
forall E, measurable E -> nu E = \int[mu]_(x in E) Radon_Nikodym mu nu x.
```
?

**Review by CohenCyril — APPROVED** [review]
Really nice. Just two small remarks.

File: theories/charge.v (hunk 1 above)
Comment (affeldt-aist) [reply theories/charge.v]: Oh! Thanks. I committed your suggestion after a rebase: https://github.com/math-comp/analysis/pull/818/commits/0a13408f8102424753812883a09ad0de8ba9052f

Comment (affeldt-aist) [reply theories/charge.v]: > There are probably a few other similar proof improvements in the proof of radon_nikodym_finite, but I didn't find any trivial ones.

I still have to take some time to look for more simplifications.

File: theories/charge.v (hunk 3 above)
Comment (affeldt-aist) [reply theories/charge.v]: That would indeed look better. Here is a first try: https://github.com/math-comp/analysis/pull/818/commits/04dfc2ca6417c8b330ab4401c5f5747222d6d18e

Comment (CohenCyril) [reply theories/charge.v]: Indeed it looks better...
I guess the code would be slightly shorter if you used the same Theorem `Radon_Nikodym` as before (maybe renaming it?) but extracted the witness out of it, rather than having both a long definition and proof for extracting the existential.

File: CHANGELOG_UNRELEASED.md [hunk 4]
```diff
@@ -64,6 +64,21 @@
...
+  + lemma `dominates_trans`
+- in `measure.v`:
+  + defintion `mfrestr`
+- in `charge.v`:
+  + definition `measure_of_charge`
+  + definition `crestr0`
+  + definitions `jordan_neg`, `jordan_pos`
+  + lemmas `jordan_decomp`, `jordan_pos_dominates`, `jordan_neg_dominates`
+  + lemma `radon_nikodym_finite`
+  + definition `Radon_Nikodym`, notation `'d nu /d mu`
```
CohenCyril, line 80: Where is it defined?
If you define it, which I think is nice, we must put it in a special scope because we might want the same kind of notation for partial derivatives someday.

affeldt-aist: Yes, I messed up with my commits, wait a minute.

theories/measure.v
CohenCyril: The name is back to `dominates`

CohenCyril (approved):
I'm fine with the changes. And root for merging.

(I involuntarily re-requested a review by @zstone1 but since he agreed before I think we're good to merge?)

zstone1 (approved):

affeldt-aist:
Thank for the careful reviews! @zstone1 @CohenCyril

PR #829, opened 2023-01-26
expeS, fin_numX

affeldt-aist:

easy lemmas added to `constructive_ereal.v`

##### Things done/to do

### DIFF
--- a/theories/constructive_ereal.v
+++ b/theories/constructive_ereal.v
@@ -717,6 +717,9 @@ Proof. by move: x => [r| |]/=; rewrite /mule/= ?mul0r// eqxx. Qed.
 
 Canonical mule_mulmonoid := @Monoid.MulLaw _ _ mule mul0e mule0.
 
+Lemma expeS x n : x ^+ n.+1 = x * x ^+ n.
+Proof. by case: n => //=; rewrite mule1. Qed.
+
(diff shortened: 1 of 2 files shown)


theories/constructive_ereal.v
```diff
@@ -814,6 +817,9 @@ Lemma fin_numM x y : x \is a fin_num -> y \is a fin_num ->
   x * y \is a fin_num.
 Proof. by move: x y => [x| |] [y| |]. Qed.
 
+Lemma fin_numX x n : x \is a fin_num -> x ^+ n \is a fin_num.
```
zstone1, line 820: Maybe fin_num_expe?

zstone1 (approved):
Small naming change suggestion.

theories/constructive_ereal.v
affeldt-aist: I think that `fin_numX` is appropriate because the suffix `X` is used in most lemmas involving `^+`.
Of course, here we are indeed talking about `expe` and not the operator coming from MathComp
but since this is a lemma about `fun_num` the user can infer this difference from the context.

zstone1: Ah, makes sense. I'm happy with it's current state then.

PR #830, opened 2023-01-26
lemma countable_bijP

affeldt-aist:
Co-authored-by: Cyril Cohen <cohen@crans.org>

Missing lemma in `functions.v`.

Also adds two small pieces of documentation to `functions.v`.
There is of course much more to do to document `functions.v` but
I cannot figure out right way a good format and the adequate level of details,
I will come back to it later.

##### Things done/to do


affeldt-aist:
> I do not see `countable_bijP` in the diff... is that an oversight?

Yes, sorry about that. It is fixed now.

PR #832, opened 2023-01-31
rename and doc finSubCover

### LINKED ISSUE #658: `finSubCover` lacks documentation
https://github.com/math-comp/analysis/blob/22bb6cc4adc629f47b4ec03ef38eb2138ed2c9a5/theories/topology.v#L3013-L3016

affeldt-aist:

fixes #658 

I added a documentation entry.
Since the naming was questioned, I propose a new one that essentially gets rid of camlCase
which is more commonly used for mixins and structures in MathComp.
I haven't changed the definition as suggested by the issue.

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -63,6 +63,8 @@
 - in `constructive_ereal.v`:
   + lemmas `adde_def_doppeD`, `adde_def_doppeB`
   + lemma `fin_num_sume_distrr`
+- in `classical_sets.v`:
+  + lemma `coverE`
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


affeldt-aist:
Quote from issue #658:
> The name suggests that it is pinpointing a (finite) subcover of some cover, but there is no such original cover in the definition.

Indeed.
At first I thought of renaming to `finite_subset_cover` so that it sounds more like we take the subset of some set but this might look like it points to the definition of `cover` in `classical_sets.v` that we do not use here.
Hence this minima change.

affeldt-aist:
Or maybe we can try to use `cover`: https://github.com/math-comp/analysis/pull/832/commits/62440d95726d44b43016b334fdb84ce7520b7e88

And then the `finite_subset_cover` makes more sense.

affeldt-aist:
@zstone1 do you think that this is a reasonable change?

zstone1 (approved):
Looks good to me. That name always threw me off.

affeldt-aist:
@t6s do you think this PR fixes your issue?

t6s (approved):

PR #790, opened 2022-10-30
Specialized Cauchy-Schwarz for expectations

hoheinzollern:
##### Motivation for this change

Define covariance and present cauchy_schwarz

mul_RV and related lemmas are required but part of a separate pull request.

##### Things done/to do

- [ ] complete cauchy_schwarz definitions and lemmas

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and put a milestone if possible.

PR #836, opened 2023-02-06
add a type for finite measures

affeldt-aist:

This PR improves the hierarchy of measures as follows:

charge_tred

Rounded boxes were already there.
Finite measures were already there but through the predicate `finite_measure` only.
Square boxes are new.
Dashed boxes and `FinNumFun` come with this PR.
`AdditiveCharge` and `Charge` come with PR #777 (rebased on this PR)

This PR is necessary to rebase the branch `kernels` PR #749 because
the latter proves Fubini for s-finite measures which relies
on Fubini for sigma-finite measures which are given
a type using HB which calls for automatic inference
of the type of s-finite measures from the type
of sigma-finite measures.
In the branch `kernels` we also need to equip the return type of
s-finite kernels with the type of s-finite measures to apply Fubini
for s-finite measures in the proof of commutativity of let-in.

This PR also adde subprobability measures (just to match the hierarchy of kernels)

##### Things done/to do

### DIFF
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -87,12 +87,14 @@ From HB Require Import structures.
 (*                   that it is semi_sigma_additive                           *)
 (*     isMeasure == factory corresponding to the type of measures             *)
 (*     Measure == structure corresponding to measures                         *)
-(*     finite_measure mu == the measure mu is finite                          *)
 (*  {sigma_finite_content set T -> \bar R} == contents that are also sigma    *)
 (*                                            finite                          *)
 (*  {sigma_finite_measure set T -> \bar R} ==                                 *)
 (*                    measures that are also sigma finite                     *)
 (*     isSigmaFinite == factory corresponding to sigma finiteness             *)
+(*  isFiniteMeasureFunction == mixin for a function mu : set T -> \bar R such *)
+(*                      that mu A is a fin_num for any measurable set A       *)
+(*  {finite_measure set T -> \bar R} == finite measure                        *)
 (*                                                                            *)
 (*  pushforward mf m == pushforward/image measure of m by f, where mf is a    *)
 (*                      proof that f is measurable                            *)
@@ -1567,20 +1569,6 @@ Arguments measure_bigcup {d R T} _ _.
 #[global] Hint Extern 0 (sigma_additive _) =>
   solve [apply: measure_sigma_additive] : core.
 
-Definition finite_measure d (T : measurableType d) (R : numDomainType)
-    (mu : set T -> \bar R) :=
-  mu setT < +oo.
-
-Lemma finite_measure_sigma_finite d (T : measurableType d) (R : realFieldType)
-  (mu : {measure set T -> \bar R}) :
-  finite_measure mu -> sigma_finite setT mu.
-Proof.
-exists (fun i => if i \in [set 0%N] then setT else set0).
-  by rewrite -bigcup_mkcondr setTI bigcup_const//; exists 0%N.
-move=> n; split; first by case: ifPn.
-by case: ifPn => // _; rewrite ?measure0//; exact: finite_measure.
-Qed.
-
 HB.mixin Record isSigmaFinite d (R : numFieldType) (T : semiRingOfSetsType d)
     (mu : set T -> \bar R) := {
   sigma_finiteT : sigma_finite setT mu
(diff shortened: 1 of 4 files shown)


affeldt-aist:
> it is also required to rebase PR #749

it looks like rebasing will require a bit more thinking

hoheinzollern:
This PR is interesting for PR #790, since we need finite measures to prove $L_2 \subset L_1$, which only holds in the case the measure is finite, like with probabilities. See for example [here](https://ocw.mit.edu/courses/18-125-measure-and-integration-fall-2003/6f21af6c40de1eccd70349bd3a3b0095_18125_lec17.pdf).

theories/measure.v
```diff
@@ -2567,6 +2535,327 @@ HB.instance Definition _ := isMeasure0.Build _ _ _ Rmu
...
+#[global] Hint Resolve sigma_finiteT : core.
+
+Notation "{ 'sigma_finite_content' 'set' T '->' '\bar' R }" :=
+  (sigma_finite_content T R) (at level 36, T, R at next level,
+    format "{ 'sigma_finite_content'  'set'  T  '->'  '\bar'  R }")
+  : ring_scope.
+
+#[short(type="sigma_finite_measure")]
+HB.structure Definition SigmaFiniteMeasure d T R :=
+  { mu of @SFiniteMeasure d T R mu & isSigmaFinite d T R mu }.
```
zstone1, line 2613: I'm surprised to see the `SFiniteMeasure` mixin instead of just `isMeasure` mixin. I see you use the `Measure_isSigmaFinite` to build it instead. Is there a particular reason to take this approach?

theories/measure.v
```diff
@@ -2567,6 +2535,327 @@ HB.instance Definition _ := isMeasure0.Build _ _ _ Rmu
...
+
+HB.mixin Record SigmaFinite_isFinite d (T : semiRingOfSetsType d)
+    (R : numDomainType) (k : set T -> \bar R) :=
+  { fin_num_measure : fin_num_fun k }.
+
+HB.structure Definition FinNumFun d (T : semiRingOfSetsType d)
+    (R : numFieldType) := { k of SigmaFinite_isFinite _ T R k }.
+
+HB.structure Definition FiniteMeasure d (T : measurableType d) (R : realType) :=
+  { k of @SigmaFiniteMeasure _ _ _ k & SigmaFinite_isFinite _ T R k }.
```
zstone1, line 2673: Same question as above: Why not use the `isMeasure` mixin?

theories/measure.v
```diff
@@ -2567,6 +2535,327 @@ HB.instance Definition _ := isMeasure0.Build _ _ _ Rmu
...
+Definition finite_measure d (T : semiRingOfSetsType d) (R : numDomainType)
+    (mu : set T -> \bar R) := mu setT < +oo.
+
+Definition sfinite_measure_def d (T : measurableType d) (R : realType)
+    (mu : set T -> \bar R) :=
+  exists2 s : {measure set T -> \bar R}^nat,
+    forall n, finite_measure (s n) &
+    forall U, measurable U -> mu U = mseries s 0 U.
+
+Definition sigma_finite d (T : semiRingOfSetsType d) (R : numDomainType)
```
zstone1, line 2547: Taking this `A : set T` is slightly uncomfortable for all of the substructuring reasons we've discussed before. 

While building better support for substructures is something we're working on elsewhere. For now it looks like we only use `(A := setT)` anyway. Is it better to just eliminate `A` as an argument?

zstone1 (commented):
The hierarchy makes sense, but I've got a few questions about HB quirks. But this generally looks good.

theories/measure.v
affeldt-aist: We took this approach primarily because this is the one that succeeded for PR #749 where the hierarchy is similar (including the apparent circularity between the definitions of s-finite and finite). Still, since sigma-finite measures are s-finite measures I would have thought that inheriting directly from them is in fact the least surprising approach. Using `isMeasure` instead of `SFiniteMeasure` would mean that one would have to do something more to guarantee that sigma-finite measures are recognized as s-finite measures. As you noticed, the "true" definition of sigma-finite measure is rather to be looked for in the factory `Measure_isSigmaFinite`. Something similar happens for, say, algebra of sets: the mixin `AlgebraOfSets_from_RingOfSets` that defines them is one possible definition (it adds the minimum to extend a ring of sets) but the direct definition that is more likely to appear in a textbook corresponds to the factory `isAlgebraOfSets`.

theories/measure.v
affeldt-aist: This is indeed a generality that we do not use, see f4b65719817d708f038b0ca06fdb1d13e7054569 . It looks fine to me to eliminate it. @CohenCyril ?

theories/measure.v
affeldt-aist: My answer would be the same as above. Maybe I can also refer to Sect. 4 of https://hal.inria.fr/hal-03917948/document that explains the hierarchy in the case of kernels (which is a generalization).

affeldt-aist:
Is it ok to keep both definitions `finite_measure` and `fin_num_fun` or should I try to get rid of the former for the latter?

affeldt-aist:
> Is it ok to keep both definitions `finite_measure` and `fin_num_fun` or should I try to get rid of the former for the latter?

It seems clearer to get rid of `finite_measure` and keep `fin_num_fun` only: https://github.com/math-comp/analysis/pull/836/commits/af15a0c7c0e1f1db037afee1496a318b1fc398b7

zstone1 (approved):
Happy either way with the extra parameter.

affeldt-aist:
I have applied a few more comments that I received offline except one requesting a set parameter to `fin_num_fun`. Is it ok if I issue about it and take care of it later?

zstone1:
If you're asking me, I think that's fine. It doesn't sound like a big burden to do later, and I'd rather get you unblocked now.

CohenCyril:
> 

I agree

CohenCyril:
> If you're asking me, I think that's fine. It doesn't sound like a big burden to do later, and I'd rather get you unblocked now.

I agree

PR #873, opened 2023-03-13
Convex exp

affeldt-aist:

This is a continuation of PR #794 (I have duplicated the PR because it looks like I cannot push commits to it).
On top of PR #794, it introduces convex spaces and use the type `{i01 R}` from the PR #869.

This PR introduce convexs spaces, show that `lmodType` form instances, and use them to show that a function with a non-negative second derivative is convex (like the exponential function @hoheinzollern).

closes #794 

##### Things done/to do

### DIFF
--- a/theories/convex.v
+++ b/theories/convex.v
@@ -9,10 +9,11 @@ From HB Require Import structures.
 
 (******************************************************************************)
 (* isConvexSpace R T == interface for convex spaces                           *)
-(* ConvexSpace R == structure of convex space                                 *)
-(* a <| t |> b == convexity operator                                          *)
+(*     ConvexSpace R == structure of convex space                             *)
+(*       a <| t |> b == convexity operator                                    *)
 (* E : lmodType R with R : realDomainType and R : realDomainType are shown to *)
 (* be convex spaces                                                           *)
+(*                                                                            *)
 (******************************************************************************)
 
 Reserved Notation "x <| p |> y" (format "x  <| p |>  y", at level 49).
@@ -29,38 +30,41 @@ Local Open Scope ring_scope.
 
 Import numFieldNormedType.Exports.
 
-(* NB: recent addition to MathComp *)
-Lemma divrNN (R : unitRingType) (x y: R) : (- x) / (- y) = x / y.
-Proof. exact: divrNN. Qed.
-
 (* TODO: move *)
-Lemma factorE {R : numDomainType} (a b x : R) : factor a b x = (a - x) / (a - b).
+Lemma factorNN {R : numDomainType} (a b x : R) :
+  factor a b x = (a - x) / (a - b).
 Proof.
 by rewrite /factor -(opprB x a) -(opprB b a) invrN mulrNN.
 Qed.
 
-(* TODO: move *)
 Lemma factor1 {R : numFieldType} (a b x : R) : a != b ->
   factor a b x + factor b a x = 1.
 Proof.
 move=> ab.
-by rewrite factorE /factor -mulrDl addrA subrK divrr// unitfE subr_eq0.
+by rewrite factorNN /factor -mulrDl addrA subrK divrr// unitfE subr_eq0.
 Qed.
 
 (* TODO: move *)
-Lemma in_IntervalW d {T : porderType d} (a b c e : T):
-  (a <= b)%O -> (c <= e)%O -> forall x, x \in `[b, c] -> x \in `[a, e].
+Lemma subset_itvW d {T : porderType d} (a b c e : T) u v :
+    (a <= b)%O -> (c <= e)%O ->
+  `]b, c[ `<=` [set` Interval (BSide u a) (BSide v e)].
 Proof.
-move=> ab ce x; rewrite 2!in_itv /= => /andP[bx xc].
-by rewrite (le_trans ab)//= (le_trans _ ce).
+move=> ab ce; apply: (@subset_trans _ `]a, e[%classic).
+  move=> x/=; rewrite 2!in_itv/= => /andP[].
+  by move=> /(le_lt_trans ab) ->/= /lt_le_trans; exact.
+move: u v => [] [] /=; [exact: subset_itv_oo_co|exact: subset_itv_oo_cc|
+                        exact: subset_refl|exact: subset_itv_oo_oc].
 Qed.
 
-Lemma in_IntervalW' d {T : porderType d} (a b c e : T):
-  (a <= b)%O -> (c <= e)%O -> forall x, x \in `]b, c[ -> x \in `]a, e[.
-Proof.
-move=> ab ce x; rewrite 2!in_itv /= => /andP[bx xc].
-by rewrite (le_lt_trans ab)//= (lt_le_trans xc).
-Qed.
+(* TODO: move *)
+Lemma cvg_at_right_filter (R : numFieldType) (f : R -> R) (x : R) :
+  f z @[z --> x] --> f x -> f z @[z --> x^'+] --> f x.
+Proof. exact: (@cvg_within_filter _ _ _ (nbhs x)). Qed.
+
+(* TODO: move *)
+Lemma cvg_at_left_filter (R : numFieldType) (f : R -> R) (x : R) :
+  f z @[z --> x] --> f x -> f z @[z --> x^'-] --> f x.
+Proof. exact: (@cvg_within_filter _ _ _ (nbhs x)). Qed.
 
 Declare Scope convex_scope.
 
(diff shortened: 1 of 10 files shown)


theories/convex.v
```diff
@@ -0,0 +1,275 @@
...
+  avg p a (avg q b c) = avg (p%:inum * q%:inum)%:i01 (avg r a b) c.
+Proof. by move=> h; rewrite /avg (convA _ _ r). Qed.
+
+HB.instance Definition _ := @isConvexSpace.Build R R^o
+  (Choice.class _) _ avg0 avgI avgC avgA.
+
+End realDomainType_convex_space.
+
+Definition derivable_interval {R : numFieldType} (f : R -> R) (a b : R) :=
+  forall x, x \in `[a, b] -> derivable f x 1.
```
affeldt-aist, line 156: [NB: review comment by @zstone1 copied verbatim from PR #794]

This definition misbehaves at the endpoints. In particular, it requires `f` to be be "differentiable from the left at `a`". That makes this definition rely on values of `f` outside `[a,b]`. One big consequence is characteristic functions like ``if x \in `[a,b] then 1 else 0`` fails this definition, as its not derivable at `a` or `b`. For what it's worth, I had an analogous issue with `realfun.v`, see #752 for how I addressed it there. You'll need something slightly different here, though.

I've seen textbooks take these approaches:
1. Demanding "continuity up the endpoints" is one approach
```
forall x, x \in `]a, b[ -> derivable f x 1.
/\
cvg (at_right a f^`()) /\ cvg (at_right b f^`())
```
or some variation of this. However, this implies some continuity of the derivative, which is maybe not ideal.
2. We could change the derivative to use a different topology. One way to do that is add an analog to the ``{within `[a,b], continuous f}`` notation of  ``{within `[a,b], derivable f}``. This would take limits only on the subspace topology. 
3. Similarly, but specialized to working in `R`, it should be fine to define it like
```
forall x, x \in `]a, b[ -> derivable f x 1
/\
"differentiable from the right at b" /\ "differentiable from the left at a"
```
For some sensible notion of "left" and "right" derivative.

theories/convex.v
```diff
@@ -0,0 +1,275 @@
...
+  b >= b' -> derivable_interval f a b -> derivable_interval f a b'.
+Proof.
+rewrite/derivable_interval => b'b deriv x xab'; apply/deriv.
+exact: in_IntervalW xab'.
+Qed.
+
+(* ref: http://www.math.wisc.edu/~nagel/convexity.pdf *)
+Section twice_derivable_convex.
+Context {R : realType}.
+Variables (f : R -> R^o) (a b : R^o).
```
affeldt-aist, line 185: [NB: review comment by @zstone1 copied verbatim from PR #794 ]

Where do we need the assumption that the range of `f` is `R`? If it's painless, generalizing to a vector space would be quite helpful for me. I'll need some stuff about convexity in both `R` and `R^n` for my work on space-filling curves, so I'm happy to have these results either way.

affeldt-aist:
[NB: review comment by @zstone1 copied verbatim from PR https://github.com/math-comp/analysis/pull/794 ]

Nice results. I've made comments about a couple generalizations, one regarding annoying endpoint issues, the other regarding generalizing to vector space ranges. I care more about the endpoints one mostly because it's much harder to fix later. Happy to work with you on that if that'd be helpful.

hoheinzollern:
On top of being helpful for the exponential function, definitions about convexity are also necessary for the formalization of Hoelder's inequality and the special case of Cauchy Schwarz's inequality from PR #790.

theories/convex.v
affeldt-aist: >     1. Demanding "continuity up the endpoints" is one approach
> 
> 
> ```
> forall x, x \in `]a, b[ -> derivable f x 1.
> /\
> cvg (at_right a f^`()) /\ cvg (at_right b f^`())
> ```
> 
> or some variation of this. However, this implies some continuity of the derivative, which is maybe not ideal. 

The last two commits implement this solution.
(Hopefully, we think that you wrote ``` cvg (at_right a f^`()) ``` but actually meant `cvg (at_right a f)`.)
Is it in the right direction?

theories/convex.v
affeldt-aist: > Where do we need the assumption that the range of f is R?

We at least need to be able to compare to `0` but I maybe need to spend more time.

theories/convex.v
```diff
@@ -0,0 +1,317 @@
...
+Import numFieldTopology.Exports.
+
+Local Open Scope classical_set_scope.
+Local Open Scope ring_scope.
+
+Import numFieldNormedType.Exports.
+
+Lemma cvg_at_right_filter (R : numFieldType)
+  (K : numDomainType) (V : pseudoMetricNormedZmodType K)
+  (f : R -> V) (x : R) :
```
zstone1, line 40:
```suggestion
Lemma cvg_at_right_filter (R : numFieldType) (V : topologicalType)
  (f : R -> V) (x : R) :
```
We might as well generalize to topologicalType. These results are super useful for other setting (homotopy/homology).

theories/convex.v
```diff
@@ -0,0 +1,317 @@
...
+  f z @[z --> x] --> f x -> f z @[z --> x^'+] --> f x.
+Proof. exact: (@cvg_within_filter _ _ _ (nbhs x)). Qed.
+
+Lemma cvg_at_left_filter (R : numFieldType)
+  (K : numDomainType) (V : pseudoMetricNormedZmodType K)
+  (f : R -> V) (x : R) :
```
zstone1, line 46:
```suggestion
Lemma cvg_at_left_filter (R : numFieldType) (V : topologicalType)
  (f : R -> V) (x : R) :
```

theories/convex.v
```diff
@@ -0,0 +1,317 @@
...
+Lemma cvg_at_left_within (R : numFieldType)
+  (K : numDomainType) (V : pseudoMetricNormedZmodType K) (F : R -> V) (y : R) :
+  F x @[x --> y^'-] --> F y ->
+  F x @[x --> within (fun u => u <= y) (nbhs y)] --> F y.
+Proof.
+move=> Fyl; apply/cvgrPdist_lt => e e0.
+rewrite /at_left in Fyl; move/cvgrPdist_lt : Fyl => /(_ _ e0).
+rewrite !near_withinE; apply: filterS => x h.
+by rewrite le_eqVlt => /predU1P[->|//]; rewrite subrr normr0.
+Unshelve. all: by end_near. Qed.
```
zstone1, line 87:
```suggestion
Lemma cvg_at_right_within (R : numFieldType) (V : topologicalType)
  (F : R -> V) (x : R) :
  F x @[x --> x^'+] --> F x ->
  F x @[x --> within (fun u => x <= u) (nbhs x)] --> F x.
Proof.
move=> Fxr U Ux; rewrite ?near_simpl ?near_withinE; near=> z; rewrite le_eqVlt.
by move/predU1P => [<-|]; [exact: nbhs_singleton | near: z; exact: Fxr].
Unshelve. all: by end_near. Qed.

Lemma cvg_at_left_within (R : numFieldType) (V : topologicalType)
  (F : R -> V) (y : R) :
  F x @[x --> y^'-] --> F y ->
  F x @[x --> within (fun u => u <= y) (nbhs y)] --> F y.
Proof.
move=> Fxr U Ux; rewrite ?near_simpl ?near_withinE; near=> z; rewrite le_eqVlt.
by move/predU1P => [->|]; [exact: nbhs_singleton | near: z; exact: Fxr].
Unshelve. all: by end_near. Qed.
```
We can generalize this to `topologicalType` as well. The proofs previously used norm lemmas, is all.

theories/convex.v
```diff
@@ -0,0 +1,317 @@
...
+
+Lemma derivable_oo_within_cc_continuous {R : numFieldType} {V : normedModType R}
+  (F : R -> V) (x y : R):
+    {in `]x, y[, forall x, derivable F x 1} ->
```
zstone1, line 91: This condition of "derivable on the interior, and continuous up to the boundary" is a fairly common idiom. Maybe it deserves a dedicated definition.

zstone1 (approved):
Much improved from before.
- It looks like all the endpoint conditions all make sense. It should now work correctly for the indicator function on `[0,1]` 
- I generalized a few things to `topologicalType`. 
- Happy to see the generalization to `convex spaces`.

Other than the small generalizations, looks good to me.

affeldt-aist:
Many thanks for the review!

theories/exp.v
```diff
@@ -464,6 +468,18 @@ have /expR_total_gt1[y [H1y H2y H3y]] : 1 <= x^-1 by rewrite ltW // !invf_cp1.
 by exists (-y); rewrite expRN H3y invrK.
 Qed.
 
+Local Open Scope convex_scope.
+Lemma convex_expR (t : {i01 R}) (a b : R^o) : a <= b ->
+  expR (a <| t |> b) <= (expR a : R^o) <| t |> (expR b : R^o).
```
CohenCyril, line 473: @affeldt-aist the multiplications of such lemmas call for a notion of convex function (or convex function on a given domain)

theories/realfun.v
```diff
@@ -10,6 +10,10 @@ From HB Require Import structures.
 (******************************************************************************)
 (* This file provides properties of standard real-valued functions over real  *)
 (* numbers (e.g., the continuity of the inverse of a continuous function).    *)
+(*                                                                            *)
+(*   derivable_oo_continuous_bnd f x y == f is derivable on `]x, y[ and       *)
+(*                                        continuous up to the boundary       *)
```
CohenCyril, line 15: ``{in `]x, y[, forall x, derivable f x 1}`` does not mean ``f is derivable on `]x, y[ a`` :shrug:

PR #877, opened 2023-03-15
fubini for s-finite measures

affeldt-aist:

This PR contains only one lemma: Fubini for s-finite measure, it comes from the PR on kernels (PR #749 ).

##### Things done/to do

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -9,6 +9,8 @@
   + canonicals `ereal_blatticeType`, `ereal_tblatticeType`
 - in `lebesgue_measure.v`:
   + lemma `emeasurable_itv`
+- in `lebesgue_integral.v`:
+  + lemma `sfinite_Fubini`
 
 ### Changed
 
--- a/theories/lebesgue_integral.v
+++ b/theories/lebesgue_integral.v
@@ -4963,3 +4963,59 @@ Theorem Fubini :
 Proof. by rewrite fubini1 -fubini2. Qed.
 
 End fubini.
+
+Section sfinite_fubini.
+Local Open Scope ereal_scope.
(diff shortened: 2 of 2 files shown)


zstone1 (approved):
Proof makes sense. Personally, I like this style of equational reasoning. The explicit terms appearing in the transitivity calls make it clear how the sums and integrals are moving around. It's very much how a measure theory textbook would feel to read. 

Inspired by agda's equational reasoning stuff, I made a little notation to for formalize it in a different library. See https://github.com/zstone1/coq-complex/blob/master/src/helpers.v around line 170. It has the benefit of being able to `_` away unchanged terms. I'd love to have something like this in mathcomp.  Also it's generalized to any transitive reflexive relation, including `<` and `<=`.

affeldt-aist:
Thanks for taking a look. Indeed, the proof mimics a paper proof presented like equational reasoning, see page 16 of http://www.cs.ox.ac.uk/people/samuel.staton/papers/esop2017.pdf.

affeldt-aist:
> Inspired by agda's equational reasoning stuff, I made a little notation to for formalize it in a different library.
> See https://github.com/zstone1/coq-complex/blob/master/src/helpers.v around line 170.
> It has the benefit of being able to `_` away unchanged terms. I'd love to have something like this in mathcomp.
> Also it's generalized to any transitive reflexive relation, including `<` and `<=`.

Example of use:
https://github.com/zstone1/coq-complex/blob/9ee101eb5b34c30d2e136dd2c080613096c593a1/src/complex_core.v#L485-L495

Indeed, I believe that there are situations where an approach which highlights what's changing improves maintainability despite the extra verbosity.

affeldt-aist:
> I'd love to have something like this in mathcomp.

Do you want to record this as a wish issue before merging?

PR #894, opened 2023-04-11
ae as a filter

### LINKED ISSUE #882: turn ae into a filter
https://github.com/math-comp/analysis/blob/dd37ba7446694c09df57843bbcf0d2f202f0c906/theories/measure.v#L3025-L3029

"let's make ae a filter and use filter lemmas rather than duplicating the theory of filters" (CC)

affeldt-aist:

fixes #882 

This requires a bit more work: the definition of the notation is ad hoc, the use filter seems to cause a specialization from semiring of sets to ring of sets, the type class hint is effective in `measure.v` but not yet in `lebesgue_integral.v`.

fyi @CohenCyril 

##### Things done/to do

  (only append to minimize problems when merging/rebasing)
  (you can consider the use of `etc/changes.sh` to generate the changelog)


theories/measure.v
```diff
@@ -3015,46 +3016,93 @@ Lemma measure_negligible (mu : {content set T -> \bar R}) (A : set T) :
...
+#[global]
+Instance ae_filter_algebraOfSetsType d {T : algebraOfSetsType d}
+    (R : realFieldType) (mu : {measure set T -> \bar R}) :
+  Filter (almost_everywhere mu).
+Proof. exact: ae_filter_ringOfSetsType. Qed.
+
+#[global]
+Instance ae_filter_measurableType d {T : measurableType d}
+    (R : realFieldType) (mu : {measure set T -> \bar R}) :
+  Filter (almost_everywhere mu).
```
zstone1, line 3093: It would be nice to have separate `ProperFilter` instances, that way we can use the `near` notation comfortably. This will require `mu [set: T] > 0`.

theories/measure.v
```diff
@@ -3015,46 +3016,93 @@ Lemma measure_negligible (mu : {content set T -> \bar R}) (A : set T) :
...
+#[global]
+Instance ae_filter_ringOfSetsType d {T : ringOfSetsType d} (R : realFieldType)
+  (mu : {measure set T -> \bar R}) : Filter (almost_everywhere mu).
+Proof.
+by split; [exact: almost_everywhereT|exact: almost_everywhereI|
+  exact: almost_everywhereS].
+Qed.
+
+#[global]
+Instance ae_filter_algebraOfSetsType d {T : algebraOfSetsType d}
```
zstone1, line 3085: I guess I'm surprised that we need a separate instance for `algebraOfSetsType`. I also don't really understand typeclass resolution in Coq, though. So if this is necessary its fine.

zstone1 (approved):
Looks good to me. A small question, and an optional task to include `ProperFilter` instances. Did you figure out the notation problem you mentioned?

affeldt-aist:
> Looks good to me. A small question, and an optional task to include `ProperFilter` instances. Did you figure out the notation problem you mentioned?

Not really. I tried to do with only one definition instead of the two definitions `almost_everywhere` and `almost_everywhere_notation` but then I failed to have the type classes working correctly. On the other hand, with the current solution, the notation is stable, so maybe it is fine.

PR #896, opened 2023-04-12
a hierarchy and a theory of kernels

affeldt-aist:

This is the "mathematical part" of PR #749 : the hierarchy of kernels and their theory (stability of s-finite kernels by composition, etc.). This is documented in https://hal.inria.fr/hal-03917948/ (which is a reviewed paper, that might
simplify the review work for this PR).

The "programming language part" of PR #749 is now in the draft PR #912 because it is still the topic of
discussion. Some parts of it might integrate master later though (e.g., kernel for scoring or for conditional branching).

(PR #749 is therefore scheduled for closing.)

@CohenCyril @AyumuSaito

##### Things done/to do

### DIFF
--- a/theories/kernel.v
+++ b/theories/kernel.v
@@ -197,7 +197,7 @@ Arguments measure_uub {_ _ _ _ _} _.
 HB.factory Record Kernel_isFinite d d'
     (X : measurableType d) (Y : measurableType d') (R : realType)
     (k : X -> {measure set Y -> \bar R}) of isKernel _ _ _ _ _ k := {
-  __factory__measure_uub : measure_fam_uub k }.
+  measure_uub : measure_fam_uub k }.
 
 Section kzero.
 Context d d' (X : measurableType d) (Y : measurableType d') (R : realType).
@@ -226,7 +226,7 @@ Let sfinite_finite :
 Proof.
 exists (fun n => if n is O then [the _.-ker _ ~> _ of k] else
   [the _.-ker _ ~> _ of @kzero _ _ X Y R]).
-  by case => [|_]; [exact: __factory__measure_uub|exact: kzero_uub].
+  by case => [|_]; [exact: measure_uub|exact: kzero_uub].
 move=> t U mU/=; rewrite /mseries.
 rewrite (nneseries_split 1%N)// big_ord_recl/= big_ord0 adde0.
 rewrite ereal_series (@eq_eseriesr _ _ (fun=> 0%E)); last by case.
@@ -237,7 +237,7 @@ HB.instance Definition _ :=
   @Kernel_isSFinite_subdef.Build d d' X Y R k sfinite_finite.
 
 HB.instance Definition _ :=
-  @SFiniteKernel_isFinite.Build  d d' X Y R k __factory__measure_uub.
+  @SFiniteKernel_isFinite.Build  d d' X Y R k measure_uub.
 
 HB.end.
 
@@ -289,7 +289,7 @@ HB.instance Definition _
 HB.factory Record Kernel_isSFinite d d'
     (X : measurableType d) (Y : measurableType d') (R : realType)
     (k : X -> {measure set Y -> \bar R}) of isKernel _ _ _ _ _ k := {
-  __factory__sfinite : exists s : (R.-fker X ~> Y)^nat,
+  sfinite : exists s : (R.-fker X ~> Y)^nat,
     forall x U, measurable U -> k x U = kseries s x U }.
 
 HB.builders Context d d' (X : measurableType d) (Y : measurableType d')
@@ -297,7 +297,7 @@ HB.builders Context d d' (X : measurableType d) (Y : measurableType d')
 
 Lemma sfinite_subdef : Kernel_isSFinite_subdef d d' X Y R k.
 Proof.
-split; have [s sE] := __factory__sfinite; exists s => //.
+split; have [s sE] := sfinite; exists s => //.
 by move=> n; exact: measure_uub.
 Qed.
 
@@ -320,7 +320,7 @@ Notation "R .-spker X ~> Y" := (sprobability_kernel X Y R).
 HB.factory Record Kernel_isSubProbability d d'
     (X : measurableType d) (Y : measurableType d') (R : realType)
     (k : X -> {measure set Y -> \bar R}) of isKernel _ _ X Y R k := {
-  __factory__sprob_kernel : ereal_sup [set k x [set: Y] | x in [set: X]] <= 1 }.
+  sprob_kernel : ereal_sup [set k x [set: Y] | x in [set: X]] <= 1 }.
 
 HB.builders Context d d' (X : measurableType d) (Y : measurableType d')
   (R : realType) k of Kernel_isSubProbability d d' X Y R k.
@@ -328,13 +328,13 @@ HB.builders Context d d' (X : measurableType d) (Y : measurableType d')
 Let finite : @Kernel_isFinite d d' X Y R k.
 Proof.
 split; exists 2%R => /= ?; rewrite (@le_lt_trans _ _ 1%:E) ?lte_fin ?ltr1n//.
-by rewrite (le_trans _ __factory__sprob_kernel)//; exact: ereal_sup_ub.
+by rewrite (le_trans _ sprob_kernel)//; exact: ereal_sup_ub.
 Qed.
 
 HB.instance Definition _ := finite.
 
 HB.instance Definition _ :=
-  @FiniteKernel_isSubProbability.Build _ _ _ _ _ k __factory__sprob_kernel.
+  @FiniteKernel_isSubProbability.Build _ _ _ _ _ k sprob_kernel.
 
 HB.end.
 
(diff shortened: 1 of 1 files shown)


theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+suff r_gt0 : (0 < r)%R by exists (PosNum r_gt0).
+by rewrite -lte_fin; apply: (le_lt_trans _ (kr point)).
+Qed.
+
+End measure_fam_uub.
+
+HB.mixin Record Kernel_isSFinite_subdef
+    d d' (X : measurableType d) (Y : measurableType d')
+    (R : realType) (k : X -> {measure set Y -> \bar R}) := {
+  sfinite_subdef : exists2 s : (R.-ker X ~> Y)^nat,
```
zstone1, line 163: Is it better to have 3 fields or have this single exisential? That is `s : R.-ker X ~> Y ^ nat`, `s_uub : forall n, measure_fam_uub (s n)` and `s_kseriesE : forall x U, measurable U -> k x U = kseries s x U`.

We talked in the mathcomp meeting about how this kind of existential qualifier is great in factories to enforce "you don't get to control the choice of `s`. But in the mixin, wouldn't it be better to have `s` exposed to avoid the inevitable `cid2` (which I see yo use later)

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+    (R : realType) (k1 k2 : R.-sfker T ~> T') :
+  (forall x U, k1 x U = k2 x U) -> k1 = k2.
+Proof.
+move: k1 k2 => [m1 [[?] [?]]] [m2 [[?] [?]]] /= k12.
+have ? : m1 = m2.
+  by apply/funext => t; apply/eq_measure; apply/funext => U; rewrite k12.
+by subst m1; f_equal; f_equal; f_equal; apply/Prop_irrelevance.
+Qed.
+
+HB.mixin Record SFiniteKernel_isFinite
```
zstone1, line 186: Something is a bit counter-intuitive with this bunch of structures. The `SFiniteKernel_isFinite `appears to work for any kind of kernel, not just `SFinite` ones. `Kernel_isFinite` factory also works for any kind of kernel, but produces a `SFiniteKernel_isFinite`. I've seen this idiom of having mixins and factories with the same fields, and it always surprises me.

I would rather see `SFiniteKernel_isFinite` actually require `k` to be an `SFiniteKernel` if that is its intended use.

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+HB.factory Record Kernel_isFinite d d' (X : measurableType d)
+    (Y : measurableType d') (R : realType) (k : X -> {measure set Y -> \bar R})
+    of isKernel _ _ _ _ _ k := {
+  measure_uub : measure_fam_uub k }.
+
+Section kzero.
+Context d d' (X : measurableType d) (Y : measurableType d') (R : realType).
+
+Definition kzero : X -> {measure set Y -> \bar R} :=
+  fun _ : X => [the measure _ _ of mzero].
```
zstone1, line 210:
```suggestion
  fun _ : X => mzero.
```

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+    d d' (X : measurableType d) (Y : measurableType d')
+    (R : realType) :=
+  {k of @FiniteKernel _ _ _ _ _ k &
+        FiniteKernel_isSubProbability _ _ X Y R k }.
+Notation "R .-spker X ~> Y" := (sprobability_kernel X Y R).
+
+HB.factory Record Kernel_isSubProbability
+    d d' (X : measurableType d) (Y : measurableType d')
+    (R : realType) (k : X -> {measure set Y -> \bar R}) of isKernel _ _ X Y R k
+  := { sprob_kernel : ereal_sup [set k x [set: Y] | x in setT] <= 1}.
```
zstone1, line 329: Same thing here with repeated fields between mixin/factory. This is also minor a maintainability issue to have the properties unfactored like this.

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+Section measurable_fun_xsection_finite_kernel.
+Context d d' (X : measurableType d) (Y : measurableType d') (R : realType).
+Variable k : R.-fker X ~> Y.
+Implicit Types A : set (X * Y).
+
+Let phi A := fun x => k x (xsection A x).
+Let XY := [set A | measurable A /\ measurable_fun setT (phi A)].
+
+Lemma measurable_fun_xsection_finite_kernel A :
+  A \in measurable -> measurable_fun setT (phi A).
```
zstone1, line 457: Why `A \in measurable` here?

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+Proof.
+move=> h.
+rewrite (_ : (fun x => _) =
+    (fun x => lim_esup (fun n => \int[l x]_y (k_ n (x, y))%:E))); last first.
+  apply/funext => x.
+  transitivity (lim (fun n => \int[l x]_y (k_ n (x, y))%:E)); last first.
+    rewrite is_cvg_lim_esupE//.
+    apply: ereal_nondecreasing_is_cvg => m n mn.
+    apply: ge0_le_integral => //.
+    - by move=> y _; rewrite lee_fin.
```
zstone1, line 491: Another proof with a ton of trivial measurability conditions. There is a `isMeasurableFun` canonical structure, but it seems like a lot of work to actually convince Coq to infer these goals.

theories/kernel.v
```diff
@@ -0,0 +1,1169 @@
...
+Arguments measurable_fun_integral_finite_kernel {_ _ _ _ _} k l.
+Arguments measurable_fun_integral_sfinite_kernel {_ _ _ _ _} k l.
+
+Section kdirac.
+Context d d' (X : measurableType d) (Y : measurableType d') (R : realType).
+Variable f : X -> Y.
+
+Definition kdirac (mf : measurable_fun setT f)
+    : X -> {measure set Y -> \bar R} :=
+  fun x => [the measure _ _ of dirac (f x)].
```
zstone1, line 573: Same as above with `mzero`. There are many more below, that can be cleaned up as well.

zstone1 (commented):
Looks pretty good. The comment about an existential vs. 3 fields is the most interesting. Otherwise I don't anything too bad. I find the identical mixin + factory idiom to be a bit strange, but it's fairly common, and I can live with it. 

In some sense, an empty factory with an extra condition is what you is maybe what you mean? I tried to do that with
```
HB.factory Record Kernel_isFinite d d' (X : measurableType d)
    (Y : measurableType d') (R : realType) (k : X -> {measure set Y -> \bar R})
    of isKernel _ _ _ _ _ k & SFiniteKernel_isFinite _ _ _ _ _ k := {
  measure_uub : measure_fam_uub k }.
``` 
However, I get 
>Error: I could not find classes covering exactly mixins: 
[indt «SFiniteKernel_isFinite.axioms_», indt «isKernel.axioms_»] 
In particular the covering [indt «FiniteKernel.axioms_»] also includes mixins: 
[indt «Kernel_isSFinite_subdef.axioms_»] 
This should never happen, please report a bug.

so that maybe doesn't work quite right.

theories/kernel.v
affeldt-aist: These annotations are really here for compatibility with Coq 8.15. We can maybe keep them for a while and create an issue to remove them once we drop the support for Coq 8.15.

theories/kernel.v
affeldt-aist: Doesn't turning this mixin into a three fields mixin make `eq_sfkernel` unprovable? (We use this lemma in PR #912 , yet that does not mean it is very unavoidable.)

theories/kernel.v
affeldt-aist: Same answer as above (compatibility with Coq 8.15). But I'll issue to remove them all once support for Coq 8.15 is dropped.

theories/kernel.v
affeldt-aist: Same answer as above (compatibility with Coq 8.15). But I'll issue to remove them all once support for Coq 8.15 is dropped.

theories/kernel.v
affeldt-aist: What about adding a few more Hints and use `auto` to get rid of trivial measurability conditions on functions? (as done in this commit https://github.com/math-comp/analysis/pull/896/commits/cca14fcda6196b8c37408c5146df4bbe55c66dfb)

theories/kernel.v
zstone1: Ah, fascinating. A genuine subtlety. I guess we want the default equality to mean "equality of the underlying function". So you're right, we do have to do it this way.

theories/kernel.v
affeldt-aist: I am investigating that. However, as a first observation,
adding an `SFiniteKernel` restriction to the `SFiniteKernel_isFinite mixin`
(or a `isKernel` restriction and a `Kernel_isSFinite_subdef` restriction,
it seems to be equivalent), causes an error when building the finite kernel
instance of `kadd`. Here is the error message to give an idea:
```
The 9th term has type
 "SFiniteKernel_isFinite.axioms_ d d' X Y R (kadd k1 k2)
    (Builders_254.HB_unnamed_factory_256 HB_unnamed_factory_351)
    (HB_unnamed_factory_345 k1 k2)"
which should be coercible to
 "SFiniteKernel_isFinite.axioms_ d d' X Y R (kadd k1 k2)
    (HB_unnamed_factory_348 k1 k2) (HB_unnamed_factory_345 k1 k2)".
```
I'll try to report more ASAP.

theories/kernel.v
affeldt-aist: Do you think that the use of `auto` in the last commit is reasonable?

affeldt-aist: I will revert the use of auto introduced by the last commit
(it is maybe not the way we want those proofs to be automated according to conversations with @proux01 and @CohenCyril ).
On the other hand, there is another PR in progress
https://github.com/math-comp/analysis/issues/916
and this issue
https://github.com/math-comp/analysis/pull/925
about measurable_fun that provide more occasions to rework about
`measurable_fun` proofs to hopefully make them a bit shorter.

theories/kernel.v
affeldt-aist: For the repeated fields, what about having (as in the last commit) the field of the factory renamed as `__factory__field` as an indication that there is another field that actually already captures the same property?

theories/kernel.v
affeldt-aist: In fact, I have nothing more to report regarding this point. I'd be happy to merge it like that and issue to remember about your point.

theories/kernel.v
zstone1: A fair question. My thought is more about the duplicated predicate, than the duplicated name So a new name doesn't change much for me. I won't block the PR on this issue, though. Sounds like there no obvious resolution, which is what I expected.

zstone1 (approved):
I'm fine with things as they are now. Most of my issues are longer term ones about factory and mixin duplication, and R vs \bar R. They're not really complaints about this PR in particular, and I'll be happy to have this kernel stuff in master.

PR #925, opened 2023-05-10
Refactoring lemmas about composition in Lebesgue Measure

marcomole00:

I've changed the formulation of some lemmas in the lebesgue measure theory.

The lemmas state that the measurability of functions in R is preserved when composed with 
- multiplication with a scalar
- negation
- exponentiation with a natural number

##### Things Done

*Lemma measurable_funrM* is now 
 ``` 
 Lemma measurable_funrM (D: set R) (k: R) : measurable_fun D ( *%R k).
```
instead of
```
Lemma measurable_funrM D f (k : R) : measurable_fun D f ->
measurable_fun D (fun x => k * f x).
```

*Lemma measurable_funN* is now 
```
Lemma measurable_funN (D: set R) f : measurable_fun D (-%R ).
```

instead of 
```
Lemma measurable_funN D f : measurable_fun D f -> measurable_fun D (-%R \o f).
```

 *Lemma measurable_fun_exprn* is now 
```
Lemma measurable_fun_exprn (D:set R) n : measurable_fun D ( (@GRing.exp R)^~ n) .
```

instead of 
```
Lemma measurable_fun_exprn D n f :
measurable_fun D f -> measurable_fun D (fun x => f x ^+ n).
```

*Lemma measurable_fun_sqr* is now
```
Lemma measurable_fun_sqr (D: set R) f : measurable_fun D ( ( (@GRing.exp R)^~ 2%N)).
```
instead of
```
Lemma measurable_fun_sqr D f :

measurable_fun D f -> measurable_fun D (fun x => f x ^+ 2).
```

### CODE REVIEW

**Comment by CohenCyril** [comment]
I would name the non compositional lemmas `operator_mesurable` (e.g. `mulr_measurable`), and prove the other (former) lemmas from this using composition lemmas (unary and binary).

**Comment by affeldt-aist** [comment]
Sorry, I failed to see you message before my nitpicking commit where I was essentially doing this renaming.

**Review by affeldt-aist — APPROVED** [review]
Thanks for your contribution Marco. You are probably right: the lemmas you pointed out ought better be provided in smaller lemmas to be composed with `measurable_fun_comp`. I just added a commit with a changelog and nitpicking. See CONTRIBUTING.md for a few clues about style.
Naming is not yet perfect but we'll have the occasion to come back to it when addressing https://github.com/math-comp/analysis/issues/916.

File: theories/lebesgue_measure.v [hunk 1]
```diff
@@ -1823,6 +1896,7 @@ move=> _ [_ [x ->] <-];rewrite esups_preimage setI_bigcupr.
 by apply: bigcup_measurable => ? ?; exact/mf/emeasurable_itv.
 Qed.
 
+(*can this be changed?*)
```
affeldt-aist: This can indeed be changed advantageously by using `measurable_fun_if` provided lemmas about the measurability of boolean comparison operators (_ < _, etc.). This approach should trigger a bit of cleaning also and might better be done as a different PR.

PR #939, opened 2023-06-01
minor fixes and additions to `exp.v`

affeldt-aist:

These fixes and lemmas are used in a proof of Hoelder's inequality to be PRed soon.

@hoheinzollern 

##### Things done/to do

### DIFF
--- a/theories/exp.v
+++ b/theories/exp.v
@@ -610,6 +610,13 @@ Proof. by rewrite /power_pos; case: ifPn => // _; exact: expR_ge0. Qed.
 Lemma power_pos_gt0 a x : 0 < a -> 0 < a `^ x.
 Proof. by move=> a0; rewrite /power_pos gt_eqF// expR_gt0. Qed.
 
+Lemma gt0_power_pos a x : 0 < x -> 0 <= a -> 0 < a `^ x -> 0 < a.
+Proof.
+move=> x0 a0; rewrite /power_pos; case: ifPn => [_|a_neq0 _].
+  by rewrite gt_eqF//= ltxx.
+by rewrite lt_neqAle a0 andbT eq_sym.
+Qed.
+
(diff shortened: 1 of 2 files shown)


theories/exp.v
```diff
@@ -604,20 +610,24 @@ Proof. by rewrite /power_pos; case: ifPn => // _; exact: expR_ge0. Qed.
...
-move=> a0; rewrite /power_pos; case: ifPn => [/eqP->|a0'].
-  by rewrite oner_eq0.
-by rewrite mul1r lnK// posrE lt_neqAle eq_sym a0'.
+move=> x0 a0; rewrite /power_pos; case: ifPn => [_|a_neq0 _].
+  by rewrite gt_eqF//= ltxx.
+by rewrite lt_neqAle a0 andbT eq_sym.
 Qed.
 
-Lemma power_posr0 a : a `^ 0 = 1.
+Lemma power_pos0 x : 0 `^ x = (x == 0)%:R.
```
zstone1, line 620: Everything about this statement is uncomfortable. The fact that we have a theorem that explicitly relies on `0 ^ 0` behavior is awkward. And this uses the most surprising coercion, `nat_of_bool`. I see that this lemma gets used a bunch. But would `x != 0 -> 0^x = 0` work in its place?

zstone1 (changes requested):
Just one confusing lemma statement. If adding a `x!=0` condition breaks too many things, at least making the coercion explicit will make it more readable for me.

theories/exp.v
affeldt-aist: Indeed, it looks a bit too ssreflectish. Changing to the format you propose reduces the size of scripts a bit (see the last commit) so if moreover the statement is more readable this is a goof change.

zstone1 (approved):

PR #942, opened 2023-06-03
Hoelder's inequality

affeldt-aist:

~~based on PR #969~~ (merged)

~~based on PR #990~~ (merged,  but the use of `^o` is still in question)

This PR is already used in the formalization of Minkowski's inequality and might be used in PR #995.

~~TODO: issue about how to extend Lnorm to handle p = +oo (change Lnorm from R to \bar R?)~~

~~TODO: issue about the introduction of a type measureType~~

@hoheinzollern 

##### Things done/to do


CohenCyril:
Why is `power_pos` called like this in the first place ?

t6s:
> Ultimately it should be subsumed by a powC function that uses the principal value of the complex log...

What is the blocking issue towards that canonical definition?  Do we need a theory for analytic continuations, or do we even need Riemann surfaces?

theories/probability.v
```diff
@@ -1,9 +1,10 @@
 (* mathcomp analysis (c) 2022 Inria and AIST. License: CeCILL-C.              *)
 From mathcomp Require Import all_ssreflect.
 From mathcomp Require Import ssralg poly ssrnum ssrint interval finmap.
-Require Import mathcomp_extra boolp reals ereal.
+From mathcomp.classical Require Import mathcomp_extra boolp classical_sets.
+From mathcomp.classical Require Import functions cardinality.
```
CohenCyril, line 5: Please use the namespace `mathcomp` rather than the too specific `mathcomp.classical`.
If this triggers an error, it means two files in mathcomp are named the same, which is bad.

affeldt-aist: This is addressed in PR #969

affeldt-aist:
> ~Why is `power_pos` called like this in the first place ?~
> 
> Nevermind, I understand why... but shouldn't we call it `powR` instead? Indeed, on negative bases it would almost give restults in C∖R, except for integer exponents, so it make sense calling it with suffix `R`.

The renaming is done in PR #969

zstone1 (commented):
One question about the notation. Otherwise the structure looks good.

theories/lebesgue_integral.v
```diff
@@ -5344,3 +5353,238 @@ Qed.
...
+
+Section L_norm.
+Context d (T : measurableType d) (R : realType)
+  (mu : {measure set T -> \bar R}).
+Local Open Scope ereal_scope.
+
+Definition L_norm (p : R) (f : T -> R) : \bar R :=
+  (\int[mu]_x (`|f x| `^ p)%:E) `^ p^-1.
+
+Local Notation "'N_ p [ f ]" := (L_norm p f).
```
CohenCyril, line 5365:
```suggestion
Local Notation "'N_ p [ f ]" := (L_norm p f).
```
Could you try using `` `| f |_ p `` instead?

theories/lebesgue_integral.v
```diff
@@ -5344,3 +5353,238 @@ Qed.
...
+Proof. by rewrite /L_norm poweR_ge0. Qed.
+
+Lemma eq_L_norm (p : R) (f g : T -> R) : f =1 g -> 'N_p[f] = 'N_p[g].
+Proof. by move=> fg; congr L_norm; exact/funext. Qed.
+
+End L_norm.
+#[global]
+Hint Extern 0 (0 <= L_norm _ _ _) => solve [apply: L_norm_ge0] : core.
+
+Notation "'N[ mu ]_ p [ f ]" := (L_norm mu p f).
```
CohenCyril, line 5377: This seems to imply we really need a `measureType`.

theories/lebesgue_integral.v
zstone1: I definitely want this notation (when the measure is clear as it is here). But we _must_ put it in an appropriate local/scope/module, because it conflicts with the p-adic norm.

theories/lebesgue_integral.v
```diff
@@ -4,7 +4,7 @@ From mathcomp Require Import all_ssreflect ssralg ssrnum ssrint interval finmap.
 From mathcomp Require Import mathcomp_extra boolp classical_sets functions.
 From mathcomp Require Import cardinality fsbigop .
 Require Import signed reals ereal topology normedtype sequences real_interval.
-Require Import esum measure lebesgue_measure numfun.
+Require Import esum measure lebesgue_measure numfun exp itv.
```
affeldt-aist, line 7: Maybe this is `exp.v` that should depend on `lebesgue_integral.v` (see the proof of convexity of `powR`).

theories/lebesgue_integral.v
affeldt-aist: It seems reasonable to use `` `| _ |_ _ `` locally, but in general in conflicts with tactics such as `` => [|_]`` which is something that we sometimes do write. So, let's replace ``'N_p[f]`` with `` `|f|_p `` with the latter only declared as `Reserved` for local use.  ``'N[mu]_p[f]`` stays available otherwise.

theories/lebesgue_integral.v
affeldt-aist: Indeed...

theories/lebesgue_integral.v
```diff
@@ -5344,3 +5353,238 @@ Qed.
...
+    have -> : (p^-1 = 1 - q^-1)%R by rewrite -pq addrK.
+    have K : (q^-1 \in `[0, 1])%R.
+      by rewrite in_itv/= invr_ge0 (ltW q0)/= -pq ler_paddl// invr_ge0 ltW.
+    exact: (convex_expR (@Itv.mk _ `[0, 1] q^-1 K)%R).
+  rewrite expRD (mulrC _ (s x)) normed_expR ?gt_eqF// -/(F x).
+  rewrite (mulrC _ (t x)) normed_expR ?gt_eqF// -/(G x) => /le_trans; apply.
+  rewrite /s /t [X in (_ * X + _)%R](@lnK _ (F x `^ p)%R); last first.
+    by rewrite posrE powR_gt0.
+  rewrite (@lnK _ (G x `^ q)%R); last by rewrite posrE powR_gt0.
+  by rewrite mulrC (mulrC _ q^-1).
```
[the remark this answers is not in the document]
affeldt-aist: Done as PR #990 on which this PR is now based.

theories/lebesgue_integral.v
```diff
@@ -5344,3 +5353,238 @@ Qed.
 
 End sfinite_fubini.
 Arguments sfinite_Fubini {d d' X Y R} m1 m2 f.
+
+Section L_norm.
+Context d (T : measurableType d) (R : realType)
+  (mu : {measure set T -> \bar R}).
+Local Open Scope ereal_scope.
+
+Definition L_norm (p : R) (f : T -> R) : \bar R :=
```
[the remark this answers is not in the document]
affeldt-aist: We haven't yet decided anything about this comment. Is it blocking?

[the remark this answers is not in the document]
zstone1: As long as we're fine to maybe change `R` to `\bar R` later. Technically it would be a backwards incompatibility, but a very minor one.

theories/lebesgue_integral.v
affeldt-aist: Oops, it seems that the mere fact of doing a `Reserved` for the notation containing `|_` causes tactics such as `` => [|_]`` to fail... Need to investigate.

affeldt-aist: Of course `` `| _ |~ _`` goes through but this is not what we want. `` `|| _ ||_ _`` goes through but this is only because we do not happen to use tactics such as `=> [||_]`.

affeldt-aist: (I have the vague memory that we already ran into that problem.)

affeldt-aist: Shall we stick to the the `N[mu]_p[f]` / `N_p[f]` notation?

CohenCyril: Let's go for the `` `N[mu]_p [ f ]`` notation then...

theories/lebesgue_integral.v
affeldt-aist: Not being able to decide whether to put it in `exp.v`, we introduce a new file `hoelder.v` for results that depend on `lebesgue_integral.v` and `exp.v` (it will soon welcome the convexity of powR and Minkowski's inequality).

theories/lebesgue_integral.v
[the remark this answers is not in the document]
affeldt-aist: We have generalized the definition with the notion of essential supremum so that it handles +oo now. This certainly to be instrumented more with lemmas but we will certainly have an opportunity to go back to it in the near future with the Minkowski PR, Hardy-Littlewood and others.

zstone1 (approved):
Great. I'll mention here that the 0 case is invalid right now. I think it's just `mu (f^-1 (setT \ {0}))` but that can be easily added later without changing any lemma definitions.

But this looks good to me.

PR #1008, opened 2023-08-09
charge factory

IshiguroYoshihiro:
Define factory corresponding to "textbook definition" of charges like measures.

##### Things done/to do


IshiguroYoshihiro:
@t6s Reviewをお願いできますか？

t6s:
PR commentとcharge. vのheaderに説明を追加してください

affeldt-aist:
Maybe you can test this factory by replacing existing applications of `isCharge.Build`.

t6s:
I still do not understand the motivation. Is this the standard, textbook definition? Then, the name isCharge0 is weird.

CohenCyril:
The current `isCharge` is misnamed, it should rather be called : `isSemiSigmaAdditive`, and the rightful name of `isCharge0` is `isCharge`.
(Indeed when one needs a syntactic `'` or `0` in the name to disambiguate between two notions, it most certainly means one of them is illnamed.)

IshiguroYoshihiro:
> Maybe you can test this factory by replacing existing applications of `isCharge.Build`.

I tested  new `isCharge.Build`.
It is no longer necessary to prove `semi_additive` for instances of `isAdditiveCharge`.
Unlike using `isSemiSigmaAdditive.Build` (old `isCharge.Build`), R gets specialized from `numFieldType` to `realFieldType`.
This seems to be due to the fact that the Lemma semi_sigma_additive_is_additive (in `measure.v`) is proven with realFieldType.

affeldt-aist:
> This seems to be due to the fact that the Lemma `semi_sigma_additive_is_additive` (in measure.v) is proven with `realFieldType`.

This is maybe related to this comment: https://github.com/math-comp/analysis/blob/master/theories/ereal.v#L758

This suggests that we have already been trying to get rid of this specialization.

theories/charge.v
```diff
@@ -26,6 +26,10 @@ Require Import esum measure realfun lebesgue_measure lebesgue_integral.
 (*  measure_of_charge nu nu0 == measure corresponding to the charge nu, nu0   *)
 (*                              is a proof that nu is non-negative            *)
 (*                                                                            *)
+(* * Structures for functions on classes of sets                              *)
+(*                 isCharge0 == factory corresponding to the "textbook        *)
```
affeldt-aist, line 30: There is no `isCharge0` anymore.

theories/charge.v
```diff
@@ -85,16 +89,49 @@ Notation "{ 'additive_charge' 'set' T '->' '\bar' R }" :=
...
-HB.mixin Record isCharge d (T : semiRingOfSetsType d) (R : numFieldType)
+HB.mixin Record isSemiSigmaAdditive d (T : semiRingOfSetsType d) (R : numFieldType)
     (mu : set T -> \bar R) := {
   charge_semi_sigma_additive : semi_sigma_additive mu }.
 
 #[short(type=charge)]
-HB.structure Definition Charge d (T : semiRingOfSetsType d) (R : numFieldType)
-  := { mu of isCharge d T R mu & AdditiveCharge d mu }.
+HB.structure Definition AdditiveCharge_SemiSigmaAdditive_isCharge d (T : semiRingOfSetsType d) (R : numFieldType)
+  := { mu of isSemiSigmaAdditive d T R mu & AdditiveCharge d mu }.
```
CohenCyril, line 98: Structures do not have a `_is*` name, I guess this is still a `Charge` structure.

CohenCyril:
> I'll appreciate a little help. We have a long-standing comment about this section https://github.com/math-comp/analysis/blob/master/theories/ereal.v#L755 being potentially generalizable to `R : numFieldType`. The first lemma (`ereal_nbhs_singleton`) obviously is (maybe it should be clearly marked as such) but I don't see how to generalize the second one:
> 
> ```coq
> Lemma ereal_nbhs_nbhs (p : \bar R) (A : set (\bar R)) :
>   ereal_nbhs p A -> ereal_nbhs p (ereal_nbhs^~ A).
> ```
> 
> I may be missing something obvious, any clue?

Unfortunately, it is wrong. The `p = +oo%R` case is equivalent to:
`\forall k \near +oo%R, A k%:E -> \forall x \near nbhs +oo%R, \near x, A x%:E`
which is true for a `realFieldType` but false for `numFieldType` in general.

affeldt-aist:
Great! (in a sense...)
If you have the working counter-example right now in front of you, we should maybe insert it in the code, it is an interesting piece of documentation, isn't it?

affeldt-aist:
> > Maybe you can test this factory by replacing existing applications of `isCharge.Build`.
> 
> I tested new `isCharge.Build`. It is no longer necessary to prove semi-additivity for `isAdditiveCharge`. Unlike using `isSemiSigmaAdditive.Build` (old `isCharge.Build`), R gets specialized from `numFieldType` to `realFieldType`. This seems to be due to the fact that the Lemma `semi_sigma_additive_is_additive` (in measure.v) is proven with `realFieldType`.

This is not a problem because in the end the lemmas and definitions provided by charge.v did not get specialized.

affeldt-aist (approved):
This PR simplifies charge.v and incidentally clarifies leftover comments about the topology of \bar R, this is therefore a nice improvement.

PR #1065, opened 2023-10-19
Lebesgue differentiation theorem and applications

affeldt-aist:

This is track D of issue #965 

~~This is based of PR #995~~ (merged)

~~Based on PR #1121~~ (merged)

##### Things done/to do

### DIFF
--- a/theories/lebesgue_integral.v
+++ b/theories/lebesgue_integral.v
@@ -5839,12 +5839,12 @@ Qed.
 Lemma locally_integrable_indic D (A : set R) :
   open D -> measurable A -> locally_integrable D \1_A.
 Proof.
-move=> oU; split => // K _ cK.
+move=> oU mA; split => // K KD_ cK.
 apply: (@le_lt_trans _ _ (\int[mu]_(x in K) cst 1 x)).
   apply: ge0_le_integral => //=; first exact: compact_measurable.
   - by do 2 apply: measurableT_comp => //.
-  - move=> y _.
-    by rewrite indicE; case: (y \in A) => /=; rewrite ?(normr1,normr0).
+  - move=> y Kx; rewrite indicE.
+    by case: (y \in A) => /=; rewrite ?(normr1,normr0,lexx,lee01).
 by rewrite integral_cst//= ?mul1e; [exact: compact_finite_measure|
                                     exact: compact_measurable].
 Qed.
--- a/theories/lebesgue_integral.v
+++ b/theories/lebesgue_integral.v
@@ -6648,7 +6648,7 @@ Proof.
 move=> mA; have := lebesgue_differentiation (locally_integrable_indic openT mA).
 apply: filter_app; first exact: (ae_filter_ringOfSetsType mu).
 apply: aeW => /= x Ax.
-apply/(cvge_sub0 _ _).1 => //.
+apply: (cvge_sub0 _ _).1 => //.
 move: Ax; rewrite /lebesgue_pt /davg /= -/mu => Ax.
 have : (fine (mu (ball x r)))^-1%:E *
        `|\int[mu]_(y in ball x r) (\1_A y - \1_A x)%:E | @[r --> 0^'+] --> 0.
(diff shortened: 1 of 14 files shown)


theories/lebesgue_integral.v
```diff
@@ -5651,3 +5660,2383 @@ by rewrite ritv //= -EFinM lee_fin mulrC.
...
+  suff : g = h by move=> <-.
+  apply/funext => n.
+  rewrite/g /h /= fineM//=; last first.
+    apply: integral_fune_fin_num => //; first exact: (reg_cvg_E _).1.
+    by apply: integrableS intf => //; exact: (reg_cvg_E _).1.
+  by rewrite muE/= invrN mulNr -mulrN.
+Qed.
+
+Lemma FTC (f : R -> R) : mu.-integrable setT (EFin \o f) ->
+  let F x := fine (\int[mu]_(t in `]-oo, x]) (f t)%:E) in
```
CohenCyril, line 7932: I advocate to redefine the notation for the integral in `ring_scope` in ordet to read this instead:
```suggestion
  let F x := \int[mu]_(t in `]-oo, x]) f t in
```

affeldt-aist:
i have not yet done the changelog and I have not yet moved the lemmas to the right files but this might wait a first round of review I think
there is also still some room for improvement in the last bit (FTC0 and density) and I plan to go on revising it but I do not want to delay the review more than that

theories/lebesgue_integral.v
```diff
@@ -6033,3 +6041,1150 @@ rewrite -ge0_integral_bigsetU//=.
...
+have r0 : (0 < r)%R by near: r; exact: nbhs_right_gt.
+have [d /= d0] := fx _ e0.
+move=> {}fx; near=> t.
+have [t0|t0] := leP t 0%R; first by rewrite davg0.
+by rewrite fin_numM// exf//=.
+Unshelve. all: by end_near. Qed.
+
+(* TODO: should this be proved without the two Let's above? *)
+Lemma continuous_cvg_davg (E : set R) (f : R -> R) :
+  measurable_fun setT f -> {in E, continuous f} ->
```
zstone1, line 6211: The condition `{in E, continuous f}` appears several times throughout this PR, but is probably not what you mean. It has bad boundary conditions which are especially important when there are characteristic functions around. That is, `{in [0,1], continuous \1_([0,1])}` is false because it's not continuous at 0 or 1. 

Just changing everything to be `{within E, continuous f}` won't work. I'm pretty sure that's false. Instead, it seems like these are really all local properties, and it'll be nicer to state is that way.
```
measurable_fun setT f -> forall x, {at x, continuous f} ->  davg f x r @[r --> 0%R] --> 0
```
And even `measurable_fun setT f ` should be "measurable in a neighborhod of x".

theories/lebesgue_integral.v
```diff
@@ -6033,3 +6041,1150 @@ rewrite -ge0_integral_bigsetU//=.
...
+
+Definition lim_sup_davg f x := lime_sup (davg f x) 0.
+
+Local Notation "f ^*" := (lim_sup_davg f).
+
+Lemma lim_sup_davg_ge0 f x : 0 <= f^* x.
+Proof. by apply: lime_sup_ge0 => y; rewrite /davg iavg_ge0. Qed.
+
+Lemma lim_sup_davg_le f g x :
+  measurable_fun setT f -> locally_integrable setT g ->
```
zstone1, line 6256: Another example of a local property. That is, since `davgD` only need assumptions about `f` and `g` near `x`, I would expect this to have something similar.

Not blocking for this PR, but it does seem like useful generalization

theories/lebesgue_integral.v
```diff
@@ -6033,3 +6041,1150 @@ rewrite -ge0_integral_bigsetU//=.
...
+  + by move=> /andP[].
+  + case: yb => /= in xy *.
+      by move=> /le_lt_trans; apply.
+    by move=> /le_trans; apply; exact/ltW.
+  + by move=> /andP[].
+- by move: x => [[|] x|[|]//]/= in xy oox *.
+Qed.
+
+(* NB: since we only need derive after this point and since this file is very
+long I am wondering whether a new file should not be created now *)
```
zstone1, line 6873: Yeah, working in 7000 line files is rather annoying. I'd be happy to see this stuff split into its own.

zstone1 (commented):
Looks much improved. the FTC statements look much better with generic Interval stuff. There's definitely some stuff to shuffle around to other files, but seems like you are on top of that.

My main concern is a few lemmas that can be rephrased to be local statements. It should clean things up to remove the `{in E, continuous f}` which can be unintuitive.

PR #1068, opened 2023-10-23
fixes #1051 (rename `lim_sup` -> `limn_sup`)

### LINKED ISSUE #1051: `lim_sup` -> `limn_sup`
https://github.com/math-comp/analysis/blob/9fbd7cd0641431c0665c240f09e2ab3f7387cfba/theories/sequences.v#L2152

Would it be ok to rename `lim_sup` to `limn_sup`?
I am asking because there are "lim_sup" lemmas for `realType` in the pipeline
and it might be better to name these ones `lim_sup` (and not, say, `limr_sup`).
This also corresponds to the `limn` naming that the HB version of MathComp-Analysis is using.
Of course, a more unified mechanism would be better but that's another story for a bit later.

affeldt-aist:

fixes #1051

as (briefly) discussed during the last meeting https://github.com/math-comp/analysis/wiki/2023-10-16-Meeting

##### Things done/to do

### DIFF
--- a/theories/lebesgue_measure.v
+++ b/theories/lebesgue_measure.v
@@ -1562,15 +1562,15 @@ move=> _ [_ [x ->] <-]; rewrite infs_preimage // setI_bigcupr.
 by apply: bigcup_measurable => k /= nk; apply: mf => //; exact: measurable_itv.
 Qed.
 
-Lemma measurable_fun_lim_sup D (h : (T -> R)^nat) :
+Lemma measurable_fun_limn_sup D (h : (T -> R)^nat) :
   (forall t, D t -> has_ubound (range (h ^~ t))) ->
   (forall t, D t -> has_lbound (range (h ^~ t))) ->
   (forall n, measurable_fun D (h n)) ->
-  measurable_fun D (fun x => lim_sup (h ^~ x)).
+  measurable_fun D (fun x => limn_sup (h ^~ x)).
 Proof.
 move=> f_ub f_lb mf.
 have : {in D, (fun x => inf [set sups (h ^~ x) n | n in [set n | 0 <= n]%N])
-              =1 (fun x => lim_sup (h^~ x))}.
+              =1 (fun x => limn_sup (h^~ x))}.
   move=> t; rewrite inE => Dt; apply/esym/cvg_lim; first exact: Rhausdorff.
   rewrite [X in _ --> X](_ : _ = inf (range (sups (h^~t)))).
     by apply: cvg_sups_inf; [exact: f_ub|exact: f_lb].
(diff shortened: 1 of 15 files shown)


affeldt-aist:
TODO: add only parsing to deprecated notations!

affeldt-aist:
@CohenCyril the renaming is as we discussed during the last meeting, plus one name change that I made to be more uniform between the real and the ereal versions
I put "only parsing" format info for all deprecated (they shoud have been here from the start)

theories/lebesgue_measure.v
```diff
@@ -1459,17 +1459,17 @@ End standard_measurable_fun.
 #[global] Hint Extern 0 (measurable_fun _ (fun x => x ^+ _)) =>
   solve [exact: measurable_exprn] : core.
 #[deprecated(since="mathcomp-analysis 0.6.3", note="use `measurable_exprn` instead")]
-Notation measurable_fun_sqr := measurable_exprn.
+Notation measurable_fun_sqr := measurable_exprn (only parsing).
```
CohenCyril, line 1462: I didn't know Coq accepted this... (only parsing syndefs...)

CohenCyril:
Sure! I agree with the renaming. Can you tell me why the "only parsing" for lemmas (where are they printed anyway?)

affeldt-aist:
I think that I only added them to `Notation`.

CohenCyril:
I meant: why bother putting only parsing on notations for lemma. What else than `Print` is going to display them anyway

affeldt-aist:
I'll remove them asap.

affeldt-aist:
> I'll remove them asap.

Well, in fact, besides `Print`, they are useful to have correct error display.
If you apply the lemma for which a deprecated notation is not `only parsing`,
the error will report about the deprecated notation instead of the right lemma,
which is a bit disturbing (it happens a lot with `measurable_*` lemmas).

CohenCyril:
> > I'll remove them asap.
> 
> Well, in fact, besides `Print`, they are useful to have correct error display. If you apply the lemma for which a deprecated notation is not `only parsing`, the error will report about the deprecated notation instead of the right lemma, which is a bit disturbing (it happens a lot with `measurable_*` lemmas).

I do not understand. It's good that the error is about the deprecated lemma, and that the warning notice gives the new lemma. Can you give me a concrete usecase, with a concrete error message and what's wrong with it. I tried to reproduce what you said but for me the result is unchanged whether I put `only parsing` or not...

affeldt-aist:
For example, in the current state, if you apply, say, `measurableT_comp` in the second half of `lebesgue_measure.v`, if it fails, it fails with the error
```
Error: Cannot apply lemma emeasurable_funN
```
which is a bit puzzling at first sight.
With `only parsing`, at least the error message is
```
Error: Cannot apply lemma measurableT_comp
```

CohenCyril:
Oh I undestand now!! Thanks for the clarification:

CohenCyril (approved):

CohenCyril:
This is really nontrivial, could you document it somewhere?

affeldt-aist:
> This is really nontrivial, could you document it somewhere?

What about an entry in CONTRIBUTING.md about how to write deprecated pragmas?

affeldt-aist:
I added a few sentence in `CONTRIBUTING.md`

PR #1121, opened 2023-12-21
isolate the theory of lime_sup

affeldt-aist:

from the PR #1065 on the Lebesgue differentiation theorem

##### Things done/to do

### DIFF
--- a/theories/realfun.v
+++ b/theories/realfun.v
@@ -152,38 +152,14 @@ Lemma cvg_at_leftP (f : R -> R) (p l : R) :
   (forall u : R^nat, (forall n, u n < p) /\ (u --> p) ->
     f (u n) @[n --> \oo] --> l).
 Proof.
-split=> [/cvgrPdist_le fpl u [up /cvgrPdist_lt ucvg]|pfl].
-  apply/cvgrPdist_le => e e0.
-  have [r /= r0 {}fpl] := fpl _ e0; have [s /= s0 {}ucvg] := ucvg _ r0.
-  near=> t; apply: fpl => //=; apply: ucvg => /=.
-  by near: t; exists s.
-apply: contrapT => fpl; move: pfl; apply/existsNP.
-suff: exists2 x : R ^nat,
-    ((forall k, x k < p) /\ x --> p) & ~ f (x n) @[n --> \oo] --> l.
-  by move=> [x_] h; exists x_; apply/not_implyP.
-have [e He] : exists e : {posnum R}, forall d : {posnum R},
-    exists xn : R, [/\ xn < p, `|xn - p| < d%:num & `|f xn - l| >= e%:num].
-  apply: contrapT; apply: contra_not fpl => /forallNP h.
-  apply/cvgrPdist_le => e e0; have /existsNP[d] := h (PosNum e0).
-  move/forallNP => {}h; near=> t.
-  have /not_and3P[abs|abs|/negP] := h t.
-  - by exfalso; apply: abs; near: t; exact: nbhs_left_lt.
-  - exfalso; apply: abs.
-    by near: t; exists d%:num => //= z/=; rewrite distrC.
-  - by rewrite -ltNge distrC => /ltW.
-have invn n : 0 < n.+1%:R^-1 :> R by rewrite invr_gt0.
-exists (fun n => sval (cid (He (PosNum (invn n))))).
-  split => [k|]; first by rewrite /sval/=; case: cid => x [].
-  apply/cvgrPdist_lt => r r0; near=> t.
-  rewrite /sval/=; case: cid => x [px xpt _].
-  rewrite distrC (lt_le_trans xpt)// -(@invrK _ r) lef_pinv ?posrE ?invr_gt0//.
-  near: t; exists `|ceil (r^-1)|%N => // s /=.
-  rewrite -ltnS -(@ltr_nat R) => /ltW; apply: le_trans.
-  by rewrite natr_absz gtr0_norm ?ceil_gt0 ?invr_gt0// ceil_ge.
-move=> /cvgrPdist_lt/(_ e%:num (ltac:(by [])))[] n _ /(_ _ (leqnn _)).
-rewrite /sval/=; case: cid => // x [px xpn].
-by rewrite leNgt distrC => /negP.
-Unshelve. all: by end_near. Qed.
+apply: (iff_trans (cvg_at_leftNP f p l)).
+apply: (iff_trans (cvg_at_rightP _ _ _)).
+split=> [pfl u [pu up]|pfl u [pu up]].
+  rewrite -(opprK u); apply: pfl.
+  by split; [move=> k; rewrite ltr_oppr opprK//|exact/cvgNP].
+apply: pfl.
+by split; [move=> k; rewrite ltr_oppl//|apply/cvgNP => /=; rewrite opprK].
+Qed.
 
 End cvgr_fun_cvg_seq.
 
--- a/theories/realfun.v
+++ b/theories/realfun.v
@@ -199,6 +199,18 @@ have {mn} := mfy_ _ mn.
 rewrite /y_ /sval; case: cid2 => /= x _.
 Unshelve. all: by end_near. Qed.
 
+Lemma cvge_at_leftP (f : R -> \bar R) (p l : R) :
+  f x @[x --> p^'-] --> l%:E <->
+  (forall u : R^nat, (forall n, u n < p) /\ u --> p ->
+    f (u n) @[n --> \oo] --> l%:E).
+Proof.
+apply: (iff_trans (cvg_at_leftNP f p l%:E)).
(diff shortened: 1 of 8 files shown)


affeldt-aist:
What about adding these (loosely related) lemmas https://github.com/zstone1/analysis/blob/total_variation/theories/realfun.v#L1323-L1352
to this PR?

affeldt-aist:
> What about adding these (loosely related) lemmas https://github.com/zstone1/analysis/blob/total_variation/theories/realfun.v#L1323-L1352 to this PR?

at least related to this one https://github.com/affeldt-aist/analysis/blob/limr_sup/theories/realfun.v#L88-L105

theories/realfun.v
```diff
@@ -80,8 +85,147 @@ apply: near_eq_cvg; near do rewrite subrK; exists M.
...
+  move=> z/= + _ => /lt_le_trans; apply.
+  by rewrite ler_pdivr_mulr// ler_pmulr// ler1n.
+Unshelve. all: by end_near. Qed.
+
 End fun_cvg_realFieldType.
 
+Section cvgr_fun_cvg_seq.
+Context {R : realType}.
+
+Lemma cvg_at_rightP (f : R -> R) (p l : R) :
```
zstone1, line 112: In a textbook all four of these proofs are, this would be a three liners:
1. In first countable spaces, filter convergence <-> sequence convergence
2. The lower (upper) limit topology (AKA sogenfrey line) is first countable
3. All sequences in the lower (upper) limit topology that converge, do so from the left (right).

The proof you have is fine, I'm not complaining about it. I am complaining that some missing classical machinery is making these proofs longer than necessary. A problem for another day, I guess.

theories/realfun.v
```diff
@@ -80,8 +85,147 @@ apply: near_eq_cvg; near do rewrite subrK; exists M.
...
+  rewrite distrC (lt_le_trans xpt)// -(@invrK _ r) lef_pinv ?posrE ?invr_gt0//.
+  near: t; exists `|ceil (r^-1)|%N => // s /=.
+  rewrite -ltnS -(@ltr_nat R) => /ltW; apply: le_trans.
+  by rewrite natr_absz gtr0_norm ?ceil_gt0 ?invr_gt0// ceil_ge.
+move=> /cvgrPdist_lt/(_ e%:num (ltac:(by [])))[] n _ /(_ _ (leqnn _)).
+rewrite /sval/=; case: cid => // x [px xpn].
+by rewrite leNgt distrC => /negP.
+Unshelve. all: by end_near. Qed.
+
+Lemma cvg_at_leftP (f : R -> R) (p l : R) :
```
zstone1, line 150: Do we have a statement about `f x @[x --> p^'-] --> l <-> f x @[x --> p^'+] --> -l`? Then at least we can leverage the one above.

theories/realfun.v
```diff
@@ -393,6 +537,314 @@ End fun_cvg_ereal.
 
 End fun_cvg.
 
+(*Section lim_sup_inf.
```
zstone1, line 540: Remove commented, or explain its purpose

theories/realfun.v
```diff
@@ -393,6 +537,314 @@ End fun_cvg_ereal.
...
+rewrite (@le_trans _ _ (ereal_sup [set f x | x in ball a d `\ a]))//; last first.
+  apply: le_ereal_sup => z/= [r [adr rp] <-{z}]; exists r => //; split => //.
+  by rewrite /ball/= (lt_le_trans adr)// /d le_minl lexx orbT.
+apply: ereal_sup_ub => /=; exists (u n).
+  split; last by apply/eqP; rewrite eq_sym lt_eqF.
+  by apply: ucvg => //=; near: n; exists m.
+by rewrite fineK//; near: n.
+Unshelve. all: by end_near. Qed.
+
+Local Lemma lime_supP' f a l :
```
zstone1, line 832: Ending in a local lemma is a bit surprising What's the intention of this one?

theories/realfun.v
```diff
@@ -393,6 +537,314 @@ End fun_cvg_ereal.
...
+Let le_sup_ball f g a :
+  (forall r, (0 < r)%R -> forall y : R, y != a -> ball a r y -> f y <= g y) ->
+  \forall r \near 0^'+, sup_ball f a r <= sup_ball g a r.
+Proof.
+move=> fg; near=> r; apply: ub_ereal_sup => /= _ [s [pas /= /eqP ps]] <-.
+apply: (@le_trans _ _ (g s)); first exact: (fg r).
+by apply: ereal_sup_ub => /=; exists s => //; split => //; exact/eqP.
+Unshelve. all: by end_near. Qed.
+
+Definition lime_sup f a : \bar R := lim (sup_ball f a e @[e --> 0^'+]).
```
zstone1, line 581: The lim_sup of a function vs. the lim_sup of a sequence is a bit weird. I normally think of it as a the sequence kind. 

It would be satisfying to end up with the sequence definition taking a `(f : R^nat)` and being `lime_sup f \oo`. Is it possible to generalize this to making `a` a filter? Then the `a^'` filter does the rest of the work.

zstone1 (commented):
The repeated proofs about swapping between sequential and filter convergence are annoying, but tough to fix without a little extra machinery. 

The interesting question is about generalizing this to consider the lim_sup of a function "at a filter". If it's an easy generalization, we should do it now. If it's tough, then we may have to revisit the material when we do sequence limsups.

theories/realfun.v
affeldt-aist: Yes we do, it is in the total variation PR ^_^ : https://github.com/zstone1/analysis/blob/total_variation/theories/realfun.v#L1323-L1338
Allow me to add it to this PR, see commit https://github.com/math-comp/analysis/pull/1121/commits/31ea75483a931094484e0270b60d1dffd1705c50

theories/realfun.v
affeldt-aist: This looks like some leftover. I deleted it.

theories/realfun.v
affeldt-aist: This has been uncommented and moved to `normedtype.v` to act as a generic definition of limsup and liminf from which `lim_esup` and `limn_esup` are now derived (see the last commit).

theories/realfun.v
affeldt-aist: This is addressed by the last commit https://github.com/math-comp/analysis/pull/1121/commits/b3a2daca7828e58fcb993fda7a2f1f61aad6ff81

theories/realfun.v
```diff
@@ -393,6 +512,293 @@ End fun_cvg_ereal.
...
+  lime_inf f a = ereal_sup [set inf_ball f a e | e in `]0, +oo[ ]%R.
+Proof. by rewrite /lime_inf lime_supE /ereal_inf oppeK image_comp. Qed.
+
+Lemma lime_infN f a : lime_inf (\- f) a = - lime_sup f a.
+Proof. by rewrite /lime_sup -limf_einfN. Qed.
+
+Lemma lime_supN f a : lime_sup (\- f) a = - lime_inf f a.
+Proof. by rewrite /lime_inf oppeK. Qed.
+
+Lemma lime_sup_ge0 f a : (forall x, 0 <= f x) -> 0 <= lime_sup f a.
```
zstone1, line 599: Most of the important lemmas like this one are probably true of `limf_sup`. At least lime_sup_ge0,, lime_supD, lime_inf_sup, and lim_lime_inf. There may be some conditions to replace with `\forall x \near F`, but otherwise I'm hoping it just works. 

Is it much effort to generalize them now, or do you want to do as a separate PR?

theories/realfun.v
```diff
@@ -393,6 +512,293 @@ End fun_cvg_ereal.
...
+  lime_inf f a = l%:E -> forall e : {posnum R}, exists d : {posnum R},
+  forall x, (ball a d%:num `\ a) x -> l%:E - e%:num%:E < f x.
+Proof.
+move=> /(congr1 oppe); rewrite -lime_supN => /lime_supP => H e.
+have [d {}H] := H e.
+by exists d => r /H; rewrite lte_oppl oppeD// EFinN oppeK.
+Qed.
+
+Lemma lime_inf_sup_lim f a l :
+  lime_sup f a = l%:E -> lime_inf f a = l%:E -> f x @[x --> a^'+] --> l%:E.
```
zstone1, line 748: This limit from the right is surprising. Isn't it true from the left too?

zstone1 (commented):
One meaningful question about the right sided limit. And then, if you're up for it, generalizing more of the lemmas to work for limf

theories/realfun.v
affeldt-aist: Indeed. See the last commit https://github.com/math-comp/analysis/pull/1121/commits/c8915c43fa6a3f83b019a28a4e9951c717df19db

theories/realfun.v
affeldt-aist: I'd prefer to issue about your comments and do it later because I have been delaying other tasks recently. I'll go back to it after the release of 0.6.7.

zstone1 (approved):

PR #1181, opened 2024-03-05
fixes #1178 (`ge0_integralZr`)

### LINKED ISSUE #1178: `ge0_integralZr` is lacking
https://github.com/math-comp/analysis/blob/0105ccc0256603930e5ffd30156bda7b81e7e4ce/theories/lebesgue_integral.v#L2288

affeldt-aist:

fixes #1178 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -7,6 +7,9 @@
 - in `exp.v`
   + lemma `ln_lt0`
 
+- in `lebesgue_integral.v`
+  + lemma `ge0_integralZr`
+
 ### Changed
 
 - move from `kernel.v` to `measure.v`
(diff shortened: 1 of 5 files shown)


affeldt-aist:
> Why is this section called "semilinearity"? It looks different from the definition of semilinearity found in other places like https://ncatlab.org/nlab/show/semilinear+map

The textbook I was using calls semilinearity of the integral the combination of the properties of ge0_integralZl_EFin and ge0_integralD (however when considered over the set of simple functions), hence the name of the section. The definition you are pointing at is indeed more general.

affeldt-aist:
> Are there any places ge0_integralZr can be used to simplify proofs? Maybe in hoelder.v

Thanks for insisting on this point.
In the last commit, I used `ge0_integralZr` instead of `ge0_integralZl` where it looked like superfluous commutativity lemmas where used, leasing to a bit shorter proofs. I have also another example in a forthcoming PR.

PR #1196, opened 2024-03-28
Bigcap series addn

IshiguroYoshihiro:

lemmas for series of bigcup, bigcap and sum with shifting index by `addn`.

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -9,105 +9,17 @@
 
 - in `sequences.v`:
   + lemma `nneseries_recl`
-- in `lebesgue_integral.v`
-  + lemma `ge0_integralZr`
-- file `function_spaces.v`
-- in `mathcomp_extra.v`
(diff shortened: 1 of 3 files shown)


affeldt-aist (approved):
Looks ok to me.

CohenCyril:
Ok except I would probably reverse the identities (the best looking term on the right, and the naming would correspond to the left hand side as generally expected)

affeldt-aist:
Ok, will do that. (In fact, I think that we were writing the statement following `big_addn` in which the shift of index appears on the right:
```coq
Lemma big_addn m n a (P : pred nat) F :
  \big[op/idx]_(m + a <= i < n | P i) F i =
     \big[op/idx]_(m <= i < n - a | P (i + a)) F (i + a).
```
But indeed putting the nice-looking term on the right gives a better sense of "simplication".)

PR #1199, opened 2024-03-28
fixes #981 (`sub_additive` vs `subadditive`)

### LINKED ISSUE #981: Naming of sigma subadditivity
https://github.com/math-comp/analysis/blob/7622cc7f7392b8866d7900ffb6ab6b1e48e81a7f/theories/measure.v#L2514

Shouldn't this be `Content_SigmaSubAdditive_isMeasure` to match the name of the predicate `sigma_sub_additive`?

Also, we have `sigma_sub_additive` and `sigma_subadditive`. 
What about using `sigma_subset_subadditive` for `sigma_sub_additive`?
(Or maybe the intent of using the substring `_sub_` was to stand for `subset`, but still...)

affeldt-aist:

fixes #981 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,6 +4,16 @@
 
 ### Added
 
+- in `classical_sets.v`:
+  + lemma `bigcup_recl`
+
+- in `sequences.v`:
+  + lemma `nneseries_recl`
+
+- in `measure.v`:
+  + definition `subset_sigma_subadditive`
+  + factory `isSubsetOuterMeasure`
+
 ### Changed
 
 ### Renamed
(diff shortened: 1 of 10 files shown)


affeldt-aist:
How about this @CohenCyril ?

affeldt-aist:
> How about this @CohenCyril ?

This way, the identifiers of predicates about sigma-subaddivity all share the "sigma_subadditivity" substring (or "SigmaSubAdditivity" for the mixin), in the default interface for outer measure we use the "version without subset" (the "alternative" version by Halmos according to https://en.wikipedia.org/wiki/Outer_measure), the other version is provided by the factory `isOuterMeasureAlt`.
In particular, `sigma_sub_additive` has been renamed `measurable_subset_sigma_subadditive` to make clear that it has an extra measurability condition, i.e., it is not the predicate used to define outer measure.

I think these renamings clarify the role of the various predicates thought there might be even more to do.

affeldt-aist:
NB: got an offline approval by @CohenCyril

PR #1244, opened 2024-06-17
fixes #330

### LINKED ISSUE #330: floor and ceil to be subsumed by MathComp
TODO: revise the use of floor and ceil once ~~math-comp/math-comp#682 is merged into MathComp~~ requiring MC >= 2.1.0

affeldt-aist:

fixes #330 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/reals.v
```diff
@@ -244,15 +244,15 @@ Variable R : realType.
 Implicit Types x y : R.
 Definition floor_set x := [set y : R | (y \is a Rint) && (y <= x)].
 
-Definition Rfloor x : R := sup (floor_set x).
+Definition floor x : int := Num.floor x.
```
proux01, line 247: Should this become a notation?

affeldt-aist: I was wondering. Surely, we do not need the definition `floor` anymore, require to import the module `Num` seems cumbersome, and imposing the use of `Num.floor` as well, so, yes, a notation is worth trying. I will try ASAP.

affeldt-aist:
@proux01 could you take a quick look and potentially approved?
You'll notice that now `reals.v` is almost free of `floor` or `ceil` usage (except three deprecations).
What could be generalized to archimedean structures has been put into `mathcomp_extra.v` and
and this content is waiting for https://github.com/math-comp/math-comp/pull/1235 to make its way into MathComp releases (so this'd rather be considered as a stage area).

After that it would remain to see if we still need the `Rceil` and `Rfloor` theories but that is maybe for a future PR.

affeldt-aist:
> so this'd rather be considered as a stage area

for example, I'm keep `ceil_ge` has a lemma even though it is subsumed by `Num.Theory.le_ceil`
because `le_ceil` already existed with a different meaning and it now in deprecation phase

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,83 @@ by elim: p => //= p <-;
...
+Lemma intr1 {R : ringType} (i : int) : i%:~R + 1 = (i + 1)%:~R :> R.
+Proof. by rewrite intrD. Qed.
+
+Lemma int1r {R : ringType} (i : int) : 1 + i%:~R = (1 + i)%:~R :> R.
+Proof. by rewrite intrD. Qed.
+
+From mathcomp Require Import archimedean.
+
+Lemma natr_def (n : int) : (n \is a Num.nat) = (0 <= n)%R.
+Proof. by []. Qed.
```
pi8027, line 387: I don't think you need to add this lemma since it is available as `Znat_def` if you import `archimedean.Num.Theory`.

theories/ereal.v
```diff
@@ -1402,21 +1402,24 @@ Lemma cvg_ereal_loc_seq (R : realType) (x : \bar R) :
 Proof.
 move=> P; rewrite /ereal_loc_seq.
 case: x => /= [x [_/posnumP[d] dP] |[d [dreal dP]] |[d [dreal dP]]]; last 2 first.
-    have /ZnatP[N Nfloor] : floor (Num.max d 0%R) \is a Znat.
-      by rewrite Znat_def floor_ge0 le_max lexx orbC.
+    have /natrP[N Nfloor] : Num.floor (Num.max d 0%R) \is a Num.nat.
+      by rewrite natr_def floor_ge0 le_max lexx orbC.
```
pi8027, line 1406: I don't think you need to replace `Znat_def` with `natr_def` (for the same reason).

classical/mathcomp_extra.v
affeldt-aist: But it is a notation for `mc_2_0.Znat_def` which looks a bit like a deprecated thing. Why this name?

theories/Rstruct.v
```diff
@@ -292,7 +292,7 @@ rewrite /Rminus Rplus_assoc [- _ + _]Rplus_comm -Rplus_assoc -!/(Rminus _ _).
 exact: Rle_minus.
 Qed.
 
-HB.instance Definition _ := Num.RealField_isArchimedean.Build R
+HB.instance Definition _ := Num.NumDomain_bounded_isArchimedean.Build R
```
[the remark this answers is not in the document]
affeldt-aist: it seems that you do not need to

classical/mathcomp_extra.v
pi8027: That's because we needed to deprecate it in `ssrint.v` in MC 2.1:
```coq
Notation Znat_def := mc_2_0.Znat_def (only parsing).
#[deprecated(since="mathcomp 2.1.0", note="Require archimedean.v.")]
```
As you can see, you have to `Require Import archimedean` (then also `Num.Theory`) to use it but don't have to rename it.

pi8027: (The archimedean structures are eventually moved to `archimedean.v`, which depends on `ssrint.v`. It's clear that we cannot state `Znat_def` in `ssrint.v` after actually moving the definitions of these structures.)

affeldt-aist: In `ereal.v`, we do `Require Import archimedean` but then using `Znat_def` generates a deprecation warning (with or without `Import Num.Theory`) and then using `Znat_def` generated a deprecation warning. You mean that this is intended?

affeldt-aist: Is it that the deprecation warning meams "don't mind the deprecation warning as long as you do Require Import archimedean"?

pi8027: No, but you also need to make sure to import `archimedean.Num.Theory` after importing `ssrint`.

theories/Rstruct.v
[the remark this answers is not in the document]
pi8027: Although the deprecation message is not perfect, we have to move the definition of this factory to `archimedan.v` in math-comp/math-comp#1237.

classical/mathcomp_extra.v
affeldt-aist: Sorry, I don't get it :-(
To be clear: You ask to use `Znat_def` even though it generates a deprecation warning?
(by the way, shouldn't `archimedean.v` be a part of `all_algebra.v`?)

theories/Rstruct.v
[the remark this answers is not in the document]
affeldt-aist: Ah but here the deprecation message is fine, and indeed doing the renaming works ok (and it compiles fine without even Require Import archimedean by the way)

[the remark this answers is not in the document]
pi8027: > it compiles fine without even Require Import archimedean by the way

That shouldn't be the case after math-comp/math-comp#1237. So please import `archimedean`.

classical/mathcomp_extra.v
affeldt-aist: ok, then I will use `Znat_def` despite the deprecation message that will go away some time in the future

theories/Rstruct.v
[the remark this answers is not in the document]
affeldt-aist: ok, I see, it is not needed yet but will be in the future, so it'd better be done asap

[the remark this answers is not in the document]
pi8027: Note that I also need it to compile analysis with my PR in Nix CI.

affeldt-aist:
The CI errors look unrelated.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,80 @@ by elim: p => //= p <-;
...
 
 End positive.
+
+Lemma intr1 {R : ringType} (i : int) : i%:~R + 1 = (i + 1)%:~R :> R.
+Proof. by rewrite intrD. Qed.
+
+Lemma int1r {R : ringType} (i : int) : 1 + i%:~R = (1 + i)%:~R :> R.
+Proof. by rewrite intrD. Qed.
+
+From mathcomp Require Import archimedean.
```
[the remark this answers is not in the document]
affeldt-aist: Because only the next section needs it; if at some point this section moves to mathcomp I will not forget to erase even the Require Import.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,80 @@ by elim: p => //= p <-;
...
+
+Section floor_ceil.
+Context {R : archiDomainType}.
+Implicit Type x : R.
+
+Lemma ge_floor x : (Num.floor x)%:~R <= x.
+Proof. by rewrite Num.Theory.ge_floor. Qed.
+
+Lemma floor_ge0 x : (0 <= Num.floor x) = (0 <= x).
+Proof. by rewrite -Num.Theory.floor_ge_int. Qed.
```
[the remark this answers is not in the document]
affeldt-aist: Yes of course. I'll do that asap if you want. (Note that I still need to double-check the naming before thinking of proposing to move the lemmas to MathComp proper, which might require us to wait a bit because of name clashes.)

affeldt-aist:
perfect

theories/altreals/realsum.v
```diff
@@ -142,27 +141,27 @@ Proof.
...
   move=> x; rewrite !inE => nz_fx.
-  pose j := `|floor (1 / `|f x|)|%N; exists j; rewrite inE.
-  rewrite ltr_pdivrMr ?ltr0z // -ltr_pdivrMl ?normr_gt0 //.
-  rewrite mulr1 /j div1r -addn1 /= PoszD intrD mulr1z.
-  rewrite gez0_abs ?floor_ge0 ?invr_ge0 ?normr_ge0 //.
-  by rewrite -RfloorE; apply lt_succ_Rfloor.
+  pose j := `|Num.floor (`|f x|^-1)|%N; exists j; rewrite inE.
+  rewrite mul1r invf_plt ?unfold_in /= ?ltr0z ?normr_gt0 //.
+  rewrite /j -addn1 /= PoszD gez0_abs ?floor_ge0 ?invr_ge0 ?normr_ge0//.
+  by rewrite mathcomp_extra.lt_succ_floor.
```
pi8027, line 147: In order to avoid this `mathcomp_extra` prefix, you should probably import `mathcomp_extra` after importing `Num.Theory`. Same for `rewrite ?num_real`.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+Proof. by move/Num.Theory.floor_le; rewrite Num.Theory.floor0. Qed.
+
+Lemma floor_lt0 x : x < 0 -> Num.floor x < 0.
+Proof. by rewrite -floor_lt_int. Qed.
+
+Lemma lt_succ_floor x : x < (Num.floor x + 1)%:~R.
+Proof. exact: Num.Theory.lt_succ_floor. Qed.
+
+Lemma floor_natz n : Num.floor (n%:R : R) = n%:~R :> int.
+Proof. by rewrite (Num.Theory.intrKfloor n) intz. Qed.
```
pi8027, line 412: You can probably remove this lemma and use `intrKfloor` instead.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+
+Lemma floor_neq0 x : (Num.floor x != 0) = (x < 0) || (x >= 1).
+Proof. by rewrite neq_lt -floor_lt_int gtz0_ge1 -floor_ge_int. Qed.
```
pi8027, line 415: I rather wish to have a lemma like:
```coq
forall (R : archiRealDomainType) (x : R) (m : int), (Num.floor x == m) = (m%:~R <= x < (m + 1)%:~R)
```
which is an "iff" version of `floor_def`.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+
+Lemma ceil_ge x : x <= (Num.ceil x)%:~R.
+Proof. exact: Num.Theory.le_ceil. Qed.
```
pi8027, line 418: Apparently, this lemma is called `le_ceil` in MathComp.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+
+Lemma ceil_ge_int x (z : int) : (x <= z%:~R) = (Num.ceil x <= z).
+Proof. exact: Num.Theory.ceil_le_int. Qed.
```
pi8027, line 421: `ceil_le_int` ?

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+
+Lemma ceil_lt_int x (z : int) : (z%:~R < x) = (z < Num.ceil x).
+Proof. by rewrite ltNge ceil_ge_int -ltNge. Qed.
```
pi8027, line 424: `ceil_gt_int` ?

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+
+Lemma ceil_ge0 x : 0 <= x -> 0 <= Num.ceil x.
+Proof. by move/(ge_trans (ceil_ge x)); rewrite -(ler_int R). Qed.
```
pi8027, line 427: Again, `0 <= x` can be generalized to `- 1 < x` here.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
 Qed.
 
 End positive.
+
+Lemma intr1 {R : ringType} (i : int) : i%:~R + 1 = (i + 1)%:~R :> R.
+Proof. by rewrite intrD. Qed.
+
+Lemma int1r {R : ringType} (i : int) : 1 + i%:~R = (1 + i)%:~R :> R.
+Proof. by rewrite intrD. Qed.
```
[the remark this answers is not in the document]
pi8027: For example, `intrD1` and `intr1D`, or `intrSr` and `intrSl`.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,72 @@ by elim: p => //= p <-;
...
+Proof. exact: Num.Theory.floor_ge_int. Qed.
+
+Lemma floor_lt_int x (z : int) : (x < z%:~R) = (Num.floor x < z).
+Proof. by rewrite ltNge floor_ge_int -ltNge. Qed.
+
+Lemma floor_ge0 x : (0 <= Num.floor x) = (0 <= x).
+Proof. by rewrite -floor_ge_int. Qed.
+
+Lemma floor_le0 x : x <= 0 -> Num.floor x <= 0.
+Proof. by move/Num.Theory.floor_le; rewrite Num.Theory.floor0. Qed.
```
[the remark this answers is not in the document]
affeldt-aist: > Do you really need this lemma?

At least it is used. Of course, you can inline the proof but I think that in practice it is good to have lemmas for constants for often show up like 0, even if they have short proofs; moreover they are easily named from the general case. I'll look a bit more in detail though.

theories/altreals/realsum.v
```diff
@@ -142,27 +142,25 @@ Proof.
 case/summableP=> M ge0_M bM; pose E (p : nat) := [pred x | `|f x| > 1 / p.+1%:~R].
 set F := [pred x | _]; have le: {subset F <= [pred x | `[< exists p, x \in E p >]]}.
   move=> x; rewrite !inE => nz_fx.
-  pose j := `|floor (1 / `|f x|)|%N; exists j; rewrite inE.
-  rewrite ltr_pdivrMr ?ltr0z // -ltr_pdivrMl ?normr_gt0 //.
-  rewrite mulr1 /j div1r -addn1 /= PoszD intrD mulr1z.
-  rewrite gez0_abs ?floor_ge0 ?invr_ge0 ?normr_ge0 //.
-  by rewrite -RfloorE; apply lt_succ_Rfloor.
+  pose j := `|Num.floor (`|f x|^-1)|%N; exists j; rewrite inE.
```
pi8027, line 145: There are many occurrences of `absz (Num.floor x)`. Did you check if `Num.trunc` can replace some of them or not? (That would be the case when `x` is positive and `absz` serves just as a type cast from `int` to `nat`.)

classical/mathcomp_extra.v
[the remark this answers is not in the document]
proux01: I prefer `intrD1` and `intr1D`, I feel `S` is specific to nat `.+1`.

classical/mathcomp_extra.v
```diff
@@ -374,3 +374,74 @@ by elim: p => //= p <-;
...
+Lemma floor_ge0 x : (0 <= Num.floor x) = (0 <= x).
+Proof. by rewrite -floor_ge_int. Qed.
+
+Lemma floor_le0 x : x < 1 -> Num.floor x <= 0.
+Proof.
+by rewrite (_ : 1 = 1%:~R)// floor_lt_int -ltzD1 add0r => /le_lt_trans; exact.
+Qed.
+
+Lemma floor_lt0 x : x < 0 -> Num.floor x < 0.
+Proof. by rewrite -floor_lt_int. Qed.
```
pi8027, line 408: These implications `->` can be replaced with equations `=`.

pi8027: Done.

classical/mathcomp_extra.v
pi8027: Removed.

classical/mathcomp_extra.v
affeldt-aist: Not pushed yet.

affeldt-aist: Now that my PC is back to life I push what I have.
I'd like to also flip equality lemmas such as ceil_ge0 to have the hand-side with the main identifier on the left.

pi8027:
https://github.com/math-comp/analysis/pull/1244#discussion_r1679062772

> I'd like to also flip equality lemmas such as ceil_ge0 to have the hand-side with the main identifier on the left.

I agree, but these lemmas are already in MathComp. We should discuss what we do in a MathComp dev meeting.

pi8027 (approved):
We should probably clean up the proofs further, but it is already mergeable IMO.

pi8027:
@affeldt-aist Shall we merge? Or do you want me to wait?

affeldt-aist:
> @affeldt-aist Shall we merge? Or do you want me to wait?

Let's merge. Many thanks for all the input.

PR #1246, opened 2024-06-20
FTC1 corollary

affeldt-aist:

See issue #965 

Provides also the Borel-Cantelli lemma (and the limit superior of the sequence of sets) and continuity of the indefinite integral at the borders (which is maybe the part that is not the cleanest).

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/ftc.v
```diff
@@ -215,3 +217,277 @@ by split; [exact: dfx|rewrite f'xE].
...
+
+Section indefinite_integral_continuous.
+Context {R : realType}.
+Notation mu := (@lebesgue_measure R).
+
+Let int := (indefinite_integral mu).
+
+Lemma indefinite_integral_near_left (a b : R) (e : R) (f : R -> R) :
+  a < b -> 0 < e -> mu.-integrable `[a, b] (EFin \o f) ->
+  \forall c \near a, `| int a c f | < e.
```
zstone1, line 234: It's a bit uncomfortable that the `\forall c` quantifier here includes `c < a`. I guess those integrals become 0, but I don't think you are intending to work with those. Just doing 
```
  \forall c \near a, a <= c -> `| int a c f | < e.
```
should eliminate the bad cases. You do this in the casing in the proof anyway, so it shouldn't change much.

theories/ftc.v
```diff
@@ -215,3 +217,277 @@ by split; [exact: dfx|rewrite f'xE].
...
+  - by move=> x _; rewrite lee_fin ler_norm.
+rewrite [leLHS](_ : _ = `| lhs |)%E//; last first.
+  by move: lhsNoo lhsoo; rewrite /lhs /rhs; case: integral.
+apply: (le_trans (le_abse_integral _ _ _)) => //=.
+by apply: (measurable_funS _ acbc) => //; exact: measurable_int intf.
+Unshelve. all: by end_near. Qed.
+
+Lemma indefinite_integral_cvg_left a b (f : R -> R) : a < b ->
+  mu.-integrable `[a, b] (EFin \o f) ->
+  int a x f @[x --> a] --> int a a f.
```
zstone1, line 290: Why do we use `int a a f` instead of 0?

theories/ftc.v
```diff
@@ -215,3 +217,277 @@ by split; [exact: dfx|rewrite f'xE].
...
+- rewrite lte_fin int_normr_cont// lebesgue_measure_itv/= lte_fin.
+  rewrite ifT// -EFinD lte_fin.
+  near: x; exists d => // z; rewrite /ball_/= => + zb.
+  by rewrite gtr0_norm// ?subr_gt0.
+- rewrite ge0_fin_numE//; last exact: integral_ge0.
+  apply: integral_fune_lt_pinfty => //=.
+  rewrite [X in _.-integrable _ X](_ : _ = abse \o (EFin \o fab))//.
+  by apply: integrable_abse; exact: integrableS intfab.
+Unshelve. all: by end_near. Qed.
+
```
zstone1, line 371: It might be nice to combine these results into the more concise
```
indefinite_integral_continuous := ... -> {within `[a,b], continuous (fun x => int a x f)} 
```
using `continuous_within_itvP`

zstone1 (commented):
A couple small comments. Going via "charge_variation is uniformly continuous" is clever indeed. That's a nice trick to stay general for as long as possible.

theories/ftc.v
affeldt-aist: 0 is better

theories/ftc.v
affeldt-aist: This is the lemma `parameterized_integral_continuous` in the last commit

affeldt-aist:
@t6s we renamed `indefinite_integral` to `parameterized_integral` following the last mathcomp-analysis-dev meeting

affeldt-aist:
Got an offline approval by Zachary for this PR, will merge soon.

PR #672, opened 2022-06-03
trying to generalize mfun

affeldt-aist:
##### Motivation for this change

@CohenCyril 

would fix issue #662 

##### Things done/to do

- [ ] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [ ] added corresponding documentation in the headers

##### Automatic note to reviewers

Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-following,-reviewing-and-playing-with-a-PR#checklist-for-reviewing-a-pr) and put a milestone if possible.

PR #1256, opened 2024-07-05
Trying to generalize mfun, updated

### LINKED ISSUE #662: generalize `{mfun _ >-> _}`
https://github.com/math-comp/analysis/blob/bfa6d4bc2fa548644b87769267ea49f523a21858/theories/lebesgue_integral.v#L68

while preserving the ring structure of simple functions @CohenCyril

hoheinzollern:

fixes #662 

closes #672 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -90,21 +90,22 @@
 	into local `Let`'s
 
 - in `lebesgue_integral.v`:
-  + lemma `cst_mfun_subproof` now a `Let`
+  + structure `SimpleFun` now inside a module `HBSimple`
+  + structure `NonNegSimpleFun` now inside a module `HBNNSimple`
+  + lemma `cst_nnfun_subproof` has now a different statement
+  + lemma `indic_nnfun_subproof` has now a different statement
+
 
 ### Renamed
 
 ### Generalized
 
 - in `lebesgue_integral.v`:
-  + generalized from `sigmaRingType`/`realType` to `sigmaRingType`
+  + generalized from `sigmaRingType`/`realType` to `sigmaRingType`/`sigmaRingType`
     * mixin `isMeasurableFun`
     * structure `MeasurableFun`
 	* definition `mfun`
     * lemmas `mfun_rect`, `mfun_valP`, `mfuneqP`
-  + generalized from `measurableType`/`realType` to `sigmaRingType`/`realType`
-    * lemmas `cst_mfun_subproof`, `mfun_cst`
-    * definition `cst_mfun`
 
 ### Deprecated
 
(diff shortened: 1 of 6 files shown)


affeldt-aist:
@CohenCyril this seems to actually generalize `isMeasurableFun` that specialized to `realType`, we had to get rid a `cst` instantiation that wouldn't go through but does not seem useful, maybe it was there to solve an inference problem that has been solved in the meantime?

affeldt-aist:
the remaining problem is to generalize the instance of `cst` which have not been able to figure out yet,
but already in the current form this PR is an improvement we think

```coq
Context {d} {aT : sigmaRingType d} {rT : realType}.

Let cst_mfun_subproof x : @measurable_fun d _ aT rT [set: aT] (cst x).
Proof. by []. Qed.

HB.instance Definition _ x := isMeasurableFun.Build d _ aT rT (cst x)
  (@cst_mfun_subproof x).
```

CohenCyril:
> the remaining problem is to generalize the instance of `cst` which have not been able to figure out yet,
> but already in the current form this PR is an improvement we think
> 
> ```coq
> Context {d} {aT : sigmaRingType d} {rT : realType}.
> 
> Let cst_mfun_subproof x : @measurable_fun d _ aT rT [set: aT] (cst x).
> Proof. by []. Qed.
> 
> HB.instance Definition _ x := isMeasurableFun.Build d _ aT rT (cst x)
>   (@cst_mfun_subproof x).
> ```

What is the issue exactly? The preimage of any set by a constant function should be measurable (because either set0 or setT)

affeldt-aist:
> What is the issue exactly? The preimage of any set by a constant function should be measurable (because either set0 or setT)

Of course generalizing `cst_mfun_subproof` is perfectly fine (it is even a hint).
The problem occurs when doing `HB.instance`. There is a type inference problem.
Inspection of the type error seems to indicate that HB really wants `cst` to be simple function,
that returns real numbers. In particular, if you move this part of the code earlier in the file
it will type check but then this is the declaration of `cst` as a simple function that causes a problem.
We haven't been able to figure it out.

affeldt-aist:
@CohenCyril The last commit is generalizing the `mfun` instance of `cst` to a codomain of type `measurableType` instead of `realType` (at least in `lebesgue_integral.v`). To prevent HB to be too eager (and infer that the codomain should be at the same `measurableType` and `realType` which seems to cause the problem---maybe we can play with `measurableTypeR` to avoid that but we were not successful at that in the past, let's maybe reconsider that but later), the structure for simple functions has been put into a module that is opened when needed. The problem is that when we do that and need to infer more elaborate types for `cst` we also need to regenerate fimfun/mfun/nnfun instances locally. This is dirty because it is only a try, but if it makes sense we can clean that up. What do you think?

CohenCyril:
> Member

This is what I advocated, but I think it is a bit too hackish, we must fix hb instead.

affeldt-aist:
> This is what I advocated, but I think it is a bit too hackish, we must fix hb instead.

Great! In the meantime, should we try to merge this hack (in a more proper form and clearly marked as a hack of course)? In particular, the generalization of `mfun` is likely to be useful for the Giry monad and for probability theory.

affeldt-aist:
CI is ok, the kludge is well-localized and documented, plus this will serve as a test bed for the forthcoming coding sprint, I can maybe interpret your thumb up @CohenCyril as an OK to merge?

PR #1266, opened 2024-07-23
integration by parts

affeldt-aist:

~~preliminary~~ usable version of integration by parts

(and "within continuous" version of `continuous_FTC2`)

NB: we've been trying to "weaken" the hypotheses of lemmas in `ftc.v` to avoid requiring integrability over the whole set in the hope to ease their applications (this seems to be confirmed by the use of the new version of `continuous_FTC1` to prove change-of-variables lemma)

~~based on PR #1246~~ (merged)

FYI @IshiguroYoshihiro 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -542,3 +542,46 @@ exact: integral_fune_fin_num.
 Unshelve. all: by end_near. Qed.
 
 End corollary_FTC1.
+
+Section integration_by_parts.
+Context {R : realType}.
+Notation mu := lebesgue_measure.
+Local Open Scope ereal_scope.
+Implicit Types (F G f g : R -> R) (a b : R).
+
+Lemma continuous_integration_by_parts F G f g a b :
+    (a < b)%R ->
+    {in `[a, b], continuous f} -> {in `[a, b], continuous F} ->
+    derivable_oo_continuous_bnd F a b ->
+    {in `]a, b[, F^`() =1 f} ->
+    {in `[a, b], continuous g} -> {in `[a, b], continuous G} ->
+    derivable_oo_continuous_bnd G a b ->
+    {in `]a, b[, G^`() =1 g} ->
+  (\int[mu]_(x in `[a, b]) (F x * g x)%:E = (F b * G b - F a * G a)%:E -
+   \int[mu]_(x in `[a, b]) (f x * G x)%:E).
+Proof.
+move=> ab cf cF Fab Ff cg cG Gab Gg.
+have cfg : {in `[a, b], continuous (f * G + F * g)%R}.
+  move=> z zab; apply: continuousD; apply: continuousM;
+    [exact: cf|exact: cG|exact: cF|exact: cg].
+have FGab : derivable_oo_continuous_bnd (F * G)%R a b.
+  move: Fab Gab => /= [abF FFa FFb] [abG GGa GGb];split; [|exact:cvgM..].
+  by move=> z zab; apply: derivableM; [exact: abF|exact: abG].
+have FGfg : {in `]a, b[, (F * G)^`() =1 f * G + F * g}%R.
+  move: Fab Gab => /= [abF FFa FFb] [abG GGa GGb] z zba.
+  rewrite derive1E deriveM; [|exact: abF|exact: abG].
+  by rewrite -derive1E Gg// -derive1E Ff// addrC (mulrC f).
+have := continuous_FTC2 ab cfg FGab FGfg; rewrite -EFinB => <-.
+under [X in _ = X - _]eq_integral do rewrite EFinD.
+have ? : mu.-integrable `[a, b] (fun x => ((f * G) x)%:E).
+  apply: continuous_compact_integrable => //; first exact: segment_compact.
+  apply: continuous_in_subspaceT => z /[!inE] zab.
+  by apply: continuousM; [exact: cf|exact: cG].
+rewrite /= integralD//=.
+- by rewrite addeAC subee ?add0e// integral_fune_fin_num.
+- apply: continuous_compact_integrable => //; first exact: segment_compact.
+  apply: continuous_in_subspaceT => z /[!inE] zab.
+  by apply: continuousM; [exact: cF|exact: cg].
+Qed.
+
+End integration_by_parts.
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -29,20 +29,59 @@ Import numFieldTopology.Exports.
 Local Open Scope classical_set_scope.
 Local Open Scope ring_scope.
 
+(* TODO: move to normedtype.v? *)
+Lemma nbhs_right_lt_lt {R : realType} (x y : R) :
+  (y < x)%R -> \forall z \near nbhs y^'+, (z < x)%R.
+Proof.
+move=> yx.
+exists (x - y)%R => /=; first by rewrite subr_gt0.
+move=> z/= /[swap] yz.
+by rewrite ltr0_norm ?subr_lt0// opprB ltrBlDl addrC subrK.
+Qed.
+
+(* TODO: move to normedtype.v? *)
+Lemma nbhs_right_lt_le {R : realType} (x y : R) :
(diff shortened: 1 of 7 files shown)


theories/ftc.v
```diff
@@ -230,3 +232,350 @@ by split; [exact: dfx|rewrite f'xE].
...
+
+End parameterized_integral_continuous.
+
+Section corollary_FTC1.
+Context {R : realType}.
+Notation mu := lebesgue_measure.
+Local Open Scope ereal_scope.
+Implicit Types (f : R -> R) (a b : R).
+
+Corollary continuous_FTC2 f F a b : (a < b)%R -> {in `[a, b], continuous f} ->
```
zstone1, line 422: Can we make this `{within [a,b], continuous f}`? This has the usual annoying boundary conditions.

theories/ftc.v
```diff
@@ -230,3 +232,350 @@ by split; [exact: dfx|rewrite f'xE].
...
+
+Section integration_by_parts.
+Context {R : realType}.
+Notation mu := lebesgue_measure.
+Local Open Scope ereal_scope.
+Implicit Types (F G f g : R -> R) (a b : R).
+
+Lemma continuous_integration_by_parts F G f g a b :
+    (a < b)%R ->
+    {in `[a, b], continuous f} -> {in `[a, b], continuous F} ->
```
zstone1, line 548: Usual boundary conditions complaints

zstone1 (commented):
A few annoying boundary condition issues. If you want, we can weaken those conditions via tietze, or other limiting approaches if the proofs are hard to edit. If you want some guidance on that I'm happy to provide.

theories/ftc.v
affeldt-aist: This is done in commit https://github.com/math-comp/analysis/pull/1266/commits/b899596a734d15a48425261a99b266989a735948

theories/ftc.v
affeldt-aist: This is also done in commit https://github.com/math-comp/analysis/pull/1266/commits/b899596a734d15a48425261a99b266989a735948

affeldt-aist:
The last commit weakens the hypotheses from `{in _, continuous _}` to `{within _, continuous _}` of the corollary of FTC1 ("FTC2 for continuous functions") and of integration by parts (newly introduced). I went through a weakening of the hypotheses of FTC1 theorems for that purpose: moving from integrability over the whole set to local integrability + integrability over a family of intervals possibly open-ended on the left (the two are redundant when dealing with bounded intervals but not when the interval on which we want to integrate is say `]-oo, x]`). By the way, the resulting theorems are used in another branch to prove properties of the beta distribution, so there are usable in practice.

theories/ftc.v
```diff
@@ -35,14 +35,14 @@ Notation mu := (@lebesgue_measure R).
 Local Open Scope ereal_scope.
 Implicit Types (f : R -> R) (a : itv_bound R).
 
-Let FTC0 f a : mu.-integrable setT (EFin \o f) ->
-  let F x := (\int[mu]_(t in [set` Interval a (BRight x)]) f t)%R in
-  forall x, a < BRight x -> lebesgue_pt f x ->
+Let FTC0 f a x (u : R) : (x < u)%R ->
```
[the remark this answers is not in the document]
affeldt-aist: Let me try...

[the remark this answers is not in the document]
affeldt-aist: You are right, I have visibly been confused, fixing the PR right now.

theories/lebesgue_integral.v
```diff
@@ -6826,13 +6864,16 @@ Variable E : R -> (set R)^nat.
...
 Local Notation mu := lebesgue_measure.
 
-Lemma nice_lebesgue_differentiation (f : R -> R) :
-  locally_integrable setT f ->
-  forall x, lebesgue_pt f x ->
+#[local] Hint Extern 0 (measurable (closed_ball _ _)) =>
+  solve [apply: measurable_closed_ball] : core.
+
+Lemma nice_lebesgue_differentiation (f : R -> R) (x : R) :
+  (\forall r \near 0^'+, locally_integrable setT (f \_ (closed_ball x r))) ->
```
zstone1, line 6871: Since R is locally compact, this is the same as 
```
(\forall r \near 0^'+, integrable setT (f \_ (closed_ball x r)))
```
There is also the powerser_filter_from thing which let's you state it as
```
(\forall U \near powerset_filter_from (nbhs x),, integrable U f)
```
Either is fine.

theories/ftc.v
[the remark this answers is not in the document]
zstone1: Ok, glad we could simplify that. My next question is, if you leave the original `FTC0` alone, and add a this new statement as `FTC0_restricted`, can we simplify the proof by just applying `FTC0` with the restricted function?

theories/ftc.v
```diff
@@ -106,137 +165,279 @@ apply: cvg_at_right_left_dnbhs.
...
+  apply: (@subset_trans _ (`](x - e)%R, u])) => //.
+    apply: subset_itvl; rewrite bnd_simp.
+    rewrite -lerBrDl.
+    by near: e; apply: nbhs_right_ltW; rewrite subr_gt0.
+  apply: subset_itvr; rewrite bnd_simp lerBrDr -lerBrDl.
+  by near: e; apply: nbhs_right_ltW; rewrite subr_gt0.
 Unshelve. all: by end_near. Qed.
 
+Corollary continuous_FTC1_closed f (a x : R) (u : R) : (x < u)%R ->
+  locally_integrable setT (f \_ [set` `[a, u]]) ->
```
zstone1, line 427: Locally intrgrable on a compact set is the same as just integrable. Any reason to prefer this formulatjon?

theories/ftc.v
```diff
@@ -542,3 +757,81 @@ exact: integral_fune_fin_num.
...
 End corollary_FTC1.
+
+Section integration_by_parts.
+Context {R : realType}.
+Notation mu := lebesgue_measure.
+Local Open Scope ereal_scope.
+Implicit Types (F G f g : R -> R) (a b : R).
+
+Lemma integration_by_parts F G f g a b : (a < b)%R ->
+    {within `[a, b], continuous f} -> {within `[a, b], continuous F} ->
```
zstone1, line 768: Does `derivable_oo_continuous_bnd` imply `continuous within [a,b]`?

zstone1 (commented):
A few more thoughts on ways to simplify the conditions does these lemmas.

theories/ftc.v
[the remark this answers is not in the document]
affeldt-aist: I think that I have a script that does that, should be pushed by tomorrow.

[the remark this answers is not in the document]
affeldt-aist: it seems so: see a078951ecc88a810949a398b8a5a73ddeb9c7836

theories/lebesgue_integral.v
affeldt-aist: I have reverted to this version:
```coq
Lemma nice_lebesgue_differentiation (f : R -> R) :
  locally_integrable setT f -> forall x, lebesgue_pt f x ->
  (fine (mu (E x n)))^-1%:E * \int[mu]_(y in E x n) (f y)%:E
    @[n --> \oo] --> (f x)%:E.
```

theories/ftc.v
affeldt-aist: Let's use `integrable` which should be better instrumented and thus more user-friendly.

theories/ftc.v
affeldt-aist: Yes, this is `derivable_oo_continuous_bnd_within`.

zstone1:
Ah, this looks much cleaner now! Thanks for all the updates. There's always more proof linting to do, but I'm not seeing any more structural issues. I'm fine to merge if/when you're happy with the state of the proof scripts.

affeldt-aist:
I did some more linting and I don't mind if you squash and merge.
There is surely room for improvement but we might come back to
it soon with integration by substitution in the pipeline anyway.

PR #1286, opened 2024-08-07
rat_in_itvoo need not be restricted to realTypes

t6s:
`reals.rat_in_itvoo` holds also for archimedean fields.


affeldt-aist (approved):
Well spotted!

affeldt-aist:
Could you mention the generalization in CHANGELOG_UNRELEASED?

PR #1280, opened 2024-08-06
Add the lemma `not_near_ninftyP` in `normedtype.v`

Yosuke-Ito-345:
##### Motivation for this change

@affeldt-aist 
Add an infinite version of the lemma `not_near_at_rightP` in `normedtype.v`.
Tell me if I should write a change log in `CHANGELOG_UNRELEASED.md`.

##### Checklist

- [ ] added corresponding entries in `CHANGELOG_UNRELEASED.md`
- [ ] added corresponding documentation in the headers

Reference: [How to document](https://github.com/math-comp/math-comp/wiki/How-to-document)

##### Reminder to reviewers

- Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs)
- Put a milestone if possible
- Check labels

Issue #1273, opened 2024-07-27
Should we make `not_near_at_rightP` an equivalence?


Yosuke-Ito-345:
In the theory `normedtype.v`, we have the lemma `not_near_at_rightP`.
The name contains the character `P`, but it states one-direct implication, not the equivalence.
For me, this seems contradictory to the naming convention stated in
https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md
Should we add the other direction of `not_near_at_rightP` and make it an equivalence?
(This is also the case with `not_near_at_leftP`.)

affeldt-aist:
Fixed by PR #1291

affeldt-aist:
I have closed this issue by mistake (Ito-san: thanks for noticing).

Yosuke-Ito-345:
In the `Section at_left_right` in `normedtype.v`, the `Variable R` is assumed to be `numFieldType`.
Isn't it more rational to assume `R` to be `realDomainType`?
I think left and right have meanings only when the type is totally ordered.
I suggest this alternative because I would like to freely use the following lemma
```
ler_norm: forall [R : realDomainType] (x : R), x <= `|x|
```

Yosuke-Ito-345:
@affeldt-aist 
Sorry to bother you, but may I ask for your opinion?

affeldt-aist:
I am not sure about the consequences.
Note that there is the lemma `real_ler_norm` which is also available;
it requires a weaker type but maybe it amounts to be the same because of the added condition.
Maybe you could show the code you're working on?
Let me also ping @CohenCyril who designed the hierarchy of numerical types.

Yosuke-Ito-345:
@affeldt-aist 
Thank you for the reply.
Here is the code I am working on in `Section at_left_right` where `Variable R : numFieldType`. (Sorry for not yet cleaned.)
```
(* generalization of "Decidable.not_and_iff" *)
Lemma imp_and_iff (A B P : Prop) : (A -> B -> P) <-> (A /\ B -> P).
Proof.
by split => [?[]//|HABP ??]; apply: HABP.
Qed.

Lemma forall_posnumP (P : R -> Prop) :
  (forall x : {posnum R}, P x%:num) <-> (forall y : R, 0 < y -> P y).
Proof.
Admitted.

Lemma not_near_at_rightP (p : R) (P : pred R) :
  ~ (\forall x \near p^'+, P x) <->
[...]

PR #1291, opened 2024-08-11
Add lemmas `not_near_inftyP` and `not_near_ninftyP` in `normedtype.v`

Yosuke-Ito-345:

@affeldt-aist @hoheinzollern
Add the infinite versions of the lemmas `not_near_at_rightP` and `not_near_at_leftP` in `normedtype.v`.

closes issue #1280

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 ### Added
+- in `normedtype.v`:
+  + lemmas `not_near_inftyP` `not_near_ninftyP`
 
 ### Changed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,7 +4,7 @@
 
 ### Added
 - in `normedtype.v`:
-  + lemmas `not_near_inftyP` `not_near_ninftyP`
+  + lemmas `not_near_inftyP`, `not_near_ninftyP`
 
 - in `topology.v`:
   + lemma `filterN`
@@ -13,6 +13,9 @@
   + lemma `ninftyN`
 
 ### Changed
+
+- in `normedtype.v`:
+  + remove superflous parameters in lemmas `not_near_at_rightP` and `not_near_at_leftP`
 
 ### Renamed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -6,6 +6,12 @@
 - in `normedtype.v`:
   + lemmas `not_near_inftyP` `not_near_ninftyP`
 
+- in `topology.v`:
+  + lemma `filterN`
+
+- in `normedtype.v`:
+  + lemma `ninftyN`
+
 ### Changed
 
 ### Renamed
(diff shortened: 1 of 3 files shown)


affeldt-aist:
This is not a review but rather a comment.
I allowed myself to push a commit https://github.com/math-comp/analysis/pull/1291/commits/e2f3e0b1c1f36439e1158ee70909d1abaa8d273e to your PR.
When you mentioned `not_near_at_rightP` and `not_near_at_leftP` is issue https://github.com/math-comp/analysis/issues/1273 I thought that one should be derived from the other using some potential useful intermediate lemma. This is indeed possible.
Similarly, for the lemmas `not_near_inftyP` and `not_near_ninftyP`: one can be derived from the other, though the proofs are globally not shorter.
I wonder whether a few more lemmas could not benefit from such factorizations.

affeldt-aist:
Maybe we do not need to distinguish `P` and `f` in these lemmas?

Yosuke-Ito-345:
Wonderful!
Thank you for adding the lemma `filterN`. This should be useful to my formalization of actuarial mathematics, too.
I also agree with you that `P` and `f` need not be distinguished. I wonder why I didn't notice that.  \\(^o^)/
I will reflect your suggestions in this pull request.

By the way, would it be better to add something like this?
```
Lemma iff_forall T (U V : T -> Prop) :
  (forall x : T, U x <-> V x) -> ((forall x, U x) <-> (forall x, V x)).
```
It shortens the proof of `not_near_inftyP`, but I am not sure whether it is compatible with the mathcomp style.

hoheinzollern:
@Yosuke-Ito-345 not sure where you would want to use `iff_forall`, but have you checked out `eq_forall`?

```coq
eq_forall :
  forall [T : Type] [U V : T -> Prop],
    (forall x : T, U x = V x) -> (forall x : T, U x) = (forall x : T, V x)
```

Yosuke-Ito-345:
@hoheinzollern 
Yes, I know `eq_forall`. But  we have to `rewrite -propeqE` before `eq_forall`, and `rewrite propeqE` after `eq_forall`.
```
Lemma not_near_inftyP (P : pred R) :
  ~ (\forall x \near +oo, P x) <->
    forall M : R, M \is Num.real -> exists2 x, M < x & ~ P x.
Proof.
rewrite nearE -forallNP -propeqE; apply: eq_forall => M.
by rewrite propeqE -implypN existsPNP.
Qed.
```
This feels inefficient to me. If we had `iff_forall`, the proof could be as follows.
```
by rewrite nearE -forallNP; apply: iff_forall => M; rewrite -implypN existsPNP.
```
I would not propose `iff_forall` if the following proof worked...
```
rewrite nearE -forallNP -propeqE; apply: eq_forall => M.
Fail rewrite -implypN existsPNP.
```

Yosuke-Ito-345:
@affeldt-aist @hoheinzollern 
I don't persist in adding `iff_forall`, so could you merge this pull request if there is no problem in the current code?
(I think I don't have the right to merge this pull request.)

hoheinzollern:
Ok, the PR looks good to me, I don't think there's a strong reason to add the `iff_forall`, so I would keep it as is.

affeldt-aist (approved):
Looks ok to me.

PR #1293, opened 2024-08-13
lemmas about Lebesgue's outer measure

affeldt-aist:
Four lemmas about the Lebesgue measure.
The second one is actually the most interesting because we use to prove
an important corollary of the Vitali's theorem (PR to come)

- `outer_measure_open_itv_cover`:
  definition of the outer measure in terms of covers of open intervals
  (proposition 20 in Li)
- `outer_measure_open`:
  for a set A with bounded outer measure and an epsilon,
  there is an open that contains A and whose measure is
  less than the outer measure A + epsilon,
  (proposition 21 1) in Li)
- `outer_measure_Gdelta`:
  for a set A with bounded outer measure, there exists a G-delta sets
  that contains A and whose measure is the same as A
  (proposition 21 2) in Li)
- `negligible_outer_measure`:
  a set N is negligible for the lebesgue measure iff its outer measure is 0
  (corollary 22 in Li)

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/lebesgue_measure.v
+++ b/theories/lebesgue_measure.v
@@ -2055,7 +2055,7 @@ Definition open_itv_cover A := [set F : (set R)^nat |
 
 Let l := (@wlength R idfun).
 
-Lemma outer_measure_open_itv_cover A : (l^*)%mu A =
+Lemma outer_measure_open_itv_cover A : (l^* A)%mu =
   ereal_inf [set \sum_(k <oo) l (F k) | F in open_itv_cover A].
 Proof.
 apply/eqP; rewrite eq_le; apply/andP; split.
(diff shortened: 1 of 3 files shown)


theories/lebesgue_measure.v
```diff
@@ -2041,6 +2041,138 @@ Notation emeasurable_fun_funeneg := measurable_funeneg (only parsing).
...
+  rewrite [X in eseries X](_ : _ = (fun k => (e / (2 ^+ (k.+2))%:R)%:E)); last first.
+    by apply/funext => n; rewrite addn2 natrX.
+  move/cvg_lim => <-//; apply: lee_nneseries => //.
+  - by move=> n _; rewrite lee_fin divr_ge0// ltW.
+  - by move=> n _; rewrite lee_fin -natrX.
+Qed.
+
+Definition mu := (@lebesgue_measure R).
+
+Lemma outer_measure_open A : (l^* A)%mu < +oo ->
```
zstone1, line 2118: Any reason to phrase it as
``` 
(l^*)%mu A = ereal_inf [set (l^* A)%mu | open U /\ A `<=` U]
```
so it is more compatible with `outer_measure_open_itv_cover`'s phrasing? It also has the benefit of being true in the unbounded case.

zstone1 (commented):
Look pretty good, just one comment on a slight generalization.

theories/lebesgue_measure.v
affeldt-aist: ``` (l^*)%mu U = ereal_inf [set (l^* A)%mu | open U /\ A `<=` U] ``` (note the typo fix)
has been added as a lemma to the last commit but I am not sure how to exploit it

in any case, it suggested that `outer_measure_Gdelta` and `outer_measure_open_le` (new name) do not need the boundedness hypothesis so it has been removed

zstone1 (approved):

PR #1294, opened 2024-08-13
change of variables by nondecreasing/nonincreasing function

IshiguroYoshihiro:
add lemmas for integration by substitution with increasing/decreasing function.

~~some intermediate lemmas are admitted but proved in #1327~~

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -784,6 +784,70 @@ Unshelve. all: end_near. Qed.
 
 End integration_by_substitution_preliminaries.
 
+Section within_continuous_derive.
+Context {R : numFieldType} {V W : normedModType R}.
+
+Definition right_derive (f : V -> W) a v :=
+  lim ((fun h => h^-1 *: ((f \o shift a) (h *: v) - f a)) @ 0^'+).
+Definition left_derive (f : V -> W) a v :=
+  lim ((fun h => h^-1 *: ((f \o shift a) (h *: v) - f a)) @ 0^'-).
+
+End within_continuous_derive.
+
+(* PR *)
+Lemma measurable_fun_set1 d d' (T : measurableType d) (U : measurableType d')
+   (a : T) (f : T -> U) :
+d.-measurable (set1 a) -> measurable_fun (set1 a) f.
+Proof.
+Admitted.
+
+(* PR *)
+Lemma in_nearW (R : realFieldType) (P : R -> Prop) (S : set R) :
+  @open R^o S ->
+ {in S, forall x, P x} ->
+  forall x, x \in S -> \forall y \near x, P y.
+Proof.
+move=> oS HP z.
+rewrite inE => Sz.
+have [e/= e0 He] := (open_subball oS Sz).
+exists (e / 2) => //=.
+  by rewrite -(add0r e) midf_lt.
+move=> t/= zte.
+apply: HP.
+move: He Sz zte.
+have [<- |zt He Sz zte] := eqVneq z t; first by rewrite inE.
+rewrite inE.
+apply: (He (e / 2)) => //=.
+rewrite sub0r normrN ger0_norm.
+    by rewrite -{1}(add0r e) midf_lt.
+  by rewrite -(add0r e) midf_le// ltW.
+by rewrite -(add0r e) midf_lt.
+Qed.
+
+Section integral_itv.
+Local Open Scope ereal_scope.
+Context {R : realType}.
+
+Local Notation mu := (@lebesgue_measure R).
+
+(* PR *)
+Lemma integral_itv_bndo_bndc' (a : itv_bound R) (r : R) (f : R -> R) :
+  measurable_fun [set` Interval a (BLeft r)] f ->
+   \int[mu]_(x in [set` Interval a (BLeft r)]) (f x)%:E =
+   \int[mu]_(x in [set` Interval a (BRight r)]) (f x)%:E.
+Proof.
+Admitted.
+
+(* PR *)
+Lemma integral_itv_obnd_cbnd' (r : R) (b : itv_bound R) (f : R -> R) :
+  measurable_fun [set` Interval (BRight r) b] f ->
+   \int[mu]_(x in [set` Interval (BRight r) b]) (f x)%:E =
+   \int[mu]_(x in [set` Interval (BLeft r) b]) (f x)%:E.
+Proof.
+Admitted.
+
+End integral_itv.
+
 Section integration_by_substitution.
 Local Open Scope ereal_scope.
 Context {R : realType}.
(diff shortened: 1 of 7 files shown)


theories/ftc.v
```diff
@@ -542,3 +769,386 @@ exact: integral_fune_fin_num.
...
+  {in `[a, b] &, {homo F : x y /~ (x < y)%R}} ->
+  F @` `]a, b[ `<=` `]F b, F a[.
+Proof.
+move=> ab iF ? [x /= xab <-].
+move: xab; rewrite !in_itv/= => /andP[ax xb].
+by apply/andP; split; apply: iF; rewrite // in_itv/= ?lexx !ltW.
+Qed.
+
+Lemma increasing_cvg_at_right_comp F G a b (l : R) : (a < b)%R ->
+  {in `[a, b] &, {homo F : x y / (x < y)%R}} ->
```
zstone1, line 875: Can we weaken this to `[a, b[`? I doubt it really matters, but it seems we might as well.

theories/ftc.v
```diff
@@ -542,3 +769,386 @@ exact: integral_fune_fin_num.
...
+
+Section integration_by_substitution.
+Local Open Scope ereal_scope.
+Context {R : realType}.
+Notation mu := lebesgue_measure.
+Implicit Types (F G f : R -> R) (a b : R).
+
+Lemma increasing_change F G a b : (a < b)%R ->
+  {in `[a, b] &, {homo F : x y / (x < y)%R}} ->
+  {within `[a, b], continuous F^`()} ->
```
zstone1, line 963: The joy of boundary condition never ceases. The term ``F^`()`` is defined as 
```
Definition derive1 V (f : R -> V) (a : R) := lim ((fun h => h^-1 *: (f (h + a) - f a)) @ 0^').
```
which is taking the limit near 0 in `R`. Which means ``F^`() `` depends on values outside of `[a,b]`. That's bad news if you want to consider `F\_[a,b]`. which will not have a (full) derivative at `a`.

Several ways to fix this 
- The "right" way to fix this is have `derive1` consider `a` under different topologies. Shouldn't be too hard, but it's still kinda invasive. Maybe a ticket for this?
- For a less invasive way, you can introduce another helper function 
``` 
{within `[a, b], continuous F1} ->
{in `]a.b[, F1 = F^`()}
```
This forces ``F^`()`` to have limits at its endpoints. That is ``F1 a = lim_(x -->a^+) F`^() x`` which is what we want.
- We can also do `{within `]a.b[, continuous F^`()}` + ``cvg (F`^() @ a^+)``

I don't have a strong preference on which approach you take. It's not clear to me how much harder this will make the proof...

theories/ftc.v
IshiguroYoshihiro: Indeed. Thank you.
Similarly, we can weaken nearby lemmas. 
https://github.com/math-comp/analysis/blob/67af2d53f353edb28a8f5650ea84eab84032ff6f/theories/ftc.v#L899-L900
https://github.com/math-comp/analysis/blob/67af2d53f353edb28a8f5650ea84eab84032ff6f/theories/ftc.v#L917-L918
https://github.com/math-comp/analysis/blob/67af2d53f353edb28a8f5650ea84eab84032ff6f/theories/ftc.v#L935-L936

theories/ftc.v
IshiguroYoshihiro: I had thought that ``{within `[a, b], continuous F^`()} `` as `` {within `[a, b], continuous F1} /\ {in `]a, b[, F1 = F^`()}`, but I hesitate to add a new function in the statement and make less convenient.
So, I think the first solution looks better. 
I will try to define right/left derivation
```
Definition right_derive1 V (f : R -> V) (a : R) :=
   lim ((fun h => h^-1 *: (f (h + a) - f a)) @ 0^'+).
Definition left_derive1 V (f : R -> V) (a : R) :=
   lim ((fun h => h^-1 *: (f (h + a) - f a)) @ 0^'-).
```
and to change `` {within `[a, b], continuous F^`()} `` to
`` {in `]a, b[, continuous F^`()} `` and `right_derive1 F` is right continuous at `a` and `left_derive1 F` is left continuous at `b`.
Thank you for advice!

zstone1: Theory wise, I think you'll end up needing to prove something along the lines of `F^`() @a^+ --> L -> right_derive1 F a /\ F^`+() a = L` anyway (see https://math.stackexchange.com/questions/3301610/proving-differentiability-at-a-point-if-limit-of-derivative-exists-at-that-point). But engineering-wise I'm happy with whatever will lead you to the nicest theorem statement. Good luck!

IshiguroYoshihiro: I noticed that the necessary parts of boundary condition of ``F^`()`` is that this converged, so I have changed hypothesis to ``cvg (F^`() @ a^'+/@ b^'-)``.
This seems easier to use as we don't necessarily have to consider one-sided derivatives.

theories/ftc.v
```diff
@@ -681,3 +681,557 @@ exact: (derivable_oo_continuous_bnd_within Gab).
...
+
+(* PR *)
+Lemma measurable_fun_set1 d d' (T : measurableType d) (U : measurableType d')
+   (a : T) (f : T -> U) :
+d.-measurable (set1 a) -> measurable_fun (set1 a) f.
+Proof.
+Admitted.
+
+(* PR *)
+Lemma in_nearW (R : realFieldType) (P : R -> Prop) (S : set R) :
```
zstone1, line 807: This seems like it should be generalized to any topological space, and moved to `topology.v`

theories/ftc.v
```diff
@@ -681,3 +681,557 @@ exact: (derivable_oo_continuous_bnd_within Gab).
...
+Lemma decreasing_change F G a b : (a < b)%R ->
+  {in `[a, b] &, {homo F : x y /~ (x < y)%R}} ->
+  {in `]a, b[, continuous F^`()} ->
+  cvg ((F^`() : R -> R) x @[x --> a^'+]) ->
+  cvg ((F^`() : R -> R) x @[x --> b^'-]) ->
+  derivable_oo_continuous_bnd F a b ->
+ {within `[F b, F a], continuous G} ->
+  \int[mu]_(x in `[F b, F a]) (G x)%:E =
+  \int[mu]_(x in `[a, b]) (((G \o F) * (- F^`())) x)%:E.
+Proof.
```
zstone1, line 1029: I would hope this proof could be done somewhat easily as a call to `increasing_change` with `-F`. You've got a proof here, so it's not a big deal. But if it saves you 200 lines of proof that would be pretty nice.

zstone1 (commented):
Looks much better with the fixed boundary conditions. A couple thoughts on potential improvements to the proofs, but nothing serious. Qed is Qed, after all. So I'm happy with it

theories/ftc.v
IshiguroYoshihiro: I have tried some times to prove `decreasing_change` by using `increasing_change` but not successed yet. 
The boundaries of intervals which integral over is fliped, so I feel it is hard to prove from increasing_change. (probably we can prove increasing_change from decreasing_change. It's worth a try.)

zstone1: Perhaps `at_rightN` and `at_leftN` make it feasible for the endpoints? And if there's a missing lemma about subspaces you need I'm happy to provide it

theories/ftc.v
IshiguroYoshihiro: Indeed.
I generalized and made a PR (#1332 ).
Thank you!

theories/ftc.v
affeldt-aist: The last commit provides a proof of the "decrease case" derived from the "increase case". It is still a bit long (62 lines) but this is just about checking conditions.

affeldt-aist:
A couple of lemmas to move to more appropriate locations and we should be done.

affeldt-aist:
Still feels a bit long but at least all comments should be addressed.

PR #1303, opened 2024-08-24
fixes #1301

### LINKED ISSUE #1301: documentation of `\;` needs fixing
https://github.com/math-comp/analysis/blob/7171aaf77ac6cadf8177c3478fbf16e5ba79c822/classical/classical_sets.v#L209

Jean-Philippe Chancelier on Zulip on 2024-08-22:
```
there is also a little inconsistency between the comments at the begining of the file and the code
for relation composition. The composition of relation is described as
A \; B == [set x | exists z, A (x.1, z) & B (z, x.2)]
in the comments which is correct but implemented as
Notation "B \; A" := ([set xy | exists2 z, A (xy.1, z) & B (z, xy.2)]) : classical_set_scope.
"B \; A" should be "A \; B"
```

affeldt-aist:
fixes #1301

- add relation_scope

While fixing the documentation for the notation `\;`, I realized that the notation for the inverse
of a relation was locally defined in three files. Since it seems to be used always along with `\;`,
I added a `relation_scope` that can be open in sections dealing with uniform spaces. I think
it is quite natural as it does not cause clashes with the already overloaded `^-1` notation and
also because it reduces the clutter in statements.

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/normedtype.v
+++ b/theories/normedtype.v
@@ -3537,12 +3537,11 @@ Lemma uniform_separatorW {T : uniformType} (A B : set T) :
 Proof. by case=> E entE AB0; exists (Uniform.class T), E; split => // ?. Qed.
 
 Section Urysohn.
+Local Open Scope relation_scope.
 Context {T : topologicalType}.
 Hypothesis normalT : normal_space T.
 Section normal_uniform_separators.
 Context (A : set T).
-
-Local Notation "A ^-1" := [set xy | A (xy.2, xy.1)] : classical_set_scope.
 
 (* Urysohn's lemma guarantees a continuous function : T -> R
    where "f @` A = [set 0]" and "f @` B = [set 1]".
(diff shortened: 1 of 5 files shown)


theories/normedtype.v
```diff
@@ -3622,10 +3621,10 @@ move/(_ (globally [set fg | fg.1 = fg.2])); apply; split.
 exact: ury_base_refl.
 Qed.
 
-Local Lemma set_prod_invK (K : set (T * T)) : (K^-1^-1)%classic = K.
+Local Lemma set_prod_invK (K : set (T * T)) : K^-1^-1 = K.
```
zstone1, line 3624: If we're cleaning things up, shouldn't this live in classical_set rather than as a local lemma hwre?

zstone1 (approved):
Quick suggestion to move a lemma around. But even as it stands this change is a strict improvement to the code, so very happy

theories/normedtype.v
affeldt-aist: Yes, I missed it. Thanks.

affeldt-aist: I also added a definition for the diagonal, since the notation was repeated several times.

PR #1314, opened 2024-09-08
fixes #1313 and moves Egorov

### LINKED ISSUE #1313: wrong documentation
https://github.com/math-comp/analysis/blob/576f7bf049f94f81725e5e4902017dca0271abff/theories/lebesgue_measure.v#L17

this is not Vitali's lemma in this file but Vitali's theorem

affeldt-aist:

fixes #1313 

This PR also shuffles the contents of the file `lebesgue_measure.v`
so that lemmas that do not depend on Lebesgue's measure appear
before it (the Egorov theorem in particular).

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/lebesgue_integral.v
+++ b/theories/lebesgue_integral.v
@@ -6051,21 +6051,16 @@ Let norm1 D f := \int[mu]_(y in D) `|(f y)%:E|.
 
 Lemma maximal_inequality f c :
   locally_integrable setT f -> (0 < c)%R ->
-  mu [set x | HL f x > c%:E] <= (3%:R / c)%:E * norm1 setT f.
+  mu [set x | HL f x > c%:E] <= (3 / c)%:E * norm1 setT f.
 Proof.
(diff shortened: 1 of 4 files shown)


zstone1 (approved):
Docs changes are good. Putting the code in a less-dependent order certainly can't hurt. Although given the shape of the diff it's hard to tell if it helps. I'll make my usual plea for splitting these big files.

affeldt-aist:
I also think that `lebesgue_measure.v` should be split, it not well named anyway, its first part is about real-valued measurable functions, I didn't go as far as appending its first part to `measure.v` because the latter is already very long, a new file like `measurable_realfun.v` might be better, what about recording the splitting idea as a issue for later (unless you think it is a better idea to do it now)?

zstone1:
I think a `measurable_realfun.v` would be great. But it doesn't have to happen now. An issue is sufficient.

PR #1332, opened 2024-10-01
add in_nearW

IshiguroYoshihiro:
add lemma which is `{in S, ...}` version of `nearW` for topologicalType and open set `S`.

this lemma is used in other my work (#1294 )

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/topology.v
+++ b/theories/topology.v
@@ -7176,3 +7176,13 @@ Qed.
 Local Close Scope relation_scope.
 
 #[global] Hint Resolve uniform_regular : core.
+
+Lemma in_nearW (R : topologicalType) (P : R -> Prop) (S : set R) :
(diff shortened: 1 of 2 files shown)


theories/topology.v
```diff
@@ -7176,3 +7176,13 @@ Qed.
 Local Close Scope relation_scope.
 
 #[global] Hint Resolve uniform_regular : core.
+
+Lemma in_nearW (R : topologicalType) (P : R -> Prop) (S : set R) :
+  @open R^o S ->
```
zstone1, line 7181: I think the R^o is unnecessary here.

zstone1 (approved):
Small clean up. Otherwise looks good

PR #1351, opened 2024-10-14
Pi irrational

affeldt-aist:

@LaurenceRideau 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/exercise6.v
```diff
@@ -0,0 +1,636 @@
...
+
+Let pirat : R := pirat na nb.
+
+Let f := @f R na nb.
+
+(* NB: helper lemma, not elegant... *)
+Let abx x : 0 < x < pirat -> 0 < a na - b nb * x.
+Proof.
+move=> /andP[x0]; rewrite subr_gt0.
+rewrite /pirat /exercise6.pirat ltr_pdivlMr 1?mulrC//.
```
zstone1, line 378: The explicit dependency on the filename is a little jarring. I would prefer at most one of the following to be true, ideally neither
- The file is called exercise6 without any context
- The text of the filename explicitly appears in the proof script.

zstone1 (commented):
Overall a nice milestone that we can do this in ~600 lines now. Sadly at least 50% of that seems to be about measurable/differentiable/integrable stuff that ought to be automated. But it's a nice application nonetheless. 

Moving some of the files around as you mentioned in your comments is probably good. Especially the polynomial stuff will be nice to have independently. And the filename thing should be an easy fix. Otherwise I don't see any blockers

theories/exercise6.v
```diff
@@ -0,0 +1,636 @@
...
+Global Instance is_derive_poly (p : {poly R}) (x : R) :
+  is_derive x (1:R) (horner p) p^`().[x].
+Proof.
+by apply: DeriveDef; [exact: derivable_horner|rewrite derivE derive1E].
+Qed.
+
+End derive_horner.
+
+Local Open Scope classical_set_scope.
+
```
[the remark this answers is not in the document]
affeldt-aist: You're right, having a `continuous_horner` lemma helps.

theories/exercise6.v
```diff
@@ -0,0 +1,636 @@
...
+Lemma nat_int {R : archiNumDomainType} (m : nat) : (m%:R)%R \is a @Num.int R.
+Proof. by rewrite intrEge0. Qed.
+
+Import numFieldNormedType.
+
+Local Notation "f ^` ()" := (derive1 f) : classical_set_scope.
+
+(* connect MathComp's polynomials with analysis *)
+Section derive_horner.
```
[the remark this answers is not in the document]
affeldt-aist: I moved most of it to `derive.v`.

theories/exercise6.v
```diff
@@ -0,0 +1,636 @@
...
+rewrite addr0 -derive1E.
+have := @D2FDF R na _ n nb0.
+rewrite -/F derivSn /fsin -/f => <-.
+rewrite -!derivE derivn1.
+rewrite [X in (X + _ = _)%R](_ : _ = (F n)^`()^`().[x]%R * sin x)%R; last first.
+  by rewrite mulrC.
+by rewrite -mulrDl hornerD.
+Qed.
+
+Definition intfsin n := \int[mu]_(x in `[0, pi]) (fsin n x).
```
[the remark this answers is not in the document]
affeldt-aist: I put everything in a module divided into two parts where internal lemmas are marked as `Let` and only lemmas that are useful outside are marked as `Lemma`, this way everything but the main lemma is quarantined and internally only lemmas that survive their section are marked as such.

theories/exercise6.v
affeldt-aist: File renamed to `pi_irrational.v`. Surely at some point it should integrate a larger development on analytic number theory.

theories/pi_irrational.v
```diff
@@ -0,0 +1,477 @@
...
+Local Open Scope classical_set_scope.
+Lemma exp_fact (R : realType) (r : R) :
+  (r ^+ n / n`!%:R @[n --> \oo] --> 0)%R.
+Proof.
+apply/cvg_series_cvg_0.
+exact: is_cvg_series_exp_coeff.
+Qed.
+
+Local Open Scope ring_scope.
+Lemma exp_fact_alternative (R : realType) (r : R) :
```
zstone1, line 40: Is there a reason to keep this version around?

zstone1 (approved):
Looks pretty good at this point. One question about an extra lemma, but I'm not seeing any blockers at this point

theories/pi_irrational.v
affeldt-aist: It is mentioned in the accompanying paper, that's why it is here, but I will move it elsewhere before merging (as an exercise to some lecture notes for example).

PR #1366, opened 2024-10-25
duality between `closure` and `interior`

t6s:

This PR adds some lemmas for the closure and interior operators that
are a basic part of the duality between these.  This addition will serve
as an initial step towards the theory of regular open/closed sets.

Lemma `closureC` is modified since its original statement seemed not
conforming to the name.

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/topology_theory/topology_structure.v
+++ b/theories/topology_theory/topology_structure.v
@@ -30,8 +30,10 @@ From mathcomp Require Export filter.
 (*               interior U == all of the points which are locally in U       *)
 (*                             i.e. the largest open set contained in U       *)
 (*                closure U == the smallest closed set containing U           *)
-(*                regopen U == U is equal to the interior of its closure      *)
-(*              regclosed U == U is equal to the closure of its interior      *)
+(*                regopen U == U is regular open                              *)
+(*                             i.e. equal to the interior of its closure      *)
+(*              regclosed U == U is regular closed                            *)
+(*                             i.e. equal to the closure of its interior      *)
 (*           open_of_nbhs B == the open sets induced by neighborhoods         *)
 (*           nbhs_of_open B == the neighborhoods induced by open sets         *)
 (*                      x^' == set of neighbourhoods of x where x is          *)
@@ -829,7 +831,7 @@ exists U=> //; apply/(subset_trans UX)/disjoints_subset; rewrite setIC.
 by apply/eqP/negbNE/negP; rewrite set0P.
 Qed.
 
-Lemma closureC (A : set T) : closure (~` A) = ~` A^°.
+Let _to_be_closureC (A : set T) : closure (~` A) = ~` A^°.
 Proof. by apply: setC_inj; rewrite -interiorC !setCK. Qed.
 
 Lemma closureU (A B : set T) : closure (A `|` B) = closure A `|` closure B.
@@ -837,7 +839,7 @@ Proof. by apply: setC_inj; rewrite setCU -!interiorC -interiorI setCU. Qed.
 
 Lemma interiorU (A B : set T) : A^° `|` B^° `<=` (A `|` B)^°.
 Proof.
-by apply: subsetC2; rewrite setCU -!closureC setCU; exact: closureI.
+by apply: subsetC2; rewrite setCU -!_to_be_closureC setCU; exact: closureI.
 Qed.
 
 Lemma closureEbigcap (A : set T) :
@@ -847,7 +849,7 @@ Proof. exact: closureE. Qed.
 Lemma interiorEbigcup (A : set T) :
   A^° = \bigcup_(x in [set U | open U /\ U `<=` A]) x.
 Proof.
-apply: setC_inj; rewrite -closureC closureEbigcap setC_bigcup.
+apply: setC_inj; rewrite -_to_be_closureC closureEbigcap setC_bigcup.
 rewrite -[RHS](bigcap_image _ setC idfun) /=.
 apply: eq_bigcapl; split=> X /=.
   by rewrite -openC -setCS setCK; exists (~` X)=> //; rewrite setCK.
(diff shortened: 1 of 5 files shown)


theories/topology_theory/topology_structure.v
```diff
@@ -820,6 +813,74 @@ Qed.
...
+Proof.
+apply: setC_inj; rewrite -closureC closureEbigcap setC_bigcup.
+rewrite -[RHS](bigcap_image _ setC idfun) /=.
+apply: eq_bigcapl; split=> X /=.
+  by rewrite -openC -setCS setCK; exists (~` X)=> //; rewrite setCK.
+by case=> Y + <-; rewrite closedC setCS.
+Qed.
+
+Lemma interior_closure_idem (A : set T) :
+  interior (closure (interior (closure A))) = interior (closure A).
```
affeldt-aist, line 857: Quick thought: could the following definition be useful?
```coq
Definition idempotent_fun (U : Type) (f : U -> U) := f \o f =1 f.
```

t6s: Nice. Which file should it be put in?

zstone1 (approved):
All looks good. I'm a bit surprised we didn't have some of this fundamental stuff before, so I'm glad to see you adding it.

theories/topology_theory/topology_structure.v
zstone1: (oh wow, mathcomp's existing `idempotent f` predicate means `f x x = x` and not `f (f x) = f x`. I am always surprised by that one.) mathcomp-extra is a suitable place for this, I think

theories/topology_theory/topology_structure.v
```diff
@@ -804,14 +805,6 @@ rewrite eqEsubset; split=> [x ? B [cB AB]|]; first exact/cB/(closure_subset AB).
 exact: (smallest_sub (@closed_closure _ _) (@subset_closure _ _)).
 Qed.
 
-Lemma closureC E :
```
affeldt-aist, line 807: Ideally, this should better be deprecated with a pragma instead of being removed (you'll find examples here and there in the code of such deprecations).

theories/topology_theory/topology_structure.v
affeldt-aist: > I am always surprised by that one.

so maybe we could propose to rename it to `idempotent_op`?

t6s:
> Don't hesitate to use `^°` when you can. @yoshihiro503 worked hard on making it clickable in the online documentation (https://math-comp.github.io/analysis/htmldoc_1_6_0/index.html) ^_^

The problem is, I cannot easily input that circle symbol.  (Do you know any SKK keystroke for that?)
But anyway, I'll do some copy-pasting later.

t6s:
`idempotent_fun` is added.
I have also introduced definitions `regopen` and `regclosed` to state generalized versions of the idempotent lemmas.
I am afraid that these names are not particularly good, but `regular_open` seems too long.

theories/topology_theory/topology_structure.v
t6s: The original `closureC` is being replaced by a different lemma (with a completely different statement) of the same name.  Does that deprecation mechanism work also for this case?

affeldt-aist: We can shorten the proof of `closureC` and keep it around for a while with, say:
```coq
#[deprecated(since="mathcomp-analysis 1.7.0", note="use this other lemma instead")]
```
so that developments using don't break right away.

theories/topology_theory/topology_structure.v
```diff
@@ -829,15 +831,15 @@ exists U=> //; apply/(subset_trans UX)/disjoints_subset; rewrite setIC.
 by apply/eqP/negbNE/negP; rewrite set0P.
 Qed.
 
-Lemma closureC (A : set T) : closure (~` A) = ~` A^°.
```
affeldt-aist: What about calling it `closure_setC` (somehow the long version of `closureC`) and keep a comment or an issue to rename it once the deprecation of `closureC` is eliminated (in a few releases)?

theories/topology_theory/topology_structure.v
```diff
@@ -887,6 +889,12 @@ Proof. move=> ?; exact/closure_open_regclosed/open_interior. Qed.
 
 End closure_interior_lemmas.
 
+Lemma closureC_deprecated (T : topologicalType) (E : set T) :
+  ~` closure E = \bigcup_(x in [set U | open U /\ U `<=` ~` E]) x.
+Proof. by rewrite -interiorC interiorEbigcup. Qed.
+#[deprecated(since="mathcomp-analysis 1.6.1", note="use `interiorC` and `interiorEbigcup` instead")]
+Notation closureC := closureC_deprecated (only parsing).
```
affeldt-aist, line 896: I heard that it is not necessary anymore to introduce a notation for deprecation pragmas to work with recent versions of Coq.

theories/topology_theory/topology_structure.v
t6s: That will indeed be better.

theories/topology_theory/topology_structure.v
```diff
@@ -820,6 +817,91 @@ Qed.
...
+
+Lemma closure_interior_idem : @idempotent_fun (set T) (closure \o interior).
+Proof. move=> ?; exact/closure_open_regclosed/open_interior. Qed.
+
+End closure_interior_lemmas.
+
+Lemma closureC_deprecated (T : topologicalType) (E : set T) :
+  ~` closure E = \bigcup_(x in [set U | open U /\ U `<=` ~` E]) x.
+Proof. by rewrite -interiorC interiorEbigcup. Qed.
+#[deprecated(since="mathcomp-analysis 1.6.1", note="use `interiorC` and `interiorEbigcup` instead")]
```
affeldt-aist, line 902:
```suggestion
#[deprecated(since="mathcomp-analysis 1.7.0", note="use `interiorC` and `interiorEbigcup` instead")]
```

affeldt-aist (approved):

PR #1371, opened 2024-10-29
proof of L'Hopital rule

ndslusarz:

Co-authored by: @affeldt-aist @hoheinzollern

Proofs of L'Hopital rule for limits taken on left and right, and Cauchy's mean value theorem.

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/realfun.v
```diff
@@ -2544,3 +2547,232 @@ apply/continuous_within_itvP => //; split.
...
 
 End variation_continuity.
+
+Section Cauchy_MVT.
+Context {R : realType}.
+Variables (f df g dg : R -> R) (a b c : R).
+Hypothesis ab : a < b.
+Hypotheses (cf : {within `[a, b], continuous f})
+           (cg : {within `[a, b], continuous g}).
+Hypotheses (fdf : forall x, x \in `]a, b[%R -> is_derive x 1 f (df x))
```
zstone1, line 2557: Should we extend the predicate `derivable_oo_continuous_bnd` to include an option for an explicit derivative as an argument (E.G. `derivable_oo_continuous_bnd_with f df x y`)?

theories/realfun.v
```diff
@@ -2544,3 +2547,232 @@ apply/continuous_within_itvP => //; split.
...
+           (gdg : forall x, x \in `]a, b[%R -> is_derive x 1 g (dg x)).
+Hypotheses (dg0 : forall x, x \in `]a, b[%R -> dg x != 0).
+
+Lemma cauchy_MVT :
+  exists2 c, c \in `]a, b[%R & df c / dg c = (f b - f a) / (g b - g a).
```
zstone1, line 2562: Seems like callers will need to know that `g b - g a != 0`. We might as well deduplicate that a bit. I would recommend either an auxiliary lemma that for any `g`, `a` and `b`, `{in ]a,b[, dg x != 0 -> g b - g a != 0`. Or maybe put a `g b - g a != 0` as an extra clause in the result of `cauchy_MVT`

theories/realfun.v
```diff
@@ -2544,3 +2547,232 @@ apply/continuous_within_itvP => //; split.
...
+have -> := @derive_val _ R _ _ _ _ _ derh.
+move=> /eqP; rewrite eq_sym subr_eq add0r => /eqP ->.
+by rewrite -mulrA divff ?mulr1//; exact: dg0.
+Qed.
+
+End Cauchy_MVT.
+
+Section lhopital.
+Context {R : realType}.
+Variables (f df g dg : R -> R) (a : R) (U : set R) (Ua : nbhs a U).
```
zstone1, line 2593: As usual, a question about boundary conditions. I'm a bit surprised to see we require `f` to be derivable in a full neighborhood of `a`. But then only take the right/left limit. Instead I would expect to see either `a^'- U` or `a^'+U` depending on the direction of the limit. Will the theorem still go through with that weakening?

zstone1 (commented):
Main comment is about weakening `nbhs a U` to be only on the side of the limit. But generally happy to see more foundational calc stuff happening

theories/realfun.v
ndslusarz: Done, as a separate lemma instead of inside the proof, but unsure if you perhaps meant outside of the MVT Cauchy section (so without all of its assumptions?)

Also I'm a bit unsure of what the naming convention would be for that one, so happy to rename it to something proper.

theories/realfun.v
ndslusarz: Gave it a try, but couldn't extend quickly, so perhaps something to extend separately for a different PR?

zstone1: Yeah, I have no problem with that. Happy to deal with it later.

theories/realfun.v
zstone1: This makes sense to me. I don't have any idea for naming here, so I'd say it's fine for now.

affeldt-aist:
moved the lemmas to `derive.v` which is earlier in the file hierarchy, removed dead code in Cauchy's MVT, fix doc and changelog, but couldn't use left/right filters to generalize the proof yet

theories/realfun.v
affeldt-aist: @ndslusarz do you have progress to share?

ndslusarz: Sorry, forgot to send an update before xmas - unfortunately not a major update, haven't been able to make it work for the restricted neighbourhoods (but still think it should be possible, so I'll get back on that). I could push the unfinished version in the meantime if you'd like?

affeldt-aist: Yes.

ndslusarz: Pushed.

affeldt-aist:
The L'Hopital's rule in this PR can be derived from the following one:
```coq
Section lhopital.
Context {R : realType}.
Variables (f df g dg : R -> R) (a b : R) (ab : a < b).
Hypotheses (fdf : forall x, x \in `]a, b[ -> is_derive x 1 f (df x))
           (gdg : forall x, x \in `]a, b[ -> is_derive x 1 g (dg x)).
Hypotheses (fa0 : f x @[x --> a^'+] --> 0) (ga0 : g x @[x --> a^'+] --> 0)
           (cdg : forall x, x \in `]a, b[ -> dg x != 0).

Lemma lhopital_right (l : R) :
  df x / dg x @[x --> a^'+] --> l -> f x / g x @[x --> a^'+] --> l.
Proof.
```
which I obtained as a formalization of a proof by Rudin.
Would it address your comment @zstone1 ?
(It sill needs to be generalized to `l` an extended real and we still need to derive the "left" version.)

zstone1:
Yes, that looks like the right boundary conditions. Proving the left version, and then the double sided versions "should" be easy consequences

PR #1391, opened 2024-11-13
independence, expectation of product

affeldt-aist:
##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


affeldt-aist:
Maybe it is better to rename `expectation_prod` to `expectation_mul` since the original intent was `expectationM`.

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+From mathcomp Require Import reals ereal signed topology normedtype sequences.
+From mathcomp Require Import esum measure exp numfun lebesgue_measure.
+From mathcomp Require Import lebesgue_integral kernel probability.
+
+(**md**************************************************************************)
+(* # Independence                                                             *)
+(*                                                                            *)
+(* ```                                                                        *)
+(* independent_events I F == the events F indexed by I are independent        *)
+(* mutual_independence I F == the classes F indexed by I are independent      *)
```
t6s, line 17: s/classes/set_systems/

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+move=> FA sFT.
+apply/seteqP; split=> [B|B].
+  apply: sub_sigma_algebra2.
+  exact: subsetUl.
+apply: smallest_sub => //.
+move=> C [FC|/= ->//].
+exact: sub_gen_smallest.
+Qed.
+
+Lemma g_sigma_algebra_measure_unique142b :
```
t6s, line 217: g_sigma_algebra_finite_measure_unique

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+
+Section lem142b.
+Context d {T : measurableType d} {R : realType}.
+Variable mu : {finite_measure set T -> \bar R}.
+Variable F : set (set T).
+Hypothesis setI_closed_F : setI_closed F. (* pi-system *)
+Hypothesis FT : F `<=` @measurable _ T.
+Hypothesis sFT : <<s F >> = @measurable _ T.
+
+Lemma lem142b : forall nu : {finite_measure set T -> \bar R},
```
t6s, line 251: This looks like just a rephrasing of the lemma immediately above. Why not use that instead?

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+Qed.
+
+End prode_ge0.
+
+Section mutual_independence_properties.
+Context {R : realType} d {T : measurableType d}.
+Variable P : probability T R.
+Local Open Scope ereal_scope.
+
+Lemma mutual_independence_finite (I0 : choiceType) (I : {fset I0}%fset)
```
t6s, line 330: mutual_independence_fset?

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+done.
+done.
+Qed.
+
+End mutual_independence_properties.
+
+Section g_sigma_algebra_mapping_lemmas.
+Context d {T : measurableType d} {R : realType}.
+
+Lemma g_sigma_algebra_mapping_comp (X : {mfun T >-> R}) (f : R -> R) :
```
t6s, line 683: g_sigma_algebra_"preimage", instead of g_sigma_algebra_mapping?

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+
+End independent_RVs.
+
+Section independent_RVs2.
+Context {R : realType} d (T : measurableType d).
+Variable P : probability T R.
+
+Definition independent_RVs2 d'' (T'' : measurableType d'')
+    (X Y : {mfun T >-> T''}) :=
+  independent_RVs P [set: bool] [eta (fun=> cst point) with false |-> X, true |-> Y].
```
t6s, line 721: simply if _ then _ else _ ?

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+
+End independent_RVs2.
+
+Section thm216.
+Context {R : realType} d (T : measurableType d).
+Context {I0 : choiceType}.
+Context {d' : I0 -> _} (T' : forall i : I0, measurableType (d' i)).
+Variable P : probability T R.
+
+Lemma thm216 (I : set I0) (F : forall i : I0, set (set (T' i)))
```
t6s, line 731: independent_RVs_g_preimage, or something like that?

t6s: in Klenke, "independent_generators"

theories/independence.v
```diff
@@ -0,0 +1,1340 @@
...
+Qed.
+
+End independent_RVs_lemmas.
+
+Definition preimage_classes I (d : I -> measure_display)
+    (Tn : forall k, semiRingOfSetsType (d k)) (T : Type) (fn : forall k, T -> Tn k) :=
+  <<s \bigcup_k preimage_class setT (fn k) measurable >>.
+Arguments preimage_classes {I} d Tn {T} fn.
+
+Lemma measurable_fun_prod d [T : measurableType d] [R : realType] [D : set T] [I : eqType]
```
t6s, line 860: measurable_prod?

t6s (commented):
I have reviewed lines 0 -- 878 and quickly looked through 879-.
The latter things need some cleaning.

theories/independence.v
affeldt-aist: What about `g_sigma_algebra_fun` for the sigma-algebra generated by one function? (this is also sometimes called pull-back sigma-algebra but I am not sure this terminology makes up for easily searchable names)

affeldt-aist: Let's go with `preimage`

theories/probability.v
```diff
@@ -862,6 +880,436 @@ Qed.
...
+
+Definition independent_RVs2 (X Y : {mfun T >-> T'}) :=
+  independent_RVs [set: bool] [eta (fun=> cst point) with false |-> X, true |-> Y].
+
+End independent_RVs.
+
+Section g_sigma_algebra_mapping_lemmas.
+Context d {T : measurableType d} {R : realType}.
+
+Lemma g_sigma_algebra_mapping_comp (X : {mfun T >-> R}) (f : R -> R) :
```
affeldt-aist, line 926: This should be generalized.

affeldt-aist:
> I forgot the context for this draft PR. What is the current status?

Rebased but for reasons I forgot some things look duplicated in `probability.v` and `independence.v`.
If I remember correctly, we introduced definitions of independence following Klenke's book and
were able to reproduce a few, non-trivial lemmas.
What about we remove duplications and PR only an `independence.v` file with what we have so far?
(That would save me the trouble of regular rebasings and certainly these definitions will be wanted at some point.)

theories/probability.v
```diff
@@ -842,6 +842,24 @@ Definition mmt_gen_fun d (T : measurableType d) (R : realType)
...
+  @isMeasurableFun.Build d _ _ _ f^\+
+    (measurable_funrpos (@measurable_funPT _ _ _ _ f)).
+
+HB.instance Definition _ (f : {mfun aT >-> rT}) :=
+  @isMeasurableFun.Build d _ _ _ f^\-
+    (measurable_funrneg (@measurable_funPT _ _ _ _ f)).
+
+HB.instance Definition _ (f : {mfun aT >-> rT}) :=
+  @isMeasurableFun.Build d _ _ _ (@normr _ _ \o f)
+    (measurableT_comp (@normr_measurable _ _) (@measurable_funPT _ _ _ _ f)).
```
t6s: This `HB.instance` is not adding a new instance, and the intended instance already exists:
```
Check fun f : {mfun aT >-> rT} => @normr _ _ \o f : {mfun _ >-> _}.  
```
this succeeds anyway.

theories/numfun.v
```diff
@@ -679,6 +682,149 @@ Proof. by apply/funext=> x; rewrite /patch/=; case: ifP; rewrite ?mule0. Qed.
...
+- rewrite ger0_norm// /funrpos /funrneg; move/max_idPl : (fx0) => ->.
+  by move: fx0; rewrite -{1}oppr0 lerNl => /max_idPr ->; rewrite addr0.
+Qed.
+
+Lemma funrposneg f : f = (fun x => f^\+ x - f^\- x).
+Proof.
+rewrite funeqE => x; rewrite /funrpos /funrneg; have [|/ltW] := leP (f x) 0.
+  by rewrite -{1}oppr0 -lerNr => /max_idPl ->; rewrite opprK add0r.
+by rewrite -{1}oppr0 -lerNl => /max_idPr ->; rewrite subr0.
+Qed.
```
t6s: I suggest renaming
- `funr_normr` -> `funrposDneg_norm`
- `funrposneg` -> `funrposBneg`

Also the latter can be stated in the form `f = f^\+ \+ f^\-`, subsuming the next lemma.

theories/numfun.v
```diff
@@ -679,6 +682,149 @@ Proof. by apply/funext=> x; rewrite /patch/=; case: ifP; rewrite ?mule0. Qed.
...
+
+Lemma funrD_Dpos f g : f \+ g = (f \+ g)^\+ \- (f \+ g)^\-.
+Proof.
+apply/funext => x; rewrite /funrpos /funrneg/=; have [|/ltW] := lerP 0 (f x + g x).
+- by rewrite -{1}oppr0 -lerNl => /max_idPr ->; rewrite subr0.
+- by rewrite -{1}oppr0 -lerNr => /max_idPl ->; rewrite opprK add0r.
+Qed.
```
t6s: This lemma should be removed as it is just an instance of `funrposneg` above.

theories/numfun.v
```diff
@@ -679,6 +682,149 @@ Proof. by apply/funext=> x; rewrite /patch/=; case: ifP; rewrite ?mule0. Qed.
...
+
+Lemma funrD_posD f g : f \+ g = (f^\+ \+ g^\+) \- (f^\- \+ g^\-).
```
t6s: This one, on the other hand, is nontrivial and should be kept.

t6s: But the name looks bad and I have no good suggestion for now.

t6s (changes requested):
> I have reviewed lines 0 -- 878 and quickly looked through 879-. The latter things need some cleaning.

It looks like I reviewed `independence.v` once, and the file looks not so bad. @hoheinzollern can you double check?

I have read the other files and put some comments requesting some changes.

theories/numfun.v
t6s: `funrD_posBneg` would be better.

This can be seen as a form of distributivity where `^\+` and `^\-` parallelly distributes over `\+` inside
expressions of the form `_^\+ \- _^\-`.  I do not know how to name it.

theories/probability.v
affeldt-aist: And moreover this section has nothing to do here, thanks.

theories/numfun.v
affeldt-aist: With this naming it is maybe better to reverse the equality to have the more complicated expression on the left, and then we can maybe drop the `_norm` suffix.

t6s: that's better

theories/numfun.v
affeldt-aist: In fact this lemma has a short proof once one observes that it almost corresponds to `interchange` with addition _and_ subtraction (most if not all uses of `interchange` in MathComp use twice the same operator). `interchange` lemmas are suffixed with `ACA` like `addrACA` but that might be an overkill here to abide by this naming scheme because we mix two operators and moreover on functions (calling for a `fctE` suffix) so maybe `addBr` could be enough to suggests `interchange sub add` and therefore the name could just be `funrDB` (and `funeDB` for the corresponding lemmas for extended real numbers).

t6s: `interchange` is indeed another structure involved here, and I also thought about incorporating `ACA` in its name, though not coming up with a tidy result. 

`funrDB` is neat but looks a bit too stripped off since it does not clearly indicate that it is about `^+` and `^-`.
I am however not sure how to restore that info without elongating the name, and weakly approving `funrDB`.

theories/independence.v
```diff
@@ -0,0 +1,1009 @@
...
+Local Open Scope classical_set_scope.
+Local Open Scope ring_scope.
+
+Section independent_events.
+Context {R : realType} d {T : measurableType d} (P : probability T R)
+  {I0 : choiceType}.
+Local Open Scope ereal_scope.
+
+Definition independent_events (I : set I0) (E : I0 -> set T) :=
+  (forall i, I i -> measurable (E i)) /\
```
CohenCyril: why is measurability built in independence of variables? I would have expected measurability to be a side condition. (Same remark in each definition below)

affeldt-aist: You mean like having the assumption of measurability be one of the parameters of `independent_events` or like having no `measurability` in the definition at all and instead having it appearing as an hypothesis of lemmas of independence?

CohenCyril: Either

theories/probability.v
affeldt-aist: Done.

affeldt-aist:
@t6s I have rebased, cleaned up the proofs, since a few things have been introduced into master since, I could eliminate duplicates. Several files are touched but that is because this PR led us to simplify lemmas about `fun{r,e}pos` and `fun{r,e}neg`. @CohenCyril The measurability conditions in the definitions of independence have been removed and kept only in the lemmas that do require them.

What about we merge? (as a step towards completion of [issue 1808](https://github.com/math-comp/analysis/issues/1808))

affeldt-aist:
(CI green at last)

t6s (approved):
My comments are all resolved and the code looks tidy enough. Looks good to me.

PR #1435, opened 2024-12-16
differentiation under integral

affeldt-aist:

this can certainly be generalized but this version has already been used to prove the Gauss integral so it is maybe worth having anyway

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -29,6 +29,177 @@ Import numFieldNormedType.Exports.
 Local Open Scope classical_set_scope.
 Local Open Scope ring_scope.
 
+Section differentiation_under_integral.
+
+Reserved Notation "'d1 f" (at level 10, f at next level, format "''d1'  f").
+
+Let partial1 {R : realType} {T : Type} (f : R -> T -> R) y : R -> R :=
+  (f ^~ y)^`().
+
+Local Notation "'d1 f" := (partial1 f).
+
+Local Open Scope ring_scope.
+Context {R : realType} d {Y : measurableType d}
+  {mu : {measure set Y -> \bar R}}.
+Variable f : R -> Y -> R.
+Variable B : set Y.
+Hypothesis mB : measurable B.
+Variable a u v : R.
+Let I : set R := `]u, v[.
+Hypothesis Ia : I a.
+
+Hypothesis intf : forall x, I x -> mu.-integrable B (EFin \o f x).
+
+Hypothesis derf1 : forall x y, I x -> B y -> derivable (f ^~ y) x 1.
+
+Variable G : Y -> R.
+Hypothesis G_ge0 : forall y, 0 <= G y.
+Hypothesis intG : mu.-integrable B (EFin \o G).
+Hypothesis G_ub : forall x y, I x -> B y -> `|'d1 f y x| <= G y.
+
+Let F x' := \int[mu]_(y in B) f x' y.
+
+Lemma cvg_differentiation_under_integral :
+  h^-1 *: (F (h + a) - F a) @[h --> 0^'] --> \int[mu]_(y in B) ('d1 f) y a.
+Proof.
+apply/cvgr_dnbhsP => t [t_neq0 t_cvg0].
+suff: forall x_, (forall n : nat, x_ n != a) ->
+      x_ n @[n --> \oo] --> a -> (forall n, I (x_ n)) ->
+    (x_ n - a)^-1 *: (F (x_ n) - F a) @[n --> \oo] -->
+      \int[mu]_(y in B) ('d1 f) y a.
+  move=> suf.
+  apply/cvgrPdist_le => /= r r0.
+  have [rho /= rho0 arhouv] := near_in_itvoo Ia.
+  move/cvgr_dist_lt : (t_cvg0) => /(_ _ rho0)[m _ t_cvg0'].
+  near \oo => N.
+  pose x k := a + t (N + k)%N.
+  have x_a n : x n != a by rewrite /x addrC eq_sym -subr_eq subrr eq_sym t_neq0.
+  have x_cvg_a : x n @[n --> \oo] --> a.
+    apply: cvg_zero.
+    rewrite [X in X @ _ --> _](_ : _ = (fun n => t (n + N)%N)); last first.
+      by apply/funext => n; rewrite /x fctE addrAC subrr add0r addnC.
+    by rewrite cvg_shiftn.
+  have Ix n : I (x n).
+    apply: arhouv => /=.
+    rewrite /x opprD addrA subrr.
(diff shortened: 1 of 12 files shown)


theories/normedtype.v
```diff
@@ -2651,6 +2660,12 @@ Qed.
 Lemma cvg_zero f a : (f - cst a) @ F --> (0 : V) -> f @ F --> a.
 Proof. by move=> Cfa; apply: cvg_sub0 Cfa (cvg_cst _). Qed.
 
+Lemma cvgr_sub0 f a : (fun x => f x - a) @ F --> 0 <-> f @ F --> a.
```
CohenCyril, line 2663: I don't understand the naming of this lemma.

affeldt-aist: I think that it was named after:
```coq
cvge_sub0 : forall {I : Type} {F : set_system I}, Filter F ->
forall {R : realFieldType} (f : I -> \bar R) [k : \bar R],
k \is a fin_num -> (f x - k)%E @[x --> F] --> 0 <-> f x @[x --> F] --> k
```
What about `cvg_zeroP`? Since it is like `cvg_zero` but going in both directions.

affeldt-aist:
I have added the gauss integral as an application of differentiation under integral

theories/normedtype.v
CohenCyril: Following the convention of `subr_eq0` the latter should be `sube_cvg0`, and the former `subr_cvg0`

affeldt-aist:
@IshiguroYoshihiro do you have improvements to propose?

affeldt-aist:
ping? :-)

theories/ftc.v
```diff
@@ -16,10 +16,14 @@ From mathcomp Require Import lebesgue_integral derive charge.
 (* for the Lebesgue integral. We derive from this theorem a corollary to      *)
 (* compute the definite integral of continuous functions.                     *)
 (*                                                                            *)
+(*             'd1 f == first partial derivative of f                         *)
+(*                      f has type R -> T -> R for R : realType               *)
```
CohenCyril, line 20: This looks very specific... maybe make the notation local, and name it partial1of2 or something like that ?

theories/gauss_integral.v
```diff
@@ -0,0 +1,436 @@
...
+- move=> x x01.
+  by rewrite derive1_atan// mul1r.
+Qed.
+
+End oneDsqr.
+
+#[global] Hint Extern 0 (0 < oneDsqr _)%R => solve[apply: oneDsqr_gt0] : core.
+#[global] Hint Extern 0 (0 <= oneDsqr _)%R => solve[apply: oneDsqr_ge0] : core.
+#[global] Hint Extern 0 (1 <= oneDsqr _)%R => solve[apply: oneDsqr_ge1] : core.
+#[global] Hint Extern 0 (oneDsqr _ != 0)%R => solve[apply: oneDsqr_ge1] : core.
```
CohenCyril, line 73: This could be integrated as part an instance of posnum.

theories/gauss_integral.v
```diff
@@ -0,0 +1,436 @@
...
+
+Section gauss.
+Context {R : realType}.
+Local Open Scope ring_scope.
+
+Definition gauss (x : R) := expR (- x ^+ 2).
```
CohenCyril, line 79: `gauss_fun` ? or `gauss_distribution` ?

theories/gauss_integral.v
```diff
@@ -0,0 +1,436 @@
...
+End gauss_integral_proof.
+End gauss_integral_proof.
+
+Section gauss_integral.
+Context {R : realType}.
+Import gauss_integral_proof.
+Let mu := @lebesgue_measure R.
+
+Lemma integral0y_gauss_pi2 :
+  \int[mu]_(x in `[0%R, +oo[) gauss x = Num.sqrt pi / 2.
```
CohenCyril, line 409:
```suggestion
Lemma integral0y_gauss :
  \int[mu]_(x in `[0%R, +oo[) gauss x = Num.sqrt pi / 2.
```

theories/lebesgue_integral.v
```diff
@@ -3440,6 +3440,77 @@ Qed.
...
+exists (\int[mu]_(x in `[0%R, n.+1%:R]) (f x)%:E); first by exists n.
+apply: ge0_subset_integral => //=.
+  by apply/measurable_EFinP; exact: measurableT_comp.
+by move=> ? _; rewrite lee_fin f0.
+Qed.
+
+Lemma ge0_cvg_integral :
+  (fun n => \int[mu]_(x in `[0%R, n%:R]) (f x)%:E) @ \oo -->
+  \int[mu]_(x in `[0%R, +oo[) (f x)%:E.
+Proof.
```
CohenCyril, line 3503:
```suggestion
Lemma ge0_cvgn_integral :
  \int[mu]_(x in `[0%R, n%:R]) (f x)%:E @[n --> \oo] -->
  \int[mu]_(x in `[0%R, +oo[) (f x)%:E.
Proof.
```

theories/gauss_integral.v
affeldt-aist: `gauss_fun` looks better to me, `distribution` might suggest it is already equipped with structures from `probability.v`

theories/gauss_integral.v
affeldt-aist: I replace the 1st, 2nd, and 3rd hints (the last one was wrong moreover) and added instead
```
Lemma oneDsqr_gt0_subproof x : Signed.spec 0 !=0 >=0 (oneDsqr x).
Proof. by rewrite /= oneDsqr_neq0/= oneDsqr_ge0. Qed.

Canonical oneDsqr_ge0_snum x := Signed.mk (oneDsqr_gt0_subproof x).
```

theories/ftc.v
affeldt-aist: Renamed *and* made local.

affeldt-aist:
@CohenCyril CI's green

theories/gauss_integral.v
affeldt-aist: Indeed, these Hints were not triggered at all. I checked that the posnum is indeed effective (it actually discharges goals that were proved previously by direct invokations). I only kept the one not covered by posnum with the is_true coercion explicit (see the last commit).

affeldt-aist:
ah, forgot to put `ge0_cvgn_integral` in the changelog, one sec

CohenCyril (changes requested):
Wait a sec

CohenCyril (approved):

affeldt-aist:
Many thanks for the careful review!

PR #1441, opened 2024-12-19
Accelerated Analysis

Tragicus:

There are a few lines that take a long time to compile (up to 12s). This PR takes every line that takes above 1s to compile and that is not an import or an HB command and rewrites it in less than a second (when I could). This saves overall 7% of the compilation time (with most of the time save (-29%) in `lebesgue_integral.v`).

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/convex.v
+++ b/theories/convex.v
@@ -226,7 +226,9 @@ have [c1 Ic1 Hc1] : exists2 c1, a < c1 < x & (f x - f a) / (x - a) = 'D_1 f c1.
     have := derivable_within_continuous HDf.
     rewrite continuous_open_subspace//; last exact: interval_open.
     by apply; rewrite inE/= in_itv/= ax.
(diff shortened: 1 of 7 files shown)


Tragicus:
@CohenCyril mentioned that `integral` should probably be locked. I tried something but did not observe any time save on the slow lines a was looking at (the set of which is *obviously* not empty), so I gave up.

Tragicus:
I am not convinced that we want all the changes, I can add the time saved be each change if this helps the decision making.

Tragicus:
Fun fact: ~16% of the total compilation time is spent on HB commands and ~27% is spent importing things.

affeldt-aist (approved):

zstone1 (approved):
Thank you for doing this! Happy to see these improvements

PR #1445, opened 2025-01-04
make not_near_at_right/left equivalence

### LINKED ISSUE #1273: Should we make `not_near_at_rightP` an equivalence?

Yosuke-Ito-345:

fixes #1273 

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -30,6 +30,9 @@
 - in `classical_sets.v`:
   + lemmas `xsectionE`, `ysectionE`
 
+- in `normedtype.v`:
+  + lemma `nbhs_right_leftP`
+
 ### Changed
 
 - in `lebesgue_integrale.v`
@@ -70,7 +73,7 @@
     `lee_nneseries`
     
 - in `normedtype.v`:
-	+ lemmas `not_near_at_rightP`, `not_near_at_leftP`
+  + lemmas `not_near_at_rightP`, `not_near_at_leftP`
 
 ### Deprecated
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -69,6 +69,9 @@
     `is_cvg_nneseries`, `is_cvg_npeseries`, `nneseries_ge0`, `npeseries_le0`,
     `lee_nneseries`
     
+- in `normedtype.v`:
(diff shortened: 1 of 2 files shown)


Yosuke-Ito-345:
Only "master / coq" failed.
Does this mean that v8.20 is incompatible with the master of [Coq](https://github.com/coq/coq)?
To settle this problem, do I have to compile the Coq source code from the master branch?

affeldt-aist:
I don't think you need to take any particular action. This is likely to be a temporary situation.

Yosuke-Ito-345:
@affeldt-aist 
Thank you for the comment.
I am relieved to hear that!

affeldt-aist (approved):
It looks good enough to me. (Though it still feels like lemmas about `at_right` and `at_left` could be streamlined a bit more but I do not see how right away.)

Yosuke-Ito-345:
@affeldt-aist 
Thank you for the review.
I still consider that the proof would be simpler if we assumed `Variable R` to be `realDomainType` rather than `numFieldType`...
(Actually I gave up that option since I could not get a reply from Cyril Cohen.)

affeldt-aist:
> I still consider that the proof would be simpler if we assumed `Variable R` to be `realDomainType` rather than `numFieldType`...

Do you think it would be more useful with `realDomainType`?
We should maybe add a comment to the code to record your observation anyway.

Yosuke-Ito-345:
Yes, I think `realDomainType` would be useful, but only when it is also a filtered type (if I understand correctly).
However, I don't know how difficult it is to make `realDomainType` also filtered, so I am not confident whether we can get enough benefit from this change.

affeldt-aist:
The last commit record your observation as a comment in the code.

Yosuke-Ito-345:
@affeldt-aist 
Thank you for the comment of my observation!

Yosuke-Ito-345:
Sorry, I mistakenly closed the pull request.
I have reopened it.

PR #1446, opened 2025-01-07
tentative generalization of FTC2

affeldt-aist:

- for continuous function on unbounded intervals
- by-product of the Gauss integral (in the sense that it used to prove it in a forthcoming PR (that shall provide yet another proof of the Gauss integral))

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -772,7 +772,7 @@ exact: integral_fune_fin_num.
 Unshelve. all: by end_near. Qed.
 
 (* TODO: move? *)
-Let in_continuous_mkseP {T : realFieldType} {U : realFieldType}
+Let in_continuous_mksetP {T : realFieldType} {U : realFieldType}
     (i : interval T) (f : T -> U) :
   {in i, continuous f} <-> {in [set` i], continuous f}.
 Proof.
@@ -791,7 +791,7 @@ Proof.
 move=> f_ge0 Fxl cf dF Fa dFE.
 have mf : measurable_fun `]a, +oo[ f.
   apply: open_continuous_measurable_fun => //.
-  by move: cf => /continuous_within_itvcyP[/in_continuous_mkseP cf _].
+  by move: cf => /continuous_within_itvcyP[/in_continuous_mksetP cf _].
 rewrite -integral_itv_obnd_cbnd// itv_bnd_infty_bigcup seqDU_bigcup_eq.
 rewrite ge0_integral_bigcup//=; last 3 first.
 - by move=> k; apply: measurableD => //; exact: bigsetU_measurable.
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -8,17 +8,36 @@ From mathcomp Require Import esum measure lebesgue_measure numfun realfun.
 From mathcomp Require Import itv real_interval lebesgue_integral derive charge.
 
 (**md**************************************************************************)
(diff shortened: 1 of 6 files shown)


affeldt-aist:
@IshiguroYoshihiro Can you check if there are generalizations you wanted to do?

theories/ftc.v
```diff
@@ -594,6 +594,152 @@ rewrite -EFinB -cE -GbFbc /G /Rintegral/= fineK//.
 exact: integral_fune_fin_num.
 Unshelve. all: by end_near. Qed.
 
+(* TODO: move? *)
+Let in_continuous_mkseP {T : realFieldType} {U : realFieldType}
```
affeldt-aist, line 598: typo

theories/ftc.v
```diff
@@ -594,6 +594,152 @@ rewrite -EFinB -cE -GbFbc /G /Rintegral/= fineK//.
...
+  {in i, continuous f} <-> {in [set` i], continuous f}.
+Proof.
+split=> [fi x|fi x xi]; first by rewrite inE/=; exact: fi.
+by apply: fi; rewrite inE.
+Qed.
+
+Lemma ge0_continuous_FTC2y (f F : R -> R) a (l : R) :
+  (forall x, a <= x -> 0 <= f x)%R ->
+  F x @[x --> +oo%R] --> l ->
+  {within `[a, +oo[, continuous f} ->
```
t6s, line 609: I suggest swapping these two lines so that conditions on f and conditions on F cohere better.

t6s (approved):
Looks good. Latexified doc for the added lemmas will help the user, but it can be done in a separate PR.

theories/ftc.v
affeldt-aist: Good idea.

affeldt-aist:
> Looks good. Latexified doc for the added lemmas will help the user, but it can be done in a separate PR.

Something like that:
image

t6s (approved):

t6s:
> > Looks good. Latexified doc for the added lemmas will help the user, but it can be done in a separate PR.
> 
> Something like that? (fyi @proux01 ) image

This looks very nice!  There are subtle lists of hypotheses for each of these lemmas, and clarifying them by latex will be also useful.

affeldt-aist:
> There are subtle lists of hypotheses for each of these lemmas, and clarifying them by latex will be also useful.

I was reluctant to do that because I thought that from a user point of view it is enough to know about the lemma is about and to check the exact hypotheses by going to where it is defined in the file. What about just emphasizing that the reader should go to the formal statement for the exact set of hypotheses?

t6s:
Let us keep the header simple and contain only the objective of lemmas. I think I am agreeing with you on this point.
I don't think that emphasis about hypotheses is not particularly necessary because Coq will appropriately show the formal 
hypotheses to the user when needed.

A desirable interface for the hypotheses would be to latex-render them when the user issues the About command,
with the markdown source for that rendering provided by the developer at each statement of lemmas.
(I remember @ybertot once suggested adding to Coq such a facility for adding extra comments to lemmas.)

Anyway, I think this PR is ready to be merged.

affeldt-aist:
> A desirable interface for the hypotheses would be to latex-render them when the user issues the About command

which is something for which @CohenCyril might have a solution

PR #1479, opened 2025-02-14
l'hopital's rule

affeldt-aist:

This PR provides a proof of L'Hopital's rule which is more general
than PR #1371  (I did not add a commit to this PR whose log is too complicated
for a proper rebase).

@ndslusarz I kept the previous versions of the lemmas in sections
lhopital0 and lhopital1, you can use them in your development but
it would be much better if you can use instead the lemmas lhopital_at_right and
lhopital_at_left instead.
Can you check that so that we remove them and your development
still stays compatible with the next version of MathComp-Analysis?

(thanks @zstone1)

closes PR #1371

better to merge PR #1477 before this one (fyi @t6s)

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,11 +4,6 @@
 
 ### Added
 
-### Changed
-
-- in `Rstruct.v`
-  + instantiate `GRinv.inv` by `Rinv` instead of `Rinvx`
-
 - in `mathcomp_extra.v`:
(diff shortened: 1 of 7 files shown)


affeldt-aist:
What about we merge this one? It addresses @zstone1 's comments and its availability in MathComp-Analysis 1.9.0 could make like easier for @ndslusarz .
@hoheinzollern

PR #1494, opened 2025-02-21
WIP inverse in ereals

CohenCyril:
##### Things done/to do

### DIFF
--- a/theories/constructive_ereal.v
+++ b/theories/constructive_ereal.v
@@ -10,7 +10,7 @@
    incorporate it into mathcomp proper where it could then be used for
    bounds of intervals*)
 From mathcomp Require Import all_ssreflect all_algebra finmap.
-From mathcomp.classical Require Import mathcomp_extra.
+From mathcomp Require Import mathcomp_extra.
 Require Import signed.
 
 (******************************************************************************)
@@ -410,11 +410,22 @@ Definition oppe x :=
 Definition mule_subdef x y :=
   match x, y with
   | x%:E , y%:E => (x * y)%:E
+  | +oo, +oo | -oo, -oo => +oo
+  | +oo, -oo | -oo, +oo => -oo
   | -oo, y | y, -oo => if y == 0 then 0 else if 0 < y then -oo else +oo
   | +oo, y | y, +oo => if y == 0 then 0 else if 0 < y then +oo else -oo
   end.
 
 Definition mule := nosimpl mule_subdef.
+
+Definition inve_subdef x :=
+  match x with
+  | x%:E => if x == 0%R then +oo else (x^-1)%:E
+  | +oo => 0
+  | -oo => -oo
+  end.
+
+Definition inve := nosimpl inve_subdef.
 
 Definition abse x := if x is r%:E then `|r|%:E else +oo.
 
@@ -438,6 +449,10 @@ Notation "- x"   := (oppe x) : ereal_scope.
 Notation "*%E"   := mule.
 Notation "x * y" := (mule (x%dE : dual_extended _) (y%dE : dual_extended _)) : ereal_dual_scope.
 Notation "x * y" := (mule x y) : ereal_scope.
+Notation "x ^-1" := (inve (x%dE : dual_extended _)) : ereal_dual_scope.
+Notation "x ^-1" := (inve x) : ereal_scope.
+Notation "x / y" := (mule x%dE (inve y%dE)) : ereal_dual_scope.
+Notation "x / y" := (mule x (inve y)) : ereal_scope.
 Notation "`| x |" := (abse (x%dE : dual_extended _)) : ereal_dual_scope.
 Notation "`| x |" := (abse x) : ereal_scope.
 Arguments abse {R}.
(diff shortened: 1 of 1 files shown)


theories/constructive_ereal.v
```diff
@@ -410,12 +410,23 @@ Definition oppe x :=
 Definition mule_subdef x y :=
   match x, y with
   | x%:E , y%:E => (x * y)%:E
+  | +oo, +oo | -oo, -oo => +oo
+  | +oo, -oo | -oo, +oo => -oo
   | -oo, y | y, -oo => if y == 0 then 0 else if 0 < y then -oo else +oo
```
t6s, line 415: `| -oo, y%:E | x%:E, -oo => ...`
what about writing like this to avoid redundancy in the pattern?

CohenCyril: Well the redundancy is already avoided by coq's pattern-marching compilation, and if you are using `y` from the `y%:E` pattern, it will give you a slightly different outcome as before, which means there will be some minor but probably unnecessary backward compatibility issues.

theories/constructive_ereal.v
```diff
@@ -438,6 +449,10 @@ Notation "- x"   := (oppe x) : ereal_scope.
 Notation "*%E"   := mule.
 Notation "x * y" := (mule (x%dE : dual_extended _) (y%dE : dual_extended _)) : ereal_dual_scope.
 Notation "x * y" := (mule x y) : ereal_scope.
+Notation "x ^-1" := (inve (x%dE : dual_extended _)) : ereal_dual_scope.
```
CohenCyril, line 452: This was probably wrong

theories/constructive_ereal.v
CohenCyril: Actually I'm realizing that we will have incompatibility issues anyway, and this patch was meant to have better conversion rules as before.
However I think we should keep the equalities and inequalities in ereals anyway to avoid going from reals to ereals all the time... but I could be wrong, please feel free to experiment.

CohenCyril:
I will not have time to push this PR forward, so @hoheinzollern please feel free to take ownership

CohenCyril:
Refreshed  for use in #1624

theories/constructive_ereal.v
CohenCyril: After refresh of the PR, the "redundancy" is taken care of by Rocq, and I think this form is more readable.

theories/constructive_ereal.v
CohenCyril: I'm actually not sure this was wrong... Even though `+oo^-1 = 0` while `-oo^-1 = -oo` I'm not sure why this should be swapped for dualized ereals. @proux01 @erikmd unless you could see the benefits. I'd rather save myself from some extra work and use the same inverse function for the dual extended reals...

affeldt-aist (approved):
Maybe it is worth emphasizing in the documentation that the inversion of `+oo` is `0` while the inverse of `-oo` is `-oo`. But this asymmetry might anyway be less surprising than using `fine` so I think we should merge and try to use it.

reals/constructive_ereal.v
```diff
@@ -557,6 +572,10 @@ Notation "x * y" := (mule x%dE y%dE : \bar^d _) : ereal_dual_scope.
 Notation "x * y" := (mule x%E y%E) : ereal_scope.
 Notation "`| x |" := (abse x%dE : \bar^d _) : ereal_dual_scope.
 Notation "`| x |" := (abse x%E) : ereal_scope.
+Notation "x ^-1" := (inve (x%dE : \bar^d _)) : ereal_dual_scope.
```
proux01:
```suggestion
Notation "x ^-1" := (inve x%dE : \bar^d _) : ereal_dual_scope.
```

reals/constructive_ereal.v
```diff
@@ -557,6 +572,10 @@ Notation "x * y" := (mule x%dE y%dE : \bar^d _) : ereal_dual_scope.
...
+Notation "x ^-1" := (inve x) : ereal_scope.
+Notation "x / y" := (mule x%dE (inve y%dE : \bar^d _) : \bar^d _) : ereal_dual_scope.
```
proux01:
```suggestion
Notation "x / y" := (mule x%dE (inve y%dE) : \bar^d _) : ereal_dual_scope.
```

reals/constructive_ereal.v
```diff
@@ -2366,6 +2377,59 @@ Qed.
...
+Lemma inve_eq0 x : (x^-1 == 0) = (x == +oo).
+Proof. by case: inveP; rewrite ?eqxx //= => r /negPf; rewrite eqe invr_eq0. Qed.
+
+Lemma inve_ge0 x : (0 <= x^-1) = (0 <= x).
+Proof.
+by case: inveP; rewrite ?le0y ?lexx //= => r; rewrite lee_fin invr_ge0.
+Qed.
+
+Lemma inve_gt0 x :
+  (0 <= x) ==> ((x != 0) && (x != +oo)) -> (0 < x^-1) = (0 < x).
```
proux01: Wouldn't just
```suggestion
Lemma inve_gt0 x :
  ((x != 0) && (x != +oo)) -> (0 < x^-1) = (0 < x).
```
hold?

proux01 (approved):
Don't we want `inve_le0` and `inve_lt0` while at it?

reals/constructive_ereal.v
CohenCyril: mmmh probably yes!

CohenCyril:
> I haven't had a look, but that sounds like there should be a dual version? What is the rationale for the choice? For addition, the rationale was making (max, adde) (resp. (min, dual_adde)) semirings.

I thought you agreed that there is no use for a dual version (https://github.com/math-comp/analysis/pull/1494#discussion_r2115222735) but maybe I miunderstood?

The rationale is to make sure that the inverse is involutive.

reals/constructive_ereal.v
CohenCyril: ~~Ah sure but it's weaker! The precondition is only necessary when X is nonnegative.~~

proux01: Isn't it always true if we don't have x nonnegative?

CohenCyril: ~~What I mean is `(0 < x^-1) = (0 < x) -> (0 <= x) ==> ((x != 0) && (x != +oo))` is true, but `(0 < x^-1) = (0 < x) -> (x != 0) && (x != +oo)` is not!~~

CohenCyril: Forget what I said. You're right!

CohenCyril: Claim proven

CohenCyril:
Additionally I fixed `inveM_def` to make it a necessary and sufficient condition.

reals/constructive_ereal.v
```diff
@@ -3422,6 +3422,24 @@ Proof.
 by rewrite /inveM_def/=; case: x y => [r||]//= [s||]//= /[!eqe]/negPf-> /negPf->.
 Qed.
 
+Lemma mulVe x : x != 0 -> x \is a fin_num -> x^-1 * x = 1.
+Proof.
+move: x => [||]// r; rewrite eqe => r0 _.
+by rewrite inver (negbTE r0) -EFinM mulVf.
+Qed.
+
+Lemma lefe_pV2 x y : x > 0 -> y > 0 -> (x <= y) = (y^-1 <= x^-1).
```
CohenCyril: This is usually phrased the other way around.

CohenCyril: cf https://github.com/math-comp/math-comp/blob/master/algebra/num_theory/numfield.v#L165-L175

reals/constructive_ereal.v
```diff
@@ -3287,6 +3368,75 @@ Variable R : realFieldType.
...
+by rewrite /inveM_def/=; case: x y => [r||]//= [s||]//= /[!eqe]/negPf-> /negPf->.
+Qed.
+
+Lemma mulVe x : x != 0 -> x \is a fin_num -> x^-1 * x = 1.
+Proof.
+move: x => [||]// r; rewrite eqe => r0 _.
+by rewrite inver (negbTE r0) -EFinM mulVf.
+Qed.
+
+Lemma lefe_pV2 x y : x > 0 -> y > 0 -> (y^-1 <= x^-1) = (x <= y).
```
CohenCyril:
```suggestion
Lemma lee_pV2 x y : x > 0 -> y > 0 -> (y^-1 <= x^-1) = (x <= y).
```

CHANGELOG_UNRELEASED.md
```diff
@@ -32,6 +32,25 @@
...
+  + lemmas `inver`, `inveP`, `fine_invr`, `inve0`, `inve1`, `invey`, `invey`,
+    `inveNy`, `inveK`, `invr_inj`, `inveN`, `inve_eq0`, `inve_ge0`, `inve_gt0`,
+    `inv_gt0P`, `inve_lt0`, `inve_le0`, `inve_le0P`,
+  + predicate `inveM_def` with notation `x *^-1? y` defining a sufficient
+    condition for the inverse and product to commute, with lemmas `inveMP`,
+    `inveM_defE`, `inveM` and `fin_inveM_def`,
+  + compatibility lemma `mule_defE` to bridge the former definition of
+    `mule_def` with the new one.
+  + lemma `fin_numV`
+  + lemmas `mulVe`, `lefe_pV2`
```
CohenCyril:
```suggestion
  + lemmas `mulVe`, `lee_pV2`
```

PR #1532, opened 2025-03-24
Adapt to rocq#19987

Tragicus:

https://github.com/coq/coq/pull/19987 refolds terms before using them to instantiate evars. In `classical/functions.v`, there is a lemma (`image_subP`) that uses `homomorphism_1`, which introduces eta-expansions. Due to unfoldings in the previous algorithm, these eta-expansions were removed, whereas they are not now, which makes `rewrite`'s matching algorithm fail. In `theories/function_spaces`, there is a second order unification problem which is solved differently.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


Tragicus:
My first round of overlays should be ready soon. Is there something to do here (e.g. rerunning the CI if it is fixed)?

affeldt-aist (approved):
Looks reasonable.

PR #1542, opened 2025-03-28
define cdf (cumulative distribution function)

Yosuke-Ito-345:

- introduce the cumulative distribution function

- prove some basic facts on the cumulative distributive function

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


Yosuke-Ito-345:
orz...
I'll take a closer look tomorrow.

affeldt-aist:
It looks like you are removing `interval_inference.v`. I am not sure we want to do that yet.

affeldt-aist:
> It looks like you are removing `interval_inference.v`. I am not sure we want to do that yet.

It also looks like you have issues with `ceil` lemmas. Note that the very last commit in `master` has introduced a file called `unstable.v` that acts like a staging area (which was the role of `mathcomp_extra.v` in practice). We now keep in `mathcomp_extra.v` only lemmas that already made their way into MathComp but that we need to duplicate to keep supporting several versions of MathComp. We are free to use `unstable.v` in MathComp-Analysis but discourage users to use it.

And really lemmas about `ceil` and `floor` (and `trunc`) are a bit unstable because we are reworking their naming. There should be enough deprecation warning to help you but you'd better be defensive but using full qualified names even if it is not necessary.

Yosuke-Ito-345:
Thank you very much for the comment.
I did not remove `interval_inference.v` manually, but `make` in `anlaysis` seems to have removed it. The output on termninal.app says
```
(echo "From mathcomp.algebra Require Import interval_inference." > test_interval_inference.v \
	 && (rocq c test_interval_inference.v > /dev/null 2>&1) \
	 && touch rm_interval_inference) || true
rm -f test_interval_inference.v
test -f rm_interval_inference \
	  && sed -i.bak '/interval_inference/ d' _CoqProject \
	  && rm -f _CoqProject.bak || true
test -f rm_interval_inference \
	  && sed -i.bak '/interval_inference/ d' all_reals.v \
	  && rm -f all_reals.v.bak || true
sed: all_reals.v: No such file or directory
test -f rm_interval_inference \
	  && sed -i.bak '/interval_inference/ d' reals/all_reals.v \
	  && rm -f reals/all_reals.v.bak || true
test -f rm_interval_inference && rm -f interval_inference.v || true
test -f rm_interval_inference && rm -f reals/interval_inference.v || true
rm -f rm_interval_inference
# Remove everything above when requiring mathcomp >= 2.4.0
# (also remove file reals/interval_inference.v and references to it)
/Users/yosukeito/.opam/default/bin/coq_makefile  -f _CoqProject -o Makefile.coq
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f Makefile.coq 
ROCQ DEP VFILES
if command -v coqc > /dev/null && (coqc --version | grep -q '8.18\|8.19\|8.20') ; then \
	  for f in ./experimental_reals/discrete.v ./reals/reals.v ; do \
	     sed -i.bak 's/From Corelib/From Coq/' ${f} ; \
	     sed -i.bak 's/PosDef/PArith/' ${f} ; \
	     rm -f ${f}.bak ; \
	  done ; \
	fi
```

Yosuke-Ito-345:
Should I avoid doing `make` in `analysis`?

affeldt-aist:
> Should I avoid doing `make` in `analysis`?

I don't think so but very likely you should not commit the removal of `interval_inference.v` @proux01 .

Yosuke-Ito-345:
I totally agree with you...
I didn't notice that I have committed the removal of `interval_inference.v`.
I try to avoid errors by doing `git status` or `git diff`, but I can't catch this kind of auto-modification.
(I don't know it is due to my immaturity or the system design.)

Yosuke-Ito-345:
I will recheck my branch and try again!

affeldt-aist:
We are in a situation where `sed` is used to modify files during make. For example, `reals.v` and `discrete.v` are modified but they shouldn't be commited. This is because we are in a transition state trying to reduce dependency on the standard library of Rocq.

I used to commit with `git commit -a` but now I use `git add` and `git commit` separately to avoid unwanted commit (easier not to do them in the first place instead of tracking misses.)

Yosuke-Ito-345:
Oh, I got it...
My knowledge on `git add` was just `git add .`, but I should have specified the file I wanted to edit.
Thank you very much for the advice.
(I'll also keep googling about git and studying it.)

Yosuke-Ito-345:
I decided to send a new pull request to take each step carefully.
I would like to ask you one thing: does "full qualified name" mean [this](https://en.wikipedia.org/wiki/Fully_qualified_name)...?

affeldt-aist:
> I decided to send a new pull request to take each step carefully. I would like to ask you one thing: does "full qualified name" mean [this](https://en.wikipedia.org/wiki/Fully_qualified_name)...?

Yes, it is about writing, e.g., `Coq.Init.Logic.False` instead of just `False`. When the library is clean, it should not make a difference, but when names are in a state of flux, fully qualified names help to see better.

Yosuke-Ito-345:
Thanks for teaching me.
You mean that I should have avoided `Num.ceil` for example?
(Sorry for the repeated questions.)

affeldt-aist:
Both `Num.ceil` and `ceil` should be fine because we do not duplicate definitions, but in case of double `Num.ceil` is safer. But the problem lies more with lemmas because we sometimes have lemmas with the same name in `mathcomp_extra.v` (and now `unstable.v`) and MathComp. In general this is not the case, but this is happening right now with `ceil` and `floor` lemmas, unfortunately.

affeldt-aist:
I took the liberty to convert this PR to draft because PR #1543 has the same contents but makes the CI happy with less commits.

Yosuke-Ito-345:
Thanks for changing the status.
I was thinking of closing this pull request.

Yosuke-Ito-345:
> Both `Num.ceil` and `ceil` should be fine because we do not duplicate definitions, but in case of double `Num.ceil` is safer. But the problem lies more with lemmas because we sometimes have lemmas with the same name in `mathcomp_extra.v` (and now `unstable.v`) and MathComp. In general this is not the case, but this is happening right now with `ceil` and `floor` lemmas, unfortunately.

I got it. Now I'm convinced of the reason. Thank you again for your explanation.

PR #1544, opened 2025-03-31
split normedtype.v

affeldt-aist:
##### Motivation for this change

fixes #1339 

This PR splits `normedtype.v` into:
- `num_normedtype.v`
- `ereal_normedtype.v`
- `pseudometric_normed_Zmodule.v`
- `normed_module.v`
- `matrix_normedtype.v`
- `complete_normed_module.v`
- `urysohn.v`
- `vitali_lemma.v`

This split was discussed during the previous MathComp-Analysis meeting:
https://github.com/math-comp/analysis/wiki/2025%E2%80%9002%E2%80%9020-Meeting

Note that similarly the recent PR (now merged):
- https://github.com/math-comp/analysis/pull/1540

proposed a split of `lebesgue_integral.v`.

##### Checklist

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`

- [x] added corresponding documentation in the headers

Reference: [How to document](https://github.com/math-comp/math-comp/wiki/How-to-document)

##### Merge policy

As a rule of thumb:
- PRs with several commits that make sense individually and that
  all compile are preferentially merged into master.
- PRs with disorganized commits are very likely to be squash-rebased.

##### Reminder to reviewers

- Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs)
- Put a milestone if possible
- Check labels

PR #1563, opened 2025-04-09
generalizations of derive incr/decr lemmas

hoheinzollern:

These are generalization of lemmas about the first derivative ported from [this branch](https://github.com/hoheinzollern/analysis/tree/sampling_20250324).
We introduce all combinations for different intervals to be able to discover them effectively.

@affeldt-aist @t6s 

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/derive.v
+++ b/theories/derive.v
@@ -1598,13 +1598,17 @@ have [c cab D] := MVT altb fdrvbl fcont.
 by exists c => //; rewrite in_itv /= ltW (itvP cab).
 Qed.
 
-Lemma ger0_derive1_homo (R : realType) (f : R^o -> R^o) (a b : R) (sa sb : bool) :
-  (forall x : R, x \in `]a, b[%R -> derivable f x 1) ->
-  (forall x : R, x \in `]a, b[%R -> 0 <= (f^`())%classic x) ->
+Section ger0_derive1_le.
+Context {R : realType}.
+Variables (f : R -> R) (a b : R).
+Hypothesis df : forall x, x \in `]a, b[%R -> derivable f x 1.
+Hypothesis dfge0 : forall x, x \in `]a, b[%R -> 0 <= (f^`())%classic x.
+
+Lemma ger0_derive1_le (sa sb : bool) :
   {within [set` (Interval (BSide sa a) (BSide sb b))], continuous f} ->
-  {in (Interval (BSide sa a) (BSide sb b)) &, {homo f : x y / x <= y >-> x <= y}}.
+  {in Interval (BSide sa a) (BSide sb b) &, {homo f : x y / x <= y >-> x <= y}}.
 Proof.
-move=> df dfge0 cf x y + + xy.
+move=> cf x y + + xy.
 rewrite !itv_boundlr /= => /andP [] ax ? /andP [] ? yb.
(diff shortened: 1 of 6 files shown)


theories/derive.v
```diff
@@ -1587,85 +1587,228 @@ have [c cab D] := MVT altb fdrvbl fcont.
...
+Proof.
+move=> df f'0 cf x y ax xy yb.
+apply: ltr0_derive1_lt_cc; [ exact: df | exact: f'0 | by [] | | | by [] ];
+  rewrite !in_itv/= ?ax ?yb ?andbT/=.
+  by rewrite (le_trans _ (ltW xy)).
+by rewrite (le_trans (ltW xy)).
+Qed.
+
+#[global] Hint Extern 0 (is_true (_ <= ?x)) => match goal with
+  H : x \is_near _ |- _ => by near: x; apply: nbhs_pinfty_ge; rewrite num_real end : core.
```
CohenCyril, line 1794: `by` is strictly forbidden `core` HintDb (causes loops)

CohenCyril (changes requested):
Great! (just change the potentially dangerous hint extern)

theories/derive.v
affeldt-aist: `by` has been removed by the last commit

theories/derive.v
```diff
@@ -1587,85 +1587,228 @@ have [c cab D] := MVT altb fdrvbl fcont.
...
+apply: ltr0_derive1_lt_cc; [ exact: df | exact: f'0 | by [] | | | by [] ];
+  rewrite !in_itv/= ?ax ?yb ?andbT/=.
+  by rewrite (le_trans _ (ltW xy)).
+by rewrite (le_trans (ltW xy)).
+Qed.
+
+#[global] Hint Extern 0 (is_true (_ <= ?x)) => match goal with
+  H : x \is_near _ |- _ => near: x; apply: nbhs_pinfty_ge; exact: num_real end : core.
+#[global] Hint Extern 0 (is_true (?x <= _)) => match goal with
+  H : x \is_near _ |- _ => near: x; apply: nbhs_ninfty_le; exact: num_real end : core.
```
CohenCyril, line 1796: ssr `exact` calls the same tactic as `by []` and `done` which calls auto with `core` db and causes loops.
You should use `solve` instead.

```suggestion
#[global] Hint Extern 0 (is_true (_ <= ?x)) => match goal with
  H : x \is_near _ |- _ => solve [near: x; apply: nbhs_pinfty_ge; apply: num_real] end : core.
#[global] Hint Extern 0 (is_true (?x <= _)) => match goal with
  H : x \is_near _ |- _ => solve [near: x; apply: nbhs_ninfty_le; apply: num_real] end : core.
```

affeldt-aist: The last commit also fixes a number of similarly problematic Hints that were already in `normedtype.v` (note for later: be careful to report these changes when updating the split of `normedtype.v` PR #1544 ).

affeldt-aist: @CohenCyril are you ok with the changes?

affeldt-aist: `by` has been removed but replaced by `now` in a few `Hint`s which should be ok because `now` is not recursive.

PR #1606, opened 2025-05-01
Essential supremum

hoheinzollern:

Introduces lemmas for essential supremum, part of the split of #1506 

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


affeldt-aist:
The CI is failing because it fact the lemmas depends on lemmas in PR #1607 , which should therefore be merged first (and I think that they are ready for merge by the way).

theories/ess_sup_inf.v
```diff
@@ -0,0 +1,366 @@
...
+Proof. by []. Qed.
+
+End essential_supremum.
+
+Section essential_supremum_lemmas.
+Context d {T : measurableType d} {R : realType}.
+Variable mu : {measure set T -> \bar R}.
+Implicit Types (f g : T -> \bar R) (h k : T -> R) (x y : \bar R) (r : R).
+
+Lemma ae_le_measureP f y : measurable_fun [set: T] f ->
```
affeldt-aist, line 48: This could also be moved earlier in the hierarchy of files but it looks quite meaningful to have it here rather than lost in another file. What about just leaving a NB here to remember that this does not really depend on the definitions introduced in this file?

t6s (approved):
looks good

PR #1628, opened 2025-05-22
pseudometric alias

affeldt-aist:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/normedtype_theory/normed_module.v
+++ b/theories/normedtype_theory/normed_module.v
@@ -1959,3 +1959,102 @@ by rewrite interior_closed_ballE //; exact: ballxx.
 Qed.
 
 End Closed_Ball_normedModType.
+
+(* equip a normedZmodType with a structure of normed module *)
+
+Definition pseudometric (K : realType) (M : normedZmodType K) : Type := M.
+
+HB.instance Definition _ (K : realType) (M : normedZmodType K) :=
+  Choice.on (pseudometric M).
+HB.instance Definition _ (K : realType) (M : normedZmodType K) :=
+  Num.NormedZmodule.on (pseudometric M).
+HB.instance Definition _ (K : realType) (M : normedZmodType K) :=
+  isPointed.Build M 0.
+
+Section isnormedmodule.
+Variables (K : realType) (M' : normedZmodType K).
+
+Notation M := (pseudometric M').
+
+Local Definition ball (x : M) (r : K) : set M := ball_ Num.norm x r.
+
+Local Definition ent : set_system (M * M) :=
+  entourage_ ball.
+
+Local Definition nbhs (x : M) : set_system M :=
+  nbhs_ ent x.
+
+Local Lemma nbhsE : nbhs = nbhs_ ent. Proof. by []. Qed.
+
+HB.instance Definition _ := hasNbhs.Build M nbhs.
+
+Local Lemma ball_center x (e : K) : 0 < e -> ball x e x.
+Proof. by rewrite /ball/= subrr normr0. Qed.
+
+Local Lemma ball_sym x y (e : K) : ball x e y -> ball y e x.
+Proof. by rewrite /ball /= distrC. Qed.
+
+Local Lemma ball_triangle x y z e1 e2 : ball x e1 y -> ball y e2 z ->
+  ball x (e1 + e2) z.
+Proof.
+rewrite /ball /= => ? ?.
+(*rewrite -[z](subrK y) -addrA. (le_lt_trans (ler_normD _ _))// addrC ltrD.*)
+Admitted.
+
+Local Lemma entourageE : ent = entourage_ ball.
+Proof. by []. Qed.
+
+HB.instance Definition _ := @Nbhs_isPseudoMetric.Build K M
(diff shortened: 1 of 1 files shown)


theories/normedtype_theory/normed_module.v
```diff
@@ -223,6 +230,95 @@ Module Exports. Export numFieldTopology.Exports. HB.reexport. End Exports.
 End numFieldNormedType.
 Import numFieldNormedType.Exports.
 
+Definition pseudoMetric_normed (K : numFieldType) (M : normedZmodType K) : Type
```
Tragicus: Is there not `pseudometric` somewhere? If so, why do we need to change the alias?

theories/normedtype_theory/normed_module.v
```diff
@@ -223,6 +230,95 @@ Module Exports. Export numFieldTopology.Exports. HB.reexport. End Exports.
...
+by rewrite scaler_nat.
+Qed.
+
+Lemma normrN x : norm (- x) = norm x.
+Proof. by have := normrZ (- 1)%R x; rewrite scaleN1r normrN normr1 mul1r. Qed.
+
+HB.instance Definition _ := Num.Zmodule_isNormed.Build
+  R M ler_normD normr0_eq0 normrMn normrN.
+
+(*HB.saturate pseudoMetric_normed.*)
```
Tragicus: Do we want to keep this?

theories/normedtype_theory/normed_module.v
```diff
@@ -223,6 +230,95 @@ Module Exports. Export numFieldTopology.Exports. HB.reexport. End Exports.
...
 Import numFieldNormedType.Exports.
 
+Definition pseudoMetric_normed (K : numFieldType) (M : normedZmodType K) : Type
+  := M.
+
+HB.instance Definition _ (K : numFieldType) (M : normedZmodType K) :=
+  Choice.on (pseudoMetric_normed M).
+HB.instance Definition _ (K : numFieldType) (M : normedZmodType K) :=
+  Num.NormedZmodule.on (pseudoMetric_normed M).
+(*HB.instance Definition _ (K : numFieldType) (M : normedZmodType K) :=
```
Tragicus: This does not do anything, so we can probably remove it.

theories/normedtype_theory/normed_module.v
```diff
@@ -223,6 +230,95 @@ Module Exports. Export numFieldTopology.Exports. HB.reexport. End Exports.
...
+  isPointed.Build M 0.*)
+
+Section pseudoMetric_from_normedZmodType.
+Variables (K : numFieldType) (M : normedZmodType K).
+
+Notation T := (pseudoMetric_normed M).
+
+Local Definition ball (x : T) (r : K) : set T := ball_ Num.norm x r.
```
Tragicus: These definitions leak in the rest of the file (e.g. in `ball_open`. We may need to hide them in a module.

theories/normedtype_theory/normed_module.v
Tragicus:
```suggestion
Definition pseudoMetric_normed (M : Type) : Type
```

theories/normedtype_theory/normed_module.v
```diff
@@ -223,6 +230,95 @@ Module Exports. Export numFieldTopology.Exports. HB.reexport. End Exports.
...
+Proof. by constructor. Qed.
+
+HB.instance Definition _ := NormedZmod_PseudoMetric_eq_pseudometric.
+
+Lemma PseudoMetricNormedZmod_Lmodule_isNormedModule_pseudometric
+  : PseudoMetricNormedZmod_Lmodule_isNormedModule R M.
+Proof. by constructor; exact: normrZ. Qed.
+
+HB.instance Definition _ :=
+  PseudoMetricNormedZmod_Lmodule_isNormedModule_pseudometric.
```
Tragicus:
```suggestion
HB.instance Definition _ := NormedZmod_PseudoMetric_eq.Build R M erefl.
HB.instance Definition _ :=
  PseudoMetricNormedZmod_Lmodule_isNormedModule.Build R M normrZ.
```

theories/normedtype_theory/normed_module.v
affeldt-aist: I looked at existing aliases (e.g., `pseudoMetric_bool` in `bool_topology.v`) and observed that they had more elaborated names, so I concluded that a name that better explains the input and the output could be more appropriate, hence the name change.

theories/normedtype_theory/normed_module.v
affeldt-aist: Let's remove the comment. (I think that it used to be useful but I don't remember in which circumstances anymore.)

theories/normedtype_theory/normed_module.v
affeldt-aist: Typo?

affeldt-aist: arg, not a good example since this precise example might actually be useless, see issue #1636

Tragicus: Ah, yes, thanks, fixed.

PR #1630, opened 2025-05-22
Conv convention

t6s:
This PR changes the convention of the convex combination operation in `convex.v` from
`a <| t |> b = (1-t)a + tb`  to `ta + (1-t)b`, to make this compatible with infotheo (https://gihtub.com/affeldt-aist/infotheo).

There has been a discussion about these two options, and the former
has been chosen for its compatibility with the homotopical notion of paths,
while the latter is algebraically natural.

After a discussion in a analysis-dev meeting, a conclusion was made to change it to the latter.
Also, there is now a `line_path` in `set_interval.v` which essentially implements the former convention
and has an appropriately geometric name.

Therefore I think it is now reasonable to do this PR.

@CohenCyril @affeldt-aist 

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,7 +4,18 @@
 
 ### Added
 
+- in `unstable.v`:
+  + lemma `subrKC`
+
+- in `convex.v`:
+  + module `ConvexAssoc`
+  + lemmas `convR_itv`, `convR_line_path`, `le_convR`
+
 ### Changed
+
+- in `convex.v`:
+  + convex combination operator `a <| t |> b` changed from
+    `(1-t)a + tb` to `ta + (1-t)b`
 
 - in `sequences.v`:
   + lemma `subset_seqDU`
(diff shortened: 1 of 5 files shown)


affeldt-aist (approved):
Great! This will unlock the port of convex spaces from InfoTheo. Regarding the PR in itself, the technical lemmas about the parameters of the quasi-associativity should be better named and documented. However, I think that it will be more efficient for us.to do that along the port of more properties of convex spaces. So I am ok to merge with the proviso that we agree to carefully review that part of the library again in the near future.

t6s:
I am afraid of naming the submodule `QuasiAssoc`, observing there are other (likely to be many other) variations of "quasi-associativity" in algebra.  E.g. https://academic.oup.com/ptep/article/2014/1/013A03/1552583

affeldt-aist:
Let's be more precise and even hide the module name behind a definition in lowercase akin to `associative`, etc.

PR #1640, opened 2025-06-05
beta distribution

affeldt-aist:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/derive.v
+++ b/theories/derive.v
@@ -1267,6 +1267,12 @@ Proof. by have /derivableP := @derivable_id x v; rewrite derive_val. Qed.
 
 End derive_id.
 
+Lemma derive1_onem {R : numFieldType} :
+  (fun x => 1 - x : R^o)^`()%classic = cst (-1).
+Proof.
+by apply/funext => x; rewrite derive1E deriveB// derive_id derive_cst sub0r.
+Qed.
+
 Section Derive_lemmasVR.
 Variables (R : numFieldType) (V : normedModType R).
 Implicit Types (f g : V -> R) (x v : V).
@@ -1322,20 +1328,36 @@ by rewrite deriveM // !derive_val.
 Qed.
 
 Global Instance is_deriveX f n x v (df : R) :
-  is_derive x v f df -> is_derive x v (f ^+ n.+1) ((n.+1%:R * f x ^+n) *: df).
+  is_derive x v f df -> is_derive x v (f ^+ n) ((n%:R * f x ^+ n.-1) *: df).
 Proof.
+case: n => [fdf|n].
+  rewrite expr0 mul0r scale0r.
+  exact: is_derive_cst.
 move=> dfx; elim: n => [|n ihn]; first by rewrite expr1 expr0 mulr1 scale1r.
 rewrite exprS; apply: is_derive_eq.
 rewrite scalerA -scalerDl mulrCA -[f x * _]exprS.
 by rewrite [in LHS]mulr_natl exprfctE -mulrSr mulr_natl.
 Qed.
 
+(*Global Instance is_deriveX f n x v (df : R) :
+  is_derive x v f df -> is_derive x v (f ^+ n.+1) ((n.+1%:R * f x ^+n) *: df).
+Proof.
+move=> dfx; elim: n => [|n ihn]; first by rewrite expr1 expr0 mulr1 scale1r.
+rewrite exprS; apply: is_derive_eq.
+rewrite scalerA -scalerDl mulrCA -[f x * _]exprS.
+by rewrite [in LHS]mulr_natl exprfctE -mulrSr mulr_natl.
+Qed.*)
+
 Lemma derivableX f n x v : derivable f x v -> derivable (f ^+ n) x v.
 Proof. by case: n => [_|n /derivableP]; [rewrite expr0|]. Qed.
 
 Lemma deriveX f n x v : derivable f x v ->
-  'D_v (f ^+ n.+1) x = (n.+1%:R * f x ^+ n) *: 'D_v f x.
+  'D_v (f ^+ n) x = (n%:R * f x ^+ n.-1) *: 'D_v f x.
 Proof. by move=> /derivableP df; rewrite derive_val. Qed.
+
+(*Lemma deriveX f n x v : derivable f x v ->
+  'D_v (f ^+ n.+1) x = (n.+1%:R * f x ^+ n) *: 'D_v f x.
+Proof. by move=> /derivableP df; rewrite derive_val. Qed.*)
 
 Fact der_inv f x v : f x != 0 -> derivable f x v ->
   (fun h => h^-1 *: (((fun y => (f y)^-1) \o shift x) (h *: v) - (f x)^-1)) @
@@ -1412,15 +1434,22 @@ rewrite (_ : _ ^~ _ = (fun x => x * x ^+ n)); last first.
 by apply: derivableM; first exact: derivable_id.
 Qed.
 
-Lemma exp_derive {R : numFieldType} n x v :
+(*Lemma exp_derive {R : numFieldType} n x v :
   'D_v (@GRing.exp R ^~ n.+1) x = n.+1%:R *: x ^+ n *: v.
+Proof.
+have /= := @deriveX R R id n x v (@derivable_id _ _ _ _).
+by rewrite fctE => ->; rewrite derive_id.
+Qed.*)
+
+Lemma exp_derive {R : numFieldType} n x v :
+  'D_v (@GRing.exp R ^~ n) x = n%:R *: x ^+ n.-1 *: v.
 Proof.
 have /= := @deriveX R R id n x v (@derivable_id _ _ _ _).
 by rewrite fctE => ->; rewrite derive_id.
 Qed.
 
 Lemma exp_derive1 {R : numFieldType} n x :
-  (@GRing.exp R ^~ n.+1)^`() x = n.+1%:R *: x ^+ n.
+  (@GRing.exp R ^~ n)^`() x = n%:R *: x ^+ n.-1.
 Proof. by rewrite derive1E exp_derive [LHS]mulr1. Qed.
 
 Lemma EVT_max (R : realType) (f : R -> R) (a b : R) : (* TODO : Filter not infered *)
--- a/theories/ftc.v
+++ b/theories/ftc.v
@@ -1793,6 +1793,46 @@ Qed.
 
 End integration_by_substitution.
 
+Section integration_by_substitution_onem.
+Context {R : realType}.
+Let mu := (@lebesgue_measure R).
+Local Open Scope ereal_scope.
+
+Lemma integration_by_substitution_onem (G : R -> R) (r : R) :
+  (0 < r <= 1)%R ->
+  {within `[0%R, r], continuous G} ->
+  (\int[mu]_(x in `[0%R, r]) (G x)%:E =
+  \int[mu]_(x in `[(1 - r)%R, 1%R]) (G (1 - x))%:E).
+Proof.
+move=> r01 cG.
+have := @integration_by_substitution_decreasing R (fun x => 1 - x)%R G (1 - r) 1.
+rewrite subKr subrr => -> //.
+- by apply: eq_integral => x xr; rewrite !fctE derive1_onem opprK mulr1.
+- by rewrite ltrBlDl ltrDr; case/andP : r01.
+- by move=> x y _ _ xy; rewrite ler_ltB.
+- by rewrite derive1_onem; move=> ? ?; exact: cvg_cst.
+- by rewrite derive1_onem; exact: is_cvg_cst.
+- by rewrite derive1_onem; exact: is_cvg_cst.
+- split => /=.
+  + by move=> x xr1; exact: derivableB.
+  + apply: cvg_at_right_filter; rewrite subKr.
+    apply: (@continuous_comp_cvg _ R^o R^o _ (fun x => 1 - x)%R)=> //=.
+      by move=> x; apply: (@continuousB _ R^o)  => //; exact: cvg_cst.
+    by under eq_fun do rewrite subKr; exact: cvg_id.
+  + by apply: cvg_at_left_filter; apply: (@cvgB _ R^o) => //; exact: cvg_cst.
+Qed.
+
+Lemma Rintegration_by_substitution_onem (G : R -> R) (r : R) :
+  (0 < r <= 1)%R ->
+  {within `[0%R, r], continuous G} ->
+  (\int[mu]_(x in `[0%R, r]) (G x) =
+  \int[mu]_(x in `[(1 - r)%R, 1%R]) (G (1 - x)))%R.
+Proof.
+by move=> r01 cG; rewrite [in LHS]/Rintegral integration_by_substitution_onem.
+Qed.
+
+End integration_by_substitution_onem.
+
 Section ge0_integralT_even.
 Context {R : realType}.
 Let mu := @lebesgue_measure R.
--- a/theories/probability.v
+++ b/theories/probability.v
@@ -8,7 +8,7 @@ From mathcomp Require Import exp numfun lebesgue_measure lebesgue_integral.
 From mathcomp Require Import reals interval_inference ereal topology normedtype.
 From mathcomp Require Import sequences derive esum measure exp trigo realfun.
 From mathcomp Require Import numfun lebesgue_measure lebesgue_integral kernel.
-From mathcomp Require Import ftc gauss_integral.
+From mathcomp Require Import charge ftc gauss_integral.
 
 (**md**************************************************************************)
 (* # Probability                                                              *)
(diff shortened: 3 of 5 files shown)


classical/unstable.v
```diff
@@ -76,6 +76,35 @@ elim: r => [|a l]; first by rewrite !big_nil exp1n.
 by rewrite !big_cons; case: ifPn => // Pa <-; rewrite expnMn.
 Qed.
 
+Lemma leq_prod2 (x y n m : nat) : (n <= x)%N -> (m <= y)%N ->
+  (\prod_(m <= i < y) i * \prod_(n <= i < x) i <= \prod_(n + m <= i < x + y) i)%N.
```
hoheinzollern: This is a weirdly specific lemma with a generic name, I would fix the naming first.
Also, you might not need the two hypotheses `(n <= x)%N` and `(m <= y)%N`.

classical/unstable.v
```diff
@@ -76,6 +76,35 @@ elim: r => [|a l]; first by rewrite !big_nil exp1n.
...
+rewrite subnKC//.
+rewrite -[in leqLHS](add0n m) big_addn.
+rewrite [in leqRHS](_ : y - m = ((y - m + x) - x))%N; last first.
+  by rewrite -addnBA// subnn addn0.
+rewrite -[X in iota X _](add0n x) big_addn -addnBA// subnn addn0.
+by apply: leq_prod => i _; rewrite leq_add2r leq_addr.
+Qed.
+
+Lemma leq_fact2 (x y n m : nat) : (n <= x) %N -> (m <= y)%N ->
+  (x`! * y`! * ((n + m).+1)`! <= n`! * m`! * ((x + y).+1)`!)%N.
```
hoheinzollern, line 95: Similarly, you might be able to drop the two assumptions. This suggests that the previous lemma is a subproof of `leq_fact2`

theories/derive.v
```diff
@@ -1267,6 +1267,12 @@ Proof. by have /derivableP := @derivable_id x v; rewrite derive_val. Qed.
 
 End derive_id.
 
+Lemma derive1_onem {R : numFieldType} :
+  (fun x => 1 - x : R^o)^`()%classic = cst (-1).
```
hoheinzollern: `onem` typically recalls the syntax ``1-`, why is it not being used here?

theories/derive.v
```diff
@@ -1332,13 +1338,26 @@ rewrite scalerA -scalerDl mulrCA -[f x * _]exprS.
...
 
+(*Global Instance is_deriveX f n x v (df : R) :
+  is_derive x v f df -> is_derive x v (f ^+ n.+1) ((n.+1%:R * f x ^+n) *: df).
+Proof.
+move=> dfx; elim: n => [|n ihn]; first by rewrite expr1 expr0 mulr1 scale1r.
+rewrite exprS; apply: is_derive_eq.
+rewrite scalerA -scalerDl mulrCA -[f x * _]exprS.
+by rewrite [in LHS]mulr_natl exprfctE -mulrSr mulr_natl.
+Qed.*)
+
```
hoheinzollern: Commented code, either remove or add a human-readable comment for why this is here still.

theories/derive.v
```diff
@@ -1332,13 +1338,26 @@ rewrite scalerA -scalerDl mulrCA -[f x * _]exprS.
...
 Proof. by case: n => [_|n /derivableP]; [rewrite expr0|]. Qed.
 
 Lemma deriveX f n x v : derivable f x v ->
   'D_v (f ^+ n) x = (n%:R * f x ^+ n.-1) *: 'D_v f x.
 Proof. by move=> /derivableP df; rewrite derive_val. Qed.
 
+(*Lemma deriveX f n x v : derivable f x v ->
+  'D_v (f ^+ n.+1) x = (n.+1%:R * f x ^+ n) *: 'D_v f x.
+Proof. by move=> /derivableP df; rewrite derive_val. Qed.*)
+
```
hoheinzollern: Likewise

theories/ftc.v
```diff
@@ -1793,6 +1793,46 @@ Qed.
...
+Section integration_by_substitution_onem.
+Context {R : realType}.
+Let mu := (@lebesgue_measure R).
+Local Open Scope ereal_scope.
+
+Lemma integration_by_substitution_onem (G : R -> R) (r : R) :
+  (0 < r <= 1)%R ->
+  {within `[0%R, r], continuous G} ->
+  (\int[mu]_(x in `[0%R, r]) (G x)%:E =
+  \int[mu]_(x in `[(1 - r)%R, 1%R]) (G (1 - x))%:E).
```
hoheinzollern:
```suggestion
  \int[mu]_(x in `[0%R, r]) (G x)%:E =
  \int[mu]_(x in `[(`1-r)%R, 1%R]) (G (`1-x))%:E.
```
no need for parentheses, use of `onem` notation?

theories/ftc.v
```diff
@@ -1793,6 +1793,46 @@ Qed.
...
+      by move=> x; apply: (@continuousB _ R^o)  => //; exact: cvg_cst.
+    by under eq_fun do rewrite subKr; exact: cvg_id.
+  + by apply: cvg_at_left_filter; apply: (@cvgB _ R^o) => //; exact: cvg_cst.
+Qed.
+
+Lemma Rintegration_by_substitution_onem (G : R -> R) (r : R) :
+  (0 < r <= 1)%R ->
+  {within `[0%R, r], continuous G} ->
+  (\int[mu]_(x in `[0%R, r]) (G x) =
+  \int[mu]_(x in `[(1 - r)%R, 1%R]) (G (1 - x)))%R.
```
hoheinzollern:
```suggestion
  \int[mu]_(x in `[0%R, r]) (G x) =
  \int[mu]_(x in `[(`1- r)%R, 1%R]) (G (`1-x))%R.
```
similar suggestion as above. Note, I'm not actually sure this works, but calling a lemma `onem` and then not using the notation is confusing to me. Perhaps it should be the other way around, i.e. changing the lemma name

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
   by move=> k x/= [_ x01] _; rewrite lee_fin poisson_pmf_ge0.
 by move=> k Ysk; apply/measurableT_comp => //; exact: measurable_poisson_pmf.
 Qed.
+
+(**md lemmas about the function $x \mapsto (1 - x)^n$ *)
+(* TODO: move? *)
```
hoheinzollern: move: where to?

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+(**md about the function $x \mapsto x^a  (1 - x)^b$ *)
+Section XMonemX.
+Context {R : numDomainType}.
+Implicit Types (x : R) (a b : nat).
+
+Definition XMonemX a b := fun x => x ^+ a * `1-x ^+ b.
+
+Lemma XMonemX_ge0 a b x : x \in `[0, 1] -> 0 <= XMonemX a b x.
+Proof.
+by rewrite in_itv=> /andP[? ?]; rewrite mulr_ge0 ?exprn_ge0 ?subr_ge0.
```
hoheinzollern, line 2246: these look like lemmas that interval_inference should be able to handle.

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Lemma integrable_XMonemX_restrict a b :
+  mu.-integrable [set: R] (EFin \o XMonemX a.-1 b.-1 \_`[0,1]).
+Proof.
+rewrite -restrict_EFin; apply/integrable_restrict => //=.
+by rewrite setTI; exact: integrable_XMonemX.
+Qed.
+
+Lemma integral_XMonemX_restrict U a b :
+  (\int[mu]_(x in U) (XMonemX a b \_ `[0, 1] x)%:E =
+   \int[mu]_(x in U `&` `[0%R, 1%R]) (XMonemX a b x)%:E)%E.
```
hoheinzollern:
```suggestion
  \int[mu]_(x in U) (XMonemX a b \_ `[0, 1] x)%:E =
  \int[mu]_(x in U `&` `[0%R, 1%R]) (XMonemX a b x)%:E.
```
remove %E

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Section beta_fun.
+Context {R : realType}.
+Notation mu := (@lebesgue_measure _).
+Local Open Scope ring_scope.
+Local Notation XMonemX := (@XMonemX R).
+
+Definition beta_fun a b : R := \int[mu]_x (XMonemX a.-1 b.-1 \_`[0,1]) x.
+
+Lemma EFin_beta_fun a b :
+  ((beta_fun a b)%:E = \int[mu]_x (XMonemX a.-1 b.-1 \_`[0,1] x)%:E)%E.
```
hoheinzollern:
```suggestion
  (beta_fun a b)%:E = \int[mu]_x (XMonemX a.-1 b.-1 \_`[0,1] x)%:E.
```

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Proof.
+rewrite /beta_fun -Rintegral_mkcond.
+under eq_Rintegral do rewrite XMonemX0n.
+by rewrite Rintegral_onemXn.
+Qed.
+
+Lemma beta_fun11 : beta_fun 1 1 = 1.
+Proof. by rewrite (beta_fun1S O) invr1. Qed.
+
+Lemma beta_funSSS a b :
```
hoheinzollern: This name is a bit weird

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+    move=> x x01.
+    rewrite mulrA mulrC mulrA (mulrA _ a.+1%:R) -(mulrA (_ * _)%R).
+    over.
+  rewrite /=.
+  rewrite RintegralZl//=; last exact: integrable_XMonemX.
+  by rewrite -mulNrn -2!mulNr -invrN -mulNrn opprK (mulrC _ a.+1%:R).
+by rewrite Rintegral_mkcond.
+Qed.
+
+Lemma beta_funSS a b : beta_fun a.+1 b.+1 =
```
hoheinzollern: Similarly, neither this nor the previous lemma name suggests what variable is being incremented. Proposal: `SaSb`? though I'm not fully satisfied either.

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Lemma beta_pdf_le_beta_funV x : beta_pdf x <= (beta_fun a b)^-1.
+Proof.
+rewrite /beta_pdf ler_pdivrMr ?beta_fun_gt0// mulVf ?gt_eqF ?beta_fun_gt0//.
+by rewrite patchE; case: ifPn => //; rewrite inE/= => ?; exact: XMonemX_le1.
+Qed.
+
+Local Notation mu := lebesgue_measure.
+
+Let int_beta_pdf01 : (\int[mu]_(x in `[0%R, 1%R]) (beta_pdf x)%:E =
+                      \int[mu]_x (beta_pdf x)%:E :> \bar R)%E.
```
hoheinzollern: Similarly, there should be no need for %E

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Qed.
+
+Lemma integrable_beta_pdf : mu.-integrable [set: _] (EFin \o beta_pdf).
+Proof.
+apply/integrableP; split.
+  by apply/measurable_EFinP; exact: measurable_beta_pdf.
+under eq_integral.
+  move=> /= x _.
+  rewrite ger0_norm//; last by rewrite beta_pdf_ge0.
+  over.
```
hoheinzollern:
```suggestion
under eq_integral=> /= x _ do rewrite ger0_norm ?beta_pdf_ge0//.
```
should work?

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+  rewrite lerNl (@le_trans _ _ 0%R)// ?invr_ge0 ?beta_fun_ge0//.
+  by rewrite lerNl oppr0 beta_pdf_ge0.
+rewrite /beta_pdf ler_pdivrMr ?beta_fun_gt0//.
+rewrite mulVf ?gt_eqF ?beta_fun_gt0//.
+by rewrite patchE; case: ifPn => //; rewrite inE => ?; exact: XMonemX_le1.
+Unshelve. all: by end_near. Qed.
+
+End beta_pdf.
+
+(* TODO: move? *)
```
hoheinzollern: where?

theories/probability.v
```diff
@@ -2087,3 +2042,768 @@ apply: ge0_emeasurable_sum.
...
+Qed.
+
+Section integral_beta_prob.
+Context {R : realType}.
+Local Notation mu := (@lebesgue_measure R).
+
+Lemma integral_beta_prob a b f U : measurable U -> measurable_fun U f ->
+  (\int[beta_prob a b]_(x in U) `|f x| < +oo)%E ->
+  (\int[beta_prob a b]_(x in U) f x =
+   \int[mu]_(x in U) (f x * (beta_pdf a b x)%:E))%E.
```
hoheinzollern, line 2799:
```suggestion
  \int[beta_prob a b]_(x in U) f x =
  \int[mu]_(x in U) (f x * (beta_pdf a b x)%:E).
```

hoheinzollern (changes requested):
I've yet to check the changelog, but please consider the comments below,

classical/unstable.v
affeldt-aist: With `n := 1` and `x := 0`, then the goal becomes:
```coq
\prod_(m <= i < y) i <= \prod_(m.+1 <= i < y) i
```
with less terms on the right than on the left so some hypothesis seems needed about `n` and `x`.

classical/unstable.v
affeldt-aist: (Similarly as the `leq_prod2`, `n,m := 1` and `x, y := 0` cannot be proved.)

classical/unstable.v
affeldt-aist: proposing `leq_Mprod_prodD`

theories/derive.v
affeldt-aist: should be used indeed, fixed

theories/derive.v
affeldt-aist: Should have been removed before, fixed.

theories/derive.v
affeldt-aist: Fixed.

theories/ftc.v
affeldt-aist: Indeed, fixed.

theories/ftc.v
affeldt-aist: For the time being, I stick with `onem` but it is possible that this notation disappears in the future, we'll rename then.

theories/probability.v
affeldt-aist: I don't know. I leave it here for the time being, near its application.

theories/probability.v
affeldt-aist: `SnSm` maybe?

theories/probability.v
affeldt-aist: Let's try `beta_funSSnSm`.

theories/probability.v
affeldt-aist: It is actually needed because we are in `ring_scope` but the presence of both `%:E` and `%E` is disgracious, so let's `Open/Scope`.

theories/probability.v
affeldt-aist: Indeed, it does. Thanks.

theories/probability.v
affeldt-aist: We can actually get rid of `invr_nonneg_proof` and `invr_nonneg` with `interval_inference.v`, so nothing to move anymore.

theories/probability.v
affeldt-aist: For example by defining:
```coq
Lemma XMonemX_ge0 a b (x : {i01 R}) : 0 <= XMonemX a b x%:num.
Proof. by rewrite mulr_ge0// exprn_ge0// subr_ge0. Qed.
```
This is true but then it triggers many changes, for example, it leads to the statement:
```coq
Lemma bounded_XMonemX a b : [bounded XMonemX a b x | x in [set: {i01 R}]].
```
which fails with the error
```text
The term "globally [set: {i01 R}]" has type "set_system {i01 R}"
while it is expected to have type "set_system R" (cannot unify 
"{i01 R}" and "Num.NumDomain.sort R").
```
because (I think) `bounded_near` requires some numeric type that `{i01 R}` does not have yet.
I am afraid a more systematic use of `{i01 R}` could trigger too many changes. @proux01

affeldt-aist:
(the last commit is actually fixing a problem in master...)

theories/probability.v
hoheinzollern: OK, good for now

PR #1646, opened 2025-06-20
Alternative Proof of Gauss Integral with improper integral

IshiguroYoshihiro:
I added an alternative proof of gauss integral and placed it in theories/showcase.
This is an experimental approach of improper integral.
The difference of proof strategy is that this version mainly relies on lemmas related to integration over unbounded intervals.

[NB(rei): this is the continuation of PR #1584 ]

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/gauss_integral.v
+++ b/theories/gauss_integral.v
@@ -62,7 +62,7 @@ Let mu : {measure set _ -> \bar R} := @lebesgue_measure R.
 
 Definition integral0y_gauss := \int[mu]_(x in `[0%R, +oo[) gauss_fun x.
 
-Let integral0y_gauss_ge0 : 0 <= integral0y_gauss.
+Lemma integral0y_gauss_ge0 : 0 <= integral0y_gauss.
 Proof. by apply: Rintegral_ge0 => //= x _; rewrite gauss_fun_ge0. Qed.
 
 Definition integral0_gauss x := \int[mu]_(t in `[0, x]) gauss_fun t.
@@ -81,10 +81,10 @@ Qed.
 
 Definition u x t := expR (- x ^+ 2 * oneDsqr t) / oneDsqr t.
 
-Let u_ge0 x t : 0 <= u x t.
+Lemma u_ge0 x t : 0 <= u x t.
 Proof. by rewrite /u divr_ge0// ?expR_ge0. Qed.
 
-Let measurable_u x : measurable_fun setT (u x).
+Lemma measurable_u x : measurable_fun setT (u x).
 Proof.
 apply: measurable_funM => //=.
   apply: measurableT_comp => //=; apply: measurable_funM => //=.
@@ -95,7 +95,7 @@ Qed.
 
 Local Notation "'d1 f" := (partial1of2 f).
 
-Let partial1_u x t : ('d1 u) x t = - 2 * x * gauss_fun x * gauss_fun (t * x).
+Lemma partial1_u x t : ('d1 u) x t = - 2 * x * gauss_fun x * gauss_fun (t * x).
 Proof.
 rewrite partial1of2E /u /= deriveMr//= -derive1E.
 rewrite derive1_comp// [in X in _ * (_ * X)]derive1Mr//.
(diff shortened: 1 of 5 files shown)


theories/gauss_integral.v
```diff
@@ -81,10 +81,10 @@ Qed.
 
 Definition u x t := expR (- x ^+ 2 * oneDsqr t) / oneDsqr t.
```
t6s, line 82: It looks horrible to pollute the global namespace by a single-letter identifier.
`Local Definition` may suffice?
(And similarly `Local Lemma` for lemmas with "u" in their names below)

affeldt-aist: Note that it is in a `Module`.

t6s: oops, I overlooked that

theories/showcase/gauss_integral_unbounded.v
```diff
@@ -0,0 +1,590 @@
...
+  by apply: cvgV => //; exact: continuous_expR.
+- by move=> x x1; rewrite derive1E derive_val mulrN1 opprK.
+by rewrite -EFinB ltry.
+Qed.
+
+End gauss_integral_preliminaries.
+
+Section u_properties.
+
+Lemma integral0y_u0 :
```
affeldt-aist, line 90: The function `u` is internal to the module `gauss_integral_proof`.
It therefore seems strange to introduce a lemma `integral0y_u0` that exposes the identifier `u`.
Maybe this new lemma should itself be inside a module.

theories/showcase/gauss_integral_unbounded.v
```diff
@@ -0,0 +1,590 @@
...
+  by rewrite lerNl oppr0 sqr_ge0.
+by rewrite integral0y_u0 ltry.
+Qed.
+
+End u_properties.
+
+Section gauss_dominates.
+Local Open Scope ring_scope.
+
+Definition max_x : R := (Num.sqrt 2)^-1.
```
affeldt-aist, line 124: `max_x` looks ill-names: it talks about `x` which is undefined and sounds generic even though it is technical.

theories/showcase/gauss_integral_unbounded.v
```diff
@@ -0,0 +1,590 @@
...
+From mathcomp Require Import lebesgue_measure numfun lebesgue_integral.
+From mathcomp Require Import interval_inference real_interval realfun derive.
+From mathcomp Require Import trigo ftc gauss_integral.
+
+(**md**************************************************************************)
+(* # Improper Integral                                                        *)
+(* # calculating gauss integrals by limit                                     *)
+(* ref: https://www.phys.uconn.edu/~rozman/Courses/P2400_17S/                 *)
+(*                                            downloads/gaussian-integral.pdf *)
+(*  u (x : R) (y : R) : R == a function dominates gauss_fun over `[0, +oo[    *)
```
affeldt-aist, line 13: The function `u` is not really defined in this file. Maybe this piece of documentation should go to `gauss_integral.v`.

affeldt-aist: And maybe the function `u` in `gauss_integral.v` could be renamed to better match its documentation. By the way, what "dominates" mean here?

theories/showcase/gauss_integral_unbounded.v
```diff
@@ -0,0 +1,590 @@
...
+End u_properties.
+
+Section gauss_dominates.
+Local Open Scope ring_scope.
+
+Definition max_x : R := (Num.sqrt 2)^-1.
+
+Let max_x_ge0 : (0 <= max_x). Proof. by rewrite invr_ge0. Qed.
+
+Definition max_y : R := 2 / Num.sqrt 2 / Num.sqrt (expR 1).
```
affeldt-aist, line 128: Like `max_x`, `may_y` looks ill-named. It should also be accompanied by a piece of documentation if it is intended to be exposed. It suffices to explain its role in the proof.

PR #1661, opened 2025-06-28
generalize pushforward

t6s:

The second (explicit) argument of `pushforward` has been a proof `mf : measurable_fun f`,
unlike the first `m : set T1 ->\barR` that is just a function.
This proof object is not needed to define a pushforward itself, only used when instantiating
a measure structure.

This PR generalizes the definition so that the second argument becomes a function `f : T1 -> T2`.
This change clarifies the definition itself and simplifies a form of its invocation where `f` is passed explicitly
(only one case so far, but I expect more in the future).

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -113,6 +113,7 @@
   + lemma `ae_eq_comp2`
   + lemma `ae_foralln`
   + lemma `ae_eqe_mul2l`
+  + definition `pushforward` (to take a function instead of a proof)
 
 - new file `ess_sup_inf.v`:
   + lemma `measure0_ae`
--- a/theories/measure.v
+++ b/theories/measure.v
@@ -157,9 +157,9 @@ From mathcomp Require Import sequences esum numfun.
 (*                                                                            *)
 (* ## Instances of measures                                                   *)
 (* ```                                                                        *)
-(*  pushforward m mf == pushforward/image measure of m by f, where mf is a    *)
-(*                      proof that f is measurable                            *)
-(*                      m has type set T -> \bar R.                           *)
+(*   pushforward m f == pushforward of a set function m : set T1 -> \bar R    *)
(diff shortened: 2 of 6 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -113,6 +113,7 @@
   + lemma `ae_eq_comp2`
   + lemma `ae_foralln`
   + lemma `ae_eqe_mul2l`
+  + definition `pushforward` (to take a function instead of a proof)
```
affeldt-aist, line 116: Note for later: In general, it seems that it is better to only append stuff in the CHANGELOG to reduce conflicts.

affeldt-aist:
Mmmhhh... I remember that this was needed in the past, maybe something to do with the introduction of `{mfun T >-> T'}` or advances with Hierarchy-Builder. Anyway, CI is green and that's definitely a step forward.

affeldt-aist (approved):

affeldt-aist:
I merge soon (maybe @CohenCyril has a comment).

theories/measure.v
```diff
@@ -2272,7 +2272,7 @@ apply: measure_semi_sigma_additive.
 Qed.
 
 HB.instance Definition _ := isMeasure.Build _ _ _
-  (pushforward m mf) pushforward0 pushforward_ge0 pushforward_sigma_additive.
+  (pushforward m f) pushforward0 pushforward_ge0 pushforward_sigma_additive.
```
CohenCyril, line 2275: This PR probably makes this instance unusable from outside the section

CohenCyril (changes requested):
The CI passing probably means there is no real usage of push forwards in the CI, or that I'm wrong. More later

theories/measure.v
CohenCyril: Nevermind, it does work!

CohenCyril (approved):

PR #1679, opened 2025-07-07
is_derive1_sqrt lemma 

yosakaon:
Co-authored with @affeldt-aist 
This PR introduces the lemma is_derive1_sqrt, establishing the derivability of the square root function at any strictly positive real number and provides its derivative formula.
This result is essential to prove of the derivability of the norm function and such calculations.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/realfun.v
+++ b/theories/realfun.v
@@ -1763,6 +1763,42 @@ Unshelve. all: by end_near. Qed.
 
 End is_derive_inverse.
 
+Global Instance is_derive1_sqrt {K : realType} (x : K) : 0 < x -> is_derive x 1 Num.sqrt (2 * Num.sqrt x)^-1.
+Proof.
+move=> x_gt0.
+have sqrtK : {in Num.pos, cancel (@Num.sqrt K) (fun x => x ^+ 2)}.
+  by move=> a a0; rewrite sqr_sqrtr// ltW.
+rewrite -[x]sqrtK//.
+apply: (@is_derive_inverse K (fun x => x ^+ 2)).
+- near=> z.
+  rewrite sqrtr_sqr gtr0_norm//.
+  have [xz|zx|->] := ltgtP z (Num.sqrt x); last first.
+  + by rewrite sqrtr_gt0.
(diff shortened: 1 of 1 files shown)


theories/realfun.v
```diff
@@ -1763,6 +1763,42 @@ Unshelve. all: by end_near. Qed.
 
 End is_derive_inverse.
 
+Global Instance is_derive1_sqrt {K : realType} (x : K) : 0 < x -> is_derive x 1 Num.sqrt (2 * Num.sqrt x)^-1.
```
t6s: this line is too long

```suggestion
Global Instance is_derive1_sqrt {K : realType} (x : K) :
  0 < x -> is_derive x 1 Num.sqrt (2 * Num.sqrt x)^-1.
```

t6s (approved):
looks good

classical/unstable.v
```diff
@@ -483,3 +483,6 @@ Notation " R ~~> R' " := (@respectful _ _ (Program.Basics.flip (R%signature)) (R
 
 Export -(notations) Morphisms.
 End ProperNotations.
+
+Lemma sqrK {K : rcfType} : {in Num.pos, cancel (@Num.sqrt K) (fun x => x ^+ 2)}.
+Proof. by move=> r r0; rewrite sqr_sqrtr// ltW. Qed.
```
CohenCyril, line 488: This looks redundant with `sqr_sqrt`, (and slightly weaker since `Num.pos` can be weakened to `Num.nneg`).
If kept, it should be generalized and renamed `sqrtK` (the prefix of a `K` lemma is the section / thing being cancelled)

affeldt-aist: Sorry for having been too hasty, will fix asap in a new PR.

CohenCyril: (an in the long run, we should actually replace `sqr_sqrt` which has an equivalent statement written in an expanded way and which name is not as canonical as `sqrtK`)

PR #1683, opened 2025-07-07
fix sqrtK

affeldt-aist:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/realfun.v
+++ b/theories/realfun.v
@@ -1766,19 +1766,19 @@ End is_derive_inverse.
 Global Instance is_derive1_sqrt {K : realType} (x : K) : 0 < x ->
   is_derive x 1 Num.sqrt (2 * Num.sqrt x)^-1.
 Proof.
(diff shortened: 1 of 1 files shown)


affeldt-aist:
A bit redundant with `sqr_sqrtr` but the name is so natural that it is not the first time I have been looking for it.

theories/realfun.v
```diff
@@ -1767,7 +1767,7 @@ Global Instance is_derive1_sqrt {K : realType} (x : K) : 0 < x ->
   is_derive x 1 Num.sqrt (2 * Num.sqrt x)^-1.
 Proof.
 move=> x0.
```
CohenCyril: I would do this instead and leave the rest unchanged.
```suggestion
move=> x_gt0; have x_ge0 := ltW x_gt0.
```

PR #1705, opened 2025-08-21
Divergence of the sum of the reciprocal of primes

Tragicus:
The divergence of the sum of the reciprocal of primes theorem, extracted from https://github.com/math-comp/analysis/pull/1690 because it has been ready and sitting there for too long. We can add it to [the famous list of formalized theorems](https://www.cs.ru.nl/~freek/100/).

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -138,7 +138,7 @@ Qed.
 Definition monotonous d (T : porderType d) (pT : predType T) (A : pT) (f : T -> T) :=
   {in A &, {mono f : x y / (x <= y)%O}} \/ {in A &, {mono f : x y /~ (x <= y)%O}}.
 
-Lemma mono_ext f : {mono f : m n / (m <= n)%N} -> forall n, (n <= f n)%N.
+Lemma mono_leq_infl f : {mono f : m n / (m <= n)%N} -> forall n, (n <= f n)%N.
 Proof.
 move=> fincr; elim=> [//| n HR]; rewrite (leq_ltn_trans HR)//.
 by rewrite ltn_neqAle fincr (inj_eq (incn_inj fincr)) -ltn_neqAle.
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -35,48 +35,45 @@ Unset Printing Implicit Defensive.
 Import Order.TTheory GRing.Theory Num.Theory.
 Local Open Scope ring_scope.
 
-Lemma coprime_prodr (I : Type) (r : seq I) (P : {pred I}) (F : I -> nat) (a : I) (l : seq I) :
+Lemma coprime_prodr (I : Type) (r : seq I) (P : {pred I}) (F : I -> nat) (a : I)
+    (l : seq I) :
   all (coprime (F a)) [seq F i | i <- [seq i <- l | P i]] ->
   coprime (F a) (\prod_(j <- l | P j) F j).
 Proof.
-elim: l => /= [_|a0 l HI].
-  by rewrite big_nil coprimen1.
-rewrite big_cons. case: (P a0) => //.
-rewrite map_cons => /= /andP [] ? ?.
-rewrite coprimeMr. apply/andP. split=> //.
-exact: HI.
+elim: l => /= [_|h t ih]; first by rewrite big_nil coprimen1.
+rewrite big_cons; case: ifPn => // Ph.
+rewrite map_cons => /= /andP[FaFh FatP].
+by rewrite coprimeMr FaFh/= ih.
 Qed.
 
-Lemma Gauss_dvd_prod (I : eqType) (r : seq I) (P : {pred I}) (F : I -> nat) (n : nat) :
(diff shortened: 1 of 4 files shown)


classical/unstable.v
```diff
@@ -97,6 +138,12 @@ Qed.
 Definition monotonous d (T : porderType d) (pT : predType T) (A : pT) (f : T -> T) :=
   {in A &, {mono f : x y / (x <= y)%O}} \/ {in A &, {mono f : x y /~ (x <= y)%O}}.
 
+Lemma mono_ext f : {mono f : m n / (m <= n)%N} -> forall n, (n <= f n)%N.
```
affeldt-aist: There is maybe a better name for `mono_ext`.
The identifier does not show it is about natural numbers
or that the conclusion is `leq`; what does `ext` mean?
What about `mono_leq`?

theories/showcase/pnt.v
```diff
@@ -0,0 +1,456 @@
...
+
+End prime_seq.
+
+Section DivergenceSumInversePrimeNumbers.
+
+Let P (k N : nat) :=
+  [set n : 'I_N.+1 |all (fun p => p < prime_seq k) (primes n)]%SET.
+Let G (k N : nat) := ~: P k N.
+
+Fact cardPcardG k N : #|G k N| + #|P k N| = N.+1.
```
affeldt-aist: `Fact cardPcardG` -> `Local Lemma cardPcardG` or `Let cardPcardG`?
Since it does not appear in the changelog, I guess this is the intent.

theories/showcase/pnt.v
```diff
@@ -0,0 +1,456 @@
...
+Proof.
+rewrite addnC.
+have : (P k N) :|: (G k N) = [set : 'I_N.+1]%SET by rewrite finset.setUCr.
+rewrite -cardsUI finset.setICr cards0.
+by rewrite -[X in _ + _ = X]card_ord addn0 -cardsT => ->.
+Qed.
+
+Fact cardG (R : realType) (k N : nat) :
```
affeldt-aist: `Fact cardG` -> `Local Lemma cardG` or `Let cardG`?
(same as just above)

theories/showcase/pnt.v
```diff
@@ -0,0 +1,456 @@
...
+  by rewrite ltnSn -ltnS ltn_subrL prime_gt0 // -mem_prime_seq mem_range.
+rewrite inE /in_mem /= => /mapP /= [] x3.
+rewrite mem_iota subnK => [/andP [] x3b1 x3b2 /(congr1 val)|]; last first.
+  by rewrite -subn_eq0; apply/eqP /Eigtpi.
+have x3b3: x3 < N.+1 by apply: (leq_ltn_trans _ (ltn_ord x1)).
+rewrite val_insubd x3b3 /= => x2eqx3. move: x3b2.
+by rewrite ltnS -x2eqx3.
+Qed.
+
+Fact cardP (k : nat) :
```
affeldt-aist: `Fact cardP` -> `Local Lemma cardP` or `Let cardP`?
Since this one does not depend on any `R : realType`
(it was wrongly flagged as such in the previous commit),
it could maybe be in a different section, or maybe
appear before the other intermediate lemmas, to highlight its different nature.

theories/showcase/pnt.v
```diff
@@ -0,0 +1,456 @@
...
+  rewrite !inE => /imsetP [] x xinPkN -> /=.
+  apply: (contra_not_neq (@congr1 _ _ val _ _)) => /=.
+  apply/eqP; rewrite -lt0n; apply: prodn_gt0 => i.
+  by rewrite expn_gt0 prime_gt0//= -mem_prime_seq inE.
+apply: (leq_trans imfleqAB).
+rewrite cardsX cardsE card_tuple card_bool cardsC1 card_ord.
+by rewrite -expnD addnA addnn.
+Qed.
+
+Theorem DivergenceSumInversePrimeNumbers (R : realType) :
```
affeldt-aist: `DivergenceSumInversePrimeNumbers` -> `dvg_sum_inv_prime_seq`?
because I think that we favor small letters and underscores
instead of caps and identifiers made up of other identifiers
appearing in the statement.

classical/unstable.v
Tragicus: Oh, my bad, these functions are called "inflationary", not something like "extensive". I feel like `mono_leq` is too general. Maybe something like `mono_leq_infl`?

theories/showcase/pnt.v
Tragicus: I do not really see the point in putting it in a different section, it is the natural course of the proof of the theorem and it just so happens to not require a hypothesis about real numbers.

affeldt-aist: Moreover if this turns into a `Let` or a `Local Lemma`, that indeed justifies to keep it in the same section.

PR #1718, opened 2025-09-16
Normed vector types, infinite norm, norm equivalence thm, continuity …

Tragicus:
Continuity of linear functions in finite dimension

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -24,6 +24,14 @@ From mathcomp Require Import vector archimedean interval.
 (*                           the dependent sum                                *)
 (*                prodA x := sends (X * Y) * Z to X * (Y * Z)                 *)
 (*               prodAr x := sends X * (Y * Z) to (X * Y) * Z                 *)
+(*           isSemiNorm f == f : K -> L is a seminorm                         *)
+(*                           K is a numDomainType.                            *)
+(*                           L is a lmodType K.                               *)
+(*                           The HB class is SemiNorm.                        *)
+(*               isNorm f == f : K -> L is a norm                             *)
+(*                           K is a numDomainType.                            *)
+(*                           L is a lmodType K.                               *)
+(*                           The HB class is Norm.                            *)
 (* ```                                                                        *)
 (*                                                                            *)
 (******************************************************************************)
@@ -599,20 +607,26 @@ Proof. exact: real_ltrNbound. Qed.
    its scalar field. *)
 Module Norm.
 
-HB.mixin Record isNorm (K : numDomainType) (L : lmodType K) (norm : L -> K) := {
-  norm0 : norm 0 = 0;
-  norm_ge0 : forall x, 0 <= norm x;
-  norm0_eq0 : forall x, norm x = 0 -> x = 0;
-  ler_normD : forall x y, norm (x + y) <= norm x + norm y;
-  normZ : forall r x, norm (r *: x) = `|r| * norm x;
+HB.mixin Record isSemiNorm (K : numDomainType) (L : lmodType K) (norm : L -> K) := {
+  norm0 : norm 0 = 0 ;
+  norm_ge0 : forall x, 0 <= norm x ;
+  ler_normD : forall x y, norm (x + y) <= norm x + norm y ;
+  normZ : forall r x, norm (r *: x) = `|r| * norm x
 }.
 
 #[export]
-HB.structure Definition Norm K L := { norm of @isNorm K L norm }.
+HB.structure Definition SemiNorm K L := { norm of @isSemiNorm K L norm }.
+
+HB.mixin Record SemiNorm_isNorm (K : numDomainType) (L : lmodType K)
+  (norm : L -> K) := { norm0_eq0 : forall x, norm x = 0 -> x = 0 }.
+
+#[export]
+HB.structure Definition Norm K L :=
+  { norm of @SemiNorm_isNorm K L norm & @SemiNorm K L norm }.
 
 Module Import Theory.
 Section Theory.
-Variables (K : numDomainType) (L : lmodType K) (norm : Norm.type L).
+Variables (K : numDomainType) (L : lmodType K) (norm : SemiNorm.type L).
 
 Lemma normMn x n : norm (x *+ n) = norm x *+ n.
 Proof. by rewrite -scaler_nat normZ normr_nat mulr_natl. Qed.
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -1,6 +1,7 @@
 (* mathcomp analysis (c) 2026 Inria and AIST. License: CeCILL-C.              *)
+From HB Require Import structures.
 From mathcomp Require Import all_ssreflect_compat finmap ssralg ssrnum ssrint.
-From mathcomp Require Import archimedean interval.
+From mathcomp Require Import vector archimedean interval.
 
 (**md**************************************************************************)
 (* # MathComp extra                                                           *)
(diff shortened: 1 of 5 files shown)


theories/normedtype_theory/normed_module.v
```diff
@@ -2079,3 +2086,241 @@ by rewrite interior_closed_ballE //; exact: ballxx.
 Qed.
 
 End Closed_Ball_normedModType.
+
+Section InfiniteNorm.
+Variables (R : numFieldType) (V : vectType R) (B : (\dim (@fullv _ V)).-tuple V).
+Let V' := @fullv _ V.
+Hypothesis (Bbasis : basis_of V' B).
+
+Definition oo_norm x := \big[Order.max/0]_(i < \dim V') `|coord B i x|.
```
affeldt-aist: Is it really a good name? Because it is a bit like using a notation inside an identifier, also we have been using `oo` to mean "open-open" in MathComp (when talking about intervals). What about `infty_norm`? (like in LaTeX). Possibly coupling it with a notation using `+oo` in some way.

theories/normedtype_theory/normed_module.v
```diff
@@ -2079,3 +2086,241 @@ by rewrite interior_closed_ballE //; exact: ballxx.
...
+  exact: segment_compact.
+apply/continuous_subspaceT/sum_continuous => i _/= x.
+exact/continuousZl/proj_continuous. 
+Qed.
+
+Lemma equivalence_norms (N : V -> R) :
+  N 0 = 0 -> (forall x, 0 <= N x) -> (forall x, N x = 0 -> x = 0) ->
+  (forall x y, N (x + y) <= N x + N y) ->
+  (forall r x, N (r *: x) = `|r| * N x) ->
+  exists M, 0 < M /\ forall x : Voo, `|x| <= M * N x /\ N x <= M * `|x|.
```
CohenCyril: I think this deserve an abstraction of norm and an abstraction of norm comparison :)

theories/normedtype_theory/normed_module.v
affeldt-aist: `max_norm` is maybe also an option

CohenCyril: "Inifinity norm" or "supremum norm" (or "norme sup" or "norme infinie" in french) is the traditional name.

affeldt-aist: `sup_norm`?

classical/unstable.v
```diff
@@ -591,3 +592,114 @@ Proof. exact: real_ltr_bound. Qed.
...
+rewrite /max_norm; elim: (index_enum _) (mem_index_enum i) => //= j l IHl.
+rewrite inE big_cons [X in _ <= X _ _]/Order.max/= => /predU1P[<-|/IHl {}IHl];
+  case: ifP => [/ltW|]// /negbT.
+set b := (X in _ < X).
+have bR : b \is Num.real by apply: bigmax_real => // a _; apply: normr_real.
+have /comparable_leNgt <- := real_comparable bR (normr_real (coord B j x)).
+by move=> /(le_trans IHl).
+Qed.
+
+Lemma max_norm0 : max_norm 0 = 0.
```
affeldt-aist: Shouldn't this lemma by declared as a `Let`? (same comment for the following lemmas in this section)

classical/unstable.v
```diff
@@ -591,3 +592,114 @@ Proof. exact: real_ltr_bound. Qed.
...
+End Norm.
+Export Norm.Exports.
+
+Section InfiniteNorm.
+Variables (K : numFieldType) (V : vectType K).
+Let V' := @fullv _ V.
+Variable B : (\dim V').-tuple V.
+Hypothesis Bbasis : basis_of V' B.
+
+Definition max_norm x := \big[Order.max/0]_(i < \dim V') `|coord B i x|.
```
affeldt-aist: Maybe `max_norm` should be documented in the header (maybe also `max_space` just below)?

classical/unstable.v
```diff
@@ -591,3 +592,114 @@ Proof. exact: real_ltr_bound. Qed.
...
+Variables (K : numDomainType) (L : lmodType K) (norm : Norm.type L).
+
+Lemma normMn x n : norm (x *+ n) = norm x *+ n.
+Proof. by rewrite -scaler_nat normZ normr_nat mulr_natl. Qed.
+
+Lemma normN x : norm (- x) = norm x.
+Proof. by rewrite -scaleN1r normZ normrN1 mul1r. Qed.
+
+Lemma ler_norm_sum (I : Type) (r : seq I) (F : I -> L) :
+    norm (\sum_(i <- r) F i) <= \sum_(i <- r) norm (F i).
```
affeldt-aist:
```suggestion
  norm (\sum_(i <- r) F i) <= \sum_(i <- r) norm (F i).
```

affeldt-aist:
For example, `Zmodule_isSemiNormed` in MathComp provides a similar interface with the difference that `norm` is one of the field of the record (whereas here it is the subject). I guess the apparent duplication is unavoidable but doesn't it deserve a comment in the code?

classical/unstable.v
```diff
@@ -592,3 +598,117 @@ Proof. exact: real_ltr_bound. Qed.
...
+End Norm.
+Export Norm.Exports.
+
+Section InfiniteNorm.
+Variables (K : numFieldType) (V : vectType K).
+Let V' := @fullv _ V.
+Variable B : (\dim V').-tuple V.
+Hypothesis Bbasis : basis_of V' B.
+
+Definition max_norm x := \big[Order.max/0]_(i < \dim V') `|coord B i x|.
```
Tragicus: Inference of a Norm.type for `max_norm` fails because the instance requires `B` to be a basis. However, it feels wrong to me to add the hypothesis in the definition of `max_norm`. What do you think?

affeldt-aist: What about adding the hypothesis "temporarily"? Just leave an issue to be addressed later or add a warning to indicate to the user that it is experimental. I apologize for proposing this option without understanding completely the problem but I was under the impression that other users (@mkerjean , @agontard) also have an interest in the abstraction.

CohenCyril: Uh oh... this is more than fishy! This definition does depend on a basis (and not only the fact that the given tuple is a basis) and is thus morally not a valid vector space definition, which ought to be invariant wrt to the choice of a basis, and this one is clearly not. This is exactly why this definition does depend on a choice of basis. What it does is build the (unique) normed space on top of V that makes the given basis normal. This means that `max_norm` is probably not the right name (it would be the case when given a function, finite function or tuple with values in a normed vector space), probably the right name is something like `norm_from_basis B Bbasis` and `normedVect_from_basis B Basis` .

classical/unstable.v
```diff
@@ -592,3 +598,117 @@ Proof. exact: real_ltr_bound. Qed.
...
+End Theory.
+End Theory.
+
+Module Import Exports. HB.reexport. End Exports.
+End Norm.
+Export Norm.Exports.
+
+Section InfiniteNorm.
+Variables (K : numFieldType) (V : vectType K).
+Let V' := @fullv _ V.
```
CohenCyril: Convention is:
```suggestion
Variables (K : numFieldType) (vT : vectType K).
Let V' := {:vT}
```
And `V'` can be inlined.

classical/unstable.v
CohenCyril: However we can do slightly better: indeed we can amend the given candidate `B`, keeping the smallest initial family that is generating, pruning it in order to make it free, completing it thus making a basis out of it. If `B` was indeed a basis, the tuple would stay the same, otherwise we have a non canonical normed vector space, but that's life. It can still be useful though, because if we give it a free family, the generated norm will be unique on the vector space generated by this family, which is what a user might expect. I believe this is the right way to go.

CohenCyril: We can probably go further and generate a Hilbert space :)

Tragicus: Giving a norm is a non-canonical choice, e.g. choosing a Euclidean norm entails choosing which bases are normal, so it is  not surprising that I have to make such a choice to define my norm. I am not sure I understand the point about "vector space definitions", this is not what I am doing here. Besides, the definition of vector spaces in `vector.v` does choose a basis, via an isomorphism with a matrix space.
It may be a consequence of the above, but I do not understand what you would like to call `max_norm`. In any case, `norm_from_basis` does not say which norm we are considering, maybe `max_norm_from_basis`.
With your optimization, I would still need to assume that `B` is generating, so I am not sure what I gain.

CohenCyril: No you do not have to assume `B` is generating, you can complete it to be generating. (I was tired when I wrote my previous messages)

Tragicus: Ah, true, and in that case I might as well start from an empty `B`, i.e. get rid of `B` completely.

CohenCyril: No... please don't do that. :crying_cat_face:  we have to be able to provide a basis!

CohenCyril: We have to be able to mention a basis that is normed.

CohenCyril: Indeed it is not useful for your application, but it is nice to control which class of basis become normed.

To answer your other question the `max_norm` I had in mind would be the norm of a product of normed vector spaces, e.g. `vT ^ i` or more generally `{dffun forall i : 'I_n, vT_ i}` where `vT_ : normedVectType^i`.

classical/unstable.v
```diff
@@ -592,3 +593,42 @@ Proof. exact: real_ltr_bound. Qed.
 
 Lemma ltrNbound {R : archiRealDomainType} (x : R) : - x < (Num.bound x)%:R.
 Proof. exact: real_ltrNbound. Qed.
+
+(* normedZmodType provide norms but the subject is not the norm. We define here
```
affeldt-aist, line 605:
```suggestion
(* normedZmodType provides norms but the subject is not the norm. We define here
```

affeldt-aist:
I made a last commit essentially to introduce semi-norms because there is a need for them and will merge soon unless something comes up.

affeldt-aist (approved):

PR #1736, opened 2025-10-21
Remove MC2.4 deprecations and cleanup mathcomp_extra.v

pi8027:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/reals_stdlib/Rstruct.v
+++ b/reals_stdlib/Rstruct.v
@@ -543,7 +543,7 @@ Qed.
 Lemma factE n : fact n = n`!.
 Proof. by elim: n => //= n ih; rewrite factS mulSn ih. Qed.
 
-(*Section bigmaxr.
+Section bigmaxr.
 Context {R : realDomainType}.
 
 (* bigop pour le max pour des listes non vides ? *)
@@ -780,6 +780,6 @@ Proof.
 by move=> inj_f i; rewrite /Num.leif bmaxrf_ler -(inj_eq inj_f) eq_index_bmaxrf.
 Qed.
 
-End bigmaxr.*)
+End bigmaxr.
 
 End ssreal_struct_contd.


reals_stdlib/Rstruct.v
```diff
@@ -551,20 +551,30 @@ Context {R : realDomainType}.
 Definition bigmaxr (r : R) s := \big[Num.max/head r s]_(i <- s) i.
 
 #[deprecated(note="To be removed. Use topology.v's bigmax/min lemmas instead.")]
+#[warning="-deprecated"]
```
pi8027, line 554: I added these attributes (see below) since I prepared this PR by adding the following lines to `_CoqProject`:
```
-arg -w -arg +deprecated-reference
-arg -w -arg +deprecated-syntactic-definition
-arg -w -arg +deprecated-notation
```
Since they are the internal use of deprecated definitions/lemmas, I think that suppressing the deprecation warning here is the right thing to do, but feel free to revert this part of the changes.

Since these deprecations date back to 3 years ago, removing all of them is also an option.

classical/fsbigop.v
```diff
@@ -316,6 +316,7 @@ Proof.
 move=> finF; elim/Peq : R => R in zero times plus a F finF *.
 have [->|a0] := eqVneq a zero.
   by rewrite Monoid.mul0m fsbig1//; move=> i _; rewrite Monoid.mul0m.
+#[warning="deprecated"] (* FIXME *)
 rewrite big_distrr [RHS](full_fsbigID (F @^-1` [set zero])); last first.
```
pi8027, line 320: I didn't manage to get rid of `full_fsbigID` here.

affeldt-aist (approved):
Thanks for having taken care of that!
As for the warnings in `Rstruct.v`, I don't have a strong opinion, apart from the fact that this file should be cleaned. :-(
The only development that I can think of that might depend on that is Damien and Cyril's work on LaSalle and the pendulum @CohenCyril . We have recently ported the latter to the last version of MathComp-Analysis, so it will be easy to confirm soon whether it is really needed or not and do cleaning accordingly (I'll try to do that asap).

pi8027:
> From the above discussion, I did not manage to understand the exact cause of the failure to replace deprecated defs/notations/lemmas by non deprecated ones. Can someone explain?

The ones in `Rstruct.v` are a deprecated definition `bigmaxr` and some deprecated lemmas about it. These lemmas depend on `bigmaxr` and some other deprecated lemmas. I consider them as an "internal use" of deprecated definitions and lemmas, and thus, I think that these deprecation warnings are safe to suppress.

To be clear, it's also ok to remove these `#[warning="-deprecated"]`.

pi8027:
IMO, the compilation of MCA produces too many warnings, and it makes it hard to understand which ones have to be fixed. So, if I'm the one maintaining it, I would suppress the ones that are totally safe to ignore, e.g., the ones about `unstable.v`.

affeldt-aist:
I am in favor of merging almost as it is (I'll maybe remove the `#[warning="-deprecated"]` in `Rstruct.v`: it looks a bit too verbose for something that could very well disappear as soon as someone really looks into it). As for the quantity of warnings, I am ok with a status-quo: we are not overwhelmed and, for having recently ported Damien's development, I found the warnings helpful (also, we rely on unstable.v in infotheo).

PR #1744, opened 2025-10-30
fixes #1733 (`ln_eq0`)

### LINKED ISSUE #1733: `ln_eq0`
https://github.com/math-comp/analysis/blob/6600f40c3ab9d8e342f1991e889599b1528a0255/theories/exp.v#L1247

we maybe need the analogous lemma for real numbers

@garrigue @gregweng

affeldt-aist:

fixes #1733 

what about this fix?

@garrigue @gregweng

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/exp.v
+++ b/theories/exp.v
@@ -784,15 +784,11 @@ have [x0|x0 x1] := leP x 0; first by rewrite ln0.
 by rewrite -ler_expR expR0 lnK.
 Qed.
 
-Lemma ln_eq0 x : (ln x == 0) = (x <= 0) || (x == 1).
+Lemma ln_eq0 x : 0 < x -> (ln x == 0) = (x == 1).
 Proof.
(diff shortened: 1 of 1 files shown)


theories/exp.v
```diff
@@ -785,6 +784,17 @@ have [x0|x0 x1] := leP x 0; first by rewrite ln0.
 by rewrite -ler_expR expR0 lnK.
 Qed.
 
+Lemma ln_eq0 x : (ln x == 0) = (x <= 0) || (x == 1).
```
t6s: I am afraid of respecting the default value in an equality lemma like this.
Can we add `0< x` as a hypothesis or add a warning to this lemma?

affeldt-aist: The last commit does that, so I guess it is ok to merge now.

PR #1753, opened 2025-11-10
PoweR with extended real exponents

jmmarulang:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -13,6 +13,9 @@
 - in set_interval.v
   + `itv_is_open_unbounded`, `itv_is_oo`, `itv_open_ends` (Prop to bool)
 
+- in `lebesgue_Rintegrable.v`:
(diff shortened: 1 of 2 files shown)


affeldt-aist:
Maybe you shouldn't remove `ln_eq0` and change the proof of `lne_eq0`.

jmmarulang:
Unrelated to that. Im starting to think that `lne`  returning `-oo` instead of `0` for negative numbers is more cumbersome than beneficial. Specially when trying to work with `poweR` you are forced to deal with a lot more cases, since it is no longer true that `(x < 0) -> x `^ y = 1`.

hoheinzollern:
@affeldt-aist, after discussing this development with @jmmarulang, we think that this new definition of the power function is more cumbersome to work with than the one proposed in #1613, we should discuss whether it's a good idea to polish the current version or find another approach.

PR #1784, opened 2025-12-03
pretty-printing of {within _, continuous _}

### LINKED ISSUE #1781: `{within _, continuous _}` notation is not documented
https://github.com/math-comp/analysis/blob/6be7e887ad1c9da93853f7d27f26e86f9a9c7d47/theories/topology_theory/subspace_topology.v#L306

affeldt-aist:
Co-author: holgerthies

fixes #1781 

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -69,6 +69,8 @@
 - in `measurable_realfun.v`:
   + lemmas `measurable_funN`
   + lemmas `nondecreasing_measurable`, `nonincreasing_measurable`
+- in `subspace_topology.v`:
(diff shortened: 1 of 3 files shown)


CohenCyril:
The problem with this notation is that it is not compositional, there are many reasons why `f` could have domain a subspace topology, and not have been tagged with `from_subspace`. However it does preserve the notation `{within ...} if you typed it in the first place, so it is still an improvement. I will work on a more general solution based on displays at some point, meanwhile this is a strict improvement.

PR #1798, opened 2025-12-15
Generalize lemma `null_set_integral`

Yosuke-Ito-345:

I noticed that we could slightly generalize the lemma `null_set_integral`.
The generalized version is named `null_set_integral_gen`.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -75,6 +75,9 @@
 - in `subspace_topology.v`:
   + definition `from_subspace`
 
+- in `lebesgue_integrable.v`:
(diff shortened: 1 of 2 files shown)


Yosuke-Ito-345:
@affeldt-aist @t6s 
It's a small generalization, but I'd like to use them for my formalization of actuarial mathematics.
I kindly ask for your review.

affeldt-aist:
It is maybe better to replace the old lemma by its generalization rather than introduce a new lemma (I did that in the last commit).

PR #1801, opened 2025-12-20
differentiable rsubmx/lsubmx lemmas + derive_sqrt

yosakaon:

When deriving differential equations using lsubmx/rsubmx, we need proofs of their differentiability.
This PR also contains a lemma making use of the "is_derive1_sqrt" lemma to simply derive square roots.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/derive.v
+++ b/theories/derive.v
@@ -760,19 +760,13 @@ move=> dfx dgfx; apply: DiffDef; first exact: differentiable_comp.
 by rewrite diff_comp // !diff_val.
 Qed.
 
-Lemma differentiable_rsubmx m {n1 n2} (f : V -> 'M[R]_(m, n1 + n2)) v :
-  (forall x, differentiable f x) -> differentiable (rsubmx \o f) v.
-Proof.
-move=> df; apply: differentiable_comp => //.
-exact/linear_differentiable/continuous_rsubmx.
-Qed.
+Lemma differentiable_rsubmx m {n1 n2} v :
+  differentiable (rsubmx : 'M[R]_(m, n1 + n2) -> 'M[R]_(m, n2)) v.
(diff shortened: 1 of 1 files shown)


affeldt-aist (approved):
There is certainly more theory to better develop but we'll be coming back to it soon so that is good enough for me now.

theories/derive.v
```diff
@@ -760,6 +760,20 @@ move=> dfx dgfx; apply: DiffDef; first exact: differentiable_comp.
...
+move=> df; apply: differentiable_comp => //.
+exact/linear_differentiable/continuous_rsubmx.
+Qed.
+
+Lemma differentiable_lsubmx m {n1 n2} (f : V -> 'M[R]_(m, n1 + n2)) v :
+  (forall x, differentiable f x) -> differentiable (lsubmx \o f) v.
+Proof.
+move=> df1; apply: differentiable_comp => //.
+exact/linear_differentiable/continuous_lsubmx.
+Qed.
```
t6s: These names do not exactly match their statements. Suggested change:
```
Lemma differentiable_rsubmx m {n1 n2} v : differentiable (@rsubmx R m n1 n2) v.
Proof. exact/linear_differentiable/continuous_rsubmx. Qed.

Lemma differentiable_lsubmx m {n1 n2} v : differentiable (@lsubmx R m n1 n2) v.
Proof. exact/linear_differentiable/continuous_lsubmx. Qed.
```

t6s: annotating types as in `continuous_?submx` might look better

t6s (commented):
I suggest changing the statements of `differentiable_?submx` and letting users combine it with other functions explicitly by `differentiable_comp`.

PR #1802, opened 2025-12-20
EVT for rV

yosakaon:

This is a generalization of the Extreme Value Theorem already present for real-valued functions to R^n.
Should we keep both statements ?

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -60,6 +60,11 @@
 
 - in `normed_module.v`:
   + lemma `within_continuous_compN`
+- in `normed_module.v`:
+  + lemma `compact_has_sup`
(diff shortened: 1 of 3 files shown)


affeldt-aist:
> This is a generalization of the Extreme Value Theorem already present for real-valued functions to R^n. Should we keep both statements ?

So @CohenCyril observed a generalization that we have implemented, so this PR actually now generalizes the original statement and provide two variants for the real line and for `R^n`. It is a strict generalization that we know is useful to others, so to me it is a candidate for merging, let's see that the CI has to say about this.

CohenCyril (approved):

PR #1816, opened 2026-01-01
fixes #1133

### LINKED ISSUE #1133: naming and definition of `monotonous`
https://github.com/math-comp/analysis/blob/437d12c836f530fe5dfc8a26e7330b2653d175a3/classical/mathcomp_extra.v#L1491

"According to WP (https://en.wikipedia.org/wiki/Monotonic_function), this should rather be monotonic and I would expect it to be non strict (with homo instead of mono). Maybe we should also have a strict version. I would also expect it to use nondecreasing_fun and the like from realfun.v.
All in all, maybe this change should come in a separate PR so as to not delay the remaining of the current PR." @proux01

affeldt-aist:

fixes #1133

the existing (ill-named) `monotonous` corresponds to
a strict version of monotonicity (using `mono`/ `<=`)

following the conversation in issue
https://github.com/math-comp/analysis/issues/1133
this PR:

- introduces the new definition `monotonic`
  (non strict, using `homo` instead of `mono`)
- renames `monotonous` into `strict_monotonic`
  and redefines it using `homo`/`<` instead of `mono`/`<=`
- adjust lemmas in `num_normedtype.v` and `realfun.v`
- in particular, in `num_normedtype.v`:
  + weaken hypo in lemma `mono_mem_image_segment`
  + weaken hypo in lemma `mono_surj_image_segment`
  + weaken hypo in lemma `inc_surj_image_segment`
  + weaken hypo in lemma `dec_surj_image_segment`
  + weaken hypo in lemma `inc_surj_image_segmentP`
  + weaken hypo in lemma `dec_surj_image_segmentP`
  + weaken hypo in lemma `mono_surj_image_segmentP`
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -132,6 +132,10 @@
 - in `derive.v`:
   + lemmas `differentiable_rsubmx`, `differentiable_lsubmx`
 
+- in `unstable.v`:
+  + definitions `monotonic`, `strict_monotonic`
+  + lemma `strict_monotonicW`
+
 ### Changed
 
 - in `charge.v`:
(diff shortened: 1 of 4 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -285,6 +293,9 @@
 - in `set_interval.v`:
   + lemma `interval_set1` (use `set_itv1` instead)
 
+- in `unstable.v`:
+  + definition `monotonous` (used `strict_monotonic` instead)
```
proux01:
```suggestion
  + definition `monotonous` (use `strict_monotonic` instead)
```

PR #1819, opened 2026-01-08
Added differentiability of the max function

lstrsrmn:
Added a lemma for showing \max is differentiable

Added two lemmas for differentiating the max function when the order is known at the point.

Differentiability of max can be useful and is general enough it should be added to the library.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/derive.v
+++ b/theories/derive.v
@@ -2110,7 +2110,7 @@ Context {K : realType} {V W : normedModType K}.
 Implicit Types f g : V -> K^o.
 Implicit Type x : V.
 
-Fact der_max f g x v : f x != g x ->
+Fact der_max_subproof f g x v : f x != g x ->
   derivable f x v -> derivable g x v ->
   {for x, continuous f} -> {for x, continuous g} ->
   (fun h => h^-1 *: (((f \max g) \o shift x) (h *: v) - (f \max g) x))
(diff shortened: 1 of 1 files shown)


theories/derive.v
```diff
@@ -2024,40 +2041,130 @@ by rewrite [LHS]linearZ mulrC.
...
+move=> fg [fav <-]; rewrite (near_eq_derive _ fg).
 by apply: DeriveDef => //; exact: near_eq_derivable fav.
 Qed.
 
+Section Derive_max.
+Context {K : realType} {V W : normedModType K}.
+Implicit Types f g : V -> K^o.
+Implicit Type x : V.
+
+Fact der_max f g x v : f x != g x ->
```
CohenCyril: Fact cannot yet be made private. I recommend calling it `der_max_subproof` to hide it.

t6s:
Should not we state these properties in terms of ``(f `|` g)%O`` instead of  `f \max g` (similarly ``(f`&`g)%O`` instead of `f \min g`)?

Not directly a comment to this PR itself, but `\max` and `\min` look bad names (e.g., `f \max g` can be neither `f` nor `g`).

PR #1847, opened 2026-02-15
Add lemma `le0_expectation_cdf`

Yosuke-Ito-345:

Add lemma `le0_expectation_cdf`, which is the counterpart of lemma `ge0_expectation_ccdf`.

##### Checklist

N.B. I created `CHANGELOG_UNRELEASED_new.md` because the current `CHANGELOG_UNRELEASED.md` appeared to be old.

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/measurable_realfun.v
```diff
@@ -1182,6 +1182,13 @@ HB.instance Definition _ f g := max_mfun_subproof f g.
 
 Definition max_mfun f g : {mfun aT >-> _} := f \max g.
 
+Lemma min_mfun_subproof f g : @isMeasurableFun d _ aT rT (f \min g).
+Proof. by split; apply: measurable_minr. Qed.
+
+HB.instance Definition _ f g := min_mfun_subproof f g.
+
+Definition min_mfun f g : {mfun aT >-> _} := f \min g.
```
t6s, line 1127: `min_mfun` should be added to the header doc.

theories/lebesgue_integral_theory/lebesgue_integral_nonneg.v
```diff
@@ -568,6 +568,38 @@ Qed.
 
 End ge0_transfer.
 
+Section change_of_variable.
+Open Scope ereal_scope.
+Context {R : realType}.
+
+Let mu : {measure set _ -> \bar R} := @lebesgue_measure R.
+
+Lemma lebesgue_measure_oppr A (mA : measurable A) :
```
t6s: I suggest renaming this to `lebesgue_measureN`.

theories/lebesgue_integral_theory/lebesgue_integral_nonneg.v
```diff
@@ -568,6 +568,38 @@ Qed.
...
+rewrite measurable_mu_extE /pushforward/preimage/=; last exact: is_ocitv.
+have ->: [set r | (-r)%R \in `]a, b]] = `[(-b)%R, (-a)%R[%classic.
+  by apply: eq_set => r; rewrite oppr_itvoc.
+rewrite lebesgue_measure_itv/= lte_fin ltrN2.
+have [ab | ba] := (ltP a b).
+- by rewrite wlength_itv_bnd ?ltW// opprK EFinD addeC.
+- by rewrite set_itv_ge ?wlength0// bnd_simp -leNgt.
+Qed.
+
+Lemma ge0_integral_oppr (f : R -> \bar R)
```
t6s: I suggest `ge0_integral_pushforwardN`.

theories/probability_theory/random_variable.v
```diff
@@ -519,6 +636,72 @@ rewrite [X in measurable X](_ : _ = fgts @^-1` [set true]).
 by apply: eq_set => r; rewrite in_itv/= s_ge0.
 Qed.
 
+Lemma ge0_expectation_prob_ge (X : {RV P >-> R}) : (forall x, 0 <= X x)%R ->
```
t6s: What about giving a name to this integrand?
It should be equal to the "closed version" of ccdf.
```
Definition ccdf P X r := distribution P X `]r, +oo[.
Definition this_integrand P X r := distribution P X `[r, +oo[.
```

(However I myself do not come up with a good name.)

Yosuke-Ito-345: This is only an auxiliary lemma for the subsequent lemma `le0_expectation_cdf`.
If I should not take this as an independent lemma, I would incorporate this into the proof of the lemma `le0_expectation_cdf`.

t6s: Keeping this proof independent would contribute to the readability.
You can do `Local Lemma ge0_expectation_prob_ge` or `Let ge0_expecation_prob_ge`, so that
this will not be exported.

Yosuke-Ito-345: All right. I changed this `Lemma` to `Let`.
Thank you for the suggestion.

theories/probability_theory/random_variable.v
```diff
@@ -136,6 +136,107 @@ Proof. by move=> mf intf; rewrite integral_pushforward. Qed.
 
 End transfer_probability.
 
+Section pmf_definition.
+Context {d} {T : measurableType d} {R : realType}.
+Variables (P : probability T R).
```
t6s:
```suggestion
Context {d} {T : measurableType d} {R : realType} (P : probability T R).
```

CHANGELOG_UNRELEASED_new.md
```diff
@@ -0,0 +1,19 @@
...
+
+- in lebesgue_integral_nonneg.v
+  + lemmas `lebesgue_measure_oppr`, `ge0_integral_oppr`
+
+- in measurable_realfun.v
+  + lemma `min_mfun_subproof`
+  + definition `min_mfun`
+
+- in random_variable.v
+  + lemmas `lebesgue_integral_pmf`, `cdf_measurable`, `ccdf_measurable`, `ge0_expectation_prob_ge`, `le0_expectation_cdf`
```
t6s:
```suggestion
  + lemmas `lebesgue_integral_pmf`, `cdf_measurable`, `ccdf_measurable`,
    `ge0_expectation_prob_ge`, `le0_expectation_cdf`
```

t6s (approved):
I have further added a few minor suggestions, but the code now generally looks good.

t6s:
@Yosuke-Ito-345 My last comment was not meant to be a clear suggestion, but a naive report on my experiment.
I am not convinced enough because my proof attempt is not completed (you can see a broken wip at https://github.com/t6s/analysis/commit/21868b382d70dd50704b42358e45fe791e70ed32 ).

This attempt may simplify the proofs and save lines of code, and then `ccdf` should also be generalized.
The `cdf` and `ccdf` would be defined by specializing `gen_cdf`, `gen_ccdf`.

affeldt-aist:
I made a first pass, I will take a look at naming later

here are a few comments:

- as a general rule, we avoid named hypotheses that are not explicitly used
  in the statement; it is often better to help reading and to detect more easily
  for example when they become useless (that does not really apply here, though)
- use a type contraint instead of the T2 argument of pushforward
  because the name of T2 can change in the future
- ge0_integral_pushforwardN is not shorter but makes appear a
  potentially useful lemma that is maybe lacking in MathComp
- the suffix _subproof usually indicates proofs that are not meant
  to be exported to the user (max_mfun_subproof should not have been a Lemma
  but it was because it was used in simple_functions.v to circumvent
  an HB limitation that has been lifted since---apparently...)
- when setT appears in statements, it is often better to use the notation [set: _]
  becomes the type provides information
- proofs inside le0_expectation_cdf are a bit shorter if one uses
  preimage's instead of set's defined by comprehension
- it looks like a lemma akin to opp_preimage_itvbndy were called for

theories/probability_theory/random_variable.v
```diff
@@ -202,37 +202,35 @@ Qed.
...
-transitivity
-  (\int[mu]_(r in pmfP) (pmf X r)%:E + \int[mu]_(r in pmf0) (pmf X r)%:E).
-  by rewrite -ge0_integral_setU ?Upmf// ?pmf0NP => [| r _];
-    [exact: measurableC | exact: fine_ge0].
-rewrite (_ : \int[mu]_(r in pmf0) (pmf X r)%:E = 0) ?addr0;
-  last by apply: integral0_eq => r <-.
-by apply: null_set_integral; [| exact: (measurable_funS(E:=setT)) |].
+rewrite -Upmf ge0_integral_setU ?Upmf//=; last 2 first.
+  by rewrite pmf0NP; exact: measurableC.
+  by move=> x _; rewrite lee_fin pmf_ge0.
```
Yosuke-Ito-345, line 229: According to the [contribution guide](https://github.com/math-comp/math-comp/blob/master/CONTRIBUTING.md#indentation-in-proof-scripts), shouldn't we add `-` in the first two subgoals?
```
- by rewrite pmf0NP; exact: measurableC.
- by move=> x _; rewrite lee_fin pmf_ge0.
```

affeldt-aist: This is indeed what is written but we rarely apply this rule in MathComp-Analysis.
For example, when there are two technical subgoals that are generated and that these subgoals are so simple that they can be proved with one line and that morally they should be automatically discharged, we often write:
```coq
rewrite tactic1; last 2 first.
  by tactic2.
  by tactic3
rest_of_the_proof.
```
I guess that this is because the `last 2 first` and the leading `by` tactical and/or the coloring are enough indications to tell the reader "this is not fundamental, skip on a cursory reading", and also because that kind of situations appears less often in MathComp than in MathComp-Analysis (where we are quickly bothered by measurability/derivability/continuity/etc. subgoals).

Of course, it is fine to apply the "official" rule. Maybe we should propose an amendment to the contribution guide. What do you think? (@CohenCyril ?)

PS: Loosely connected topic: it is very likely that in a near future the default order of generation of subgoals will change when rewriting, and that measurability/derivability/continuity/etc. subgoals will be handled more automatically (in a bit more distant future). That could impact style.

Yosuke-Ito-345: Thank you for the detail explanation. Now I understand the policy, I do not require to add `-` in this proof.
Personally, I think this policy should be written down somewhere. I support proposing an amendment to the contribution guide.

affeldt-aist:
I think I figured out the CI problem. This was specific to the dev version of Rocq, the behavior of `simpl` has slightly changed so that measurability of `R` disappeared and caused a type mismatch.
I have also made some small generalizations so that the proof of `le0_expectation_cdf` becomes less unnatural.
I am afraid you suffered a bit because of not general enough lemmas.

theories/probability_theory/random_variable.v
```diff
@@ -512,13 +625,70 @@ apply: eq_integral => s; rewrite /D inE/= in_itv/= andbT => s_ge0.
...
-pose fgts (r : R) := (s < r)%R.
-have : measurable_fun setT fgts by exact: measurable_fun_ltr.
-rewrite [X in measurable X](_ : _ = fgts @^-1` [set true]).
+have : measurable_fun setT (<%R s) by exact: measurable_fun_ltr.
+rewrite [X in measurable X](_ : _ = (<%R s) @^-1` [set true]).
   by move=> /(_ measurableT [set true]); rewrite setTI; exact.
 by apply: eq_set => r; rewrite in_itv/= s_ge0.
 Qed.
 
+Let ge0_expectation_prob_ge (X : {RV P >-> R}) : (forall x, 0 <= X x)%R ->
```
affeldt-aist: Since the (large) inequality that was on the right-hand side has disappears in favor of `preimage` maybe `ge0_expectation_preimage` is a bit more appropriate now.

affeldt-aist:
(CI should be fixed now, with the last commit)

theories/probability_theory/random_variable.v
Yosuke-Ito-345: Yes, I agree with you on changing `ge0_expectation_prob_ge` to `ge0_expectation_preimage`.

theories/lebesgue_integral_theory/lebesgue_integral_nonneg.v
```diff
@@ -568,6 +568,33 @@ Qed.
...
+
+Lemma lebesgue_measureN (A : set R) : measurable A ->
+  pushforward mu (-%R : _ -> measurableTypeR R) A = mu A.
+Proof.
+move=> mA; apply/esym/lebesgue_measure_unique => //= _ [[a b]] _ <-.
+rewrite /pushforward opp_preimage_itvbndbnd/= !lebesgue_measure_itv/=.
+by rewrite !lte_fin ltrN2 opprK addrC.
+Qed.
+
+Lemma ge0_integral_pushforwardN (f : R -> \bar R) :
```
affeldt-aist: `pushforward` does not appear in the statement so maybe it'd better not appear in the name.
By the way, this is a variant of `ge0_integration_by_substitutionNy` in `ftc.v` (fyi: @IshiguroYoshihiro) (maybe even a strict generalization provided the constant bound is generalized).
I'd like to give a bit more thought to naming, generalization, and potentially trying to use the general version when possible to see what can be gained in other files.

affeldt-aist: No obvious generalization, I propose to go with the name `ge0_integration_by_substitution0` which is a bit long but more uniform with the similar lemmas in `ftc.v`: by choosing this name, when somebody searches for lemmas with the substring "substitution", they might find this along the others and maybe benefit from the weaker assumption.

Yosuke-Ito-345: I agree with your proposal.
By the way, I am considering formalizing the substitution formula for the Lebesgue integral in which the continuity of the integrand is not assumed. In fact, the integrand only needs to be measurable.
I think it is not easy and takes time, so I will send another pull request.

affeldt-aist: That would be great. By the way, @IshiguroYoshihiro knows best about these lemmas.

Yosuke-Ito-345:
Thank you for the detailed review.

PR #1674, opened 2025-07-04
Generalize integration_by_parts

IshiguroYoshihiro:
##### Motivation for this change
Add a part of generalized version of integration by parts for unbound intervals  `` `[a, +oo[ ``.
This PR is over #1656 and #1662.
I don't sure that I should add 4 more lemmas and complete all variation of integration by parts for unbound intervals.
I think that 8 similar lemmas would be a mess. So, I'd like to somehow make them up into one or two lemmas, but I don't have an idea.

##### Checklist

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`

- [ ] added corresponding documentation in the headers

Reference: [How to document](https://github.com/math-comp/math-comp/wiki/How-to-document)

##### Merge policy

As a rule of thumb:
- PRs with several commits that make sense individually and that
  all compile are preferentially merged into master.
- PRs with disorganized commits are very likely to be squash-rebased.

##### Reminder to reviewers

- Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs)
- Put a milestone if possible
- Check labels

PR #1848, opened 2026-02-16
add `itv_closureE`, `itv_interiorE`, and helper lemmas

t6s:

Reviewing PR #1674 , it was noticed that there is not much theory
connecting closure/interior operators and set-intervals.
This PR adds lemmas to amend that, especially `itv_interiorE` and `itv_closureE`
stating that the interior/closure of an interval is an open/closed interval respectively,
for intervals in `realFieldType`.

This should resolve the following comment:
https://github.com/math-comp/analysis/pull/1674/changes/BASE..34697b2642259157cd3678e64092864dcd0ca4ef#r2685444293

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/set_interval.v
+++ b/classical/set_interval.v
@@ -182,37 +182,65 @@ Definition set_itvE := (set_itv1, set_itvoo0, set_itvoc0, set_itvco0, set_itvoo,
 Lemma set_itvxx a : [set` Interval a a] = set0.
 Proof. by move: a => [[|] a |[|]]; rewrite !set_itvE. Qed.
 
-Lemma setUitv1 a x : (a <= BLeft x)%O ->
-  [set` Interval a (BLeft x)] `|` [set x] = [set` Interval a (BRight x)].
+Let setUitv1_gen a x b : (a <= BLeft x)%O ->
+  [set` Interval a (BSide b x)] `|` [set x] = [set` Interval a (BRight x)].
 Proof.
-move=> ax; apply/predeqP => z /=; rewrite itv_splitU1// [in X in _ <-> X]inE.
+move=> ax; case: b; last first.
+  by apply: setUidl => ? /= ->; rewrite itv_boundlr ax lexx.
+apply/predeqP => z /=; rewrite itv_splitU1// [in X in _ <-> X]inE.
 by rewrite (rwP eqP) (rwP orP) orbC.
 Qed.
+
+Let setU1itv_gen a x b : (BRight x <= a)%O ->
+  x |` [set` Interval (BSide b x) a] = [set` Interval (BLeft x) a].
+Proof.
+move=> ax; case: b.
+  by apply: setUidr => ? /= ->; rewrite itv_boundlr ax lexx.
+apply/predeqP => z /=; rewrite itv_split1U// [in X in _ <-> X]inE.
+by rewrite (rwP eqP) (rwP orP) orbC.
+Qed.
+
+Lemma setUitv1 a x : (a <= BLeft x)%O ->
+  [set` Interval a (BLeft x)] `|` [set x] = [set` Interval a (BRight x)].
+Proof. exact: setUitv1_gen. Qed.
 
 Lemma setU1itv a x : (BRight x <= a)%O ->
   x |` [set` Interval (BRight x) a] = [set` Interval (BLeft x) a].
+Proof. exact: setU1itv_gen. Qed.
+
+Lemma setUitv2 x y b1 b2 :
+  (x <= y)%O ->
+  [set` Interval (BSide b1 x) (BSide b2 y)] `|` [set x; y] = `[x, y]%classic.
 Proof.
-move=> ax; apply/predeqP => z /=; rewrite itv_split1U// [in X in _ <-> X]inE.
-by rewrite (rwP eqP) (rwP orP) orbC.
+rewrite le_eqVlt => /orP [/eqP->|xy].
+  by case: b1; case: b2; rewrite !set_itvE !setUid // set0U.
+rewrite setUCA setUitv1_gen; last by case: b1; rewrite bnd_simp// ltW.
+by rewrite setU1itv_gen// bnd_simp ltW.
 Qed.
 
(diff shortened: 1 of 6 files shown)


classical/unstable.v
```diff
@@ -553,3 +553,207 @@ End ProperNotations.
...
+
+Module EndlessDenseOrderTheory.
+
+Section theory.
+Local Open Scope order_scope.
+
+Definition is_endless_porderType {d} (T : porderType d) :=
+  forall x : T, (exists y, y < x) /\ (exists y, x < y).
+
+Definition is_dense_porderType {d} (T : porderType d) :=
```
affeldt-aist, line 578: This might need to be related to the definition `dense` in `topology_structure.v`.

classical/set_interval.v
```diff
@@ -182,38 +179,66 @@ Definition set_itvE := (set_itv1, set_itvoo0, set_itvoc0, set_itvco0, set_itvoo,
 Lemma set_itvxx a : [set` Interval a a] = set0.
 Proof. by move: a => [[|] a |[|]]; rewrite !set_itvE. Qed.
 
-Lemma setUitv1 a x : (a <= BLeft x)%O ->
-  [set` Interval a (BLeft x)] `|` [set x] = [set` Interval a (BRight x)].
+Let setUitv1_gen a x b : (a <= BLeft x)%O ->
```
affeldt-aist, line 182: This indeed looks like a generalization, so maybe it should not be a `Let` but a replacement for `setUitv1`. Doing this generalization in a small, preliminary PR (maybe containing also `setDitv1r` and `setDitv1l`) could simplify the review of this one.

classical/unstable.v
```diff
@@ -553,3 +553,207 @@ End ProperNotations.
...
+Qed.
+
+End interval.
+
+Module EndlessDenseOrderTheory.
+
+Section theory.
+Local Open Scope order_scope.
+
+Definition is_endless_porderType {d} (T : porderType d) :=
```
affeldt-aist, line 575: Definitions need to be documented, but I guess you were planning to do it later.

affeldt-aist: Did you consider using an `HB.mixin` for this property? Since, for example, `numDomainType` is an instance, that might simplify the presentation of later lemmas.

classical/unstable.v
t6s: Denseness of a subset in a topological space is a different notion, and there does not seem to be a unifying theory.
Cyril suggested that using a fully qualified name like `Order.dense` would suffice for differentiating these two
(creating a PR to mathcomp/order.v).

classical/set_interval.v
t6s: One bad thing about this generalization is that all use cases of `setUitv1` and `setU1itv` in analysis master are
backwards rewriting, and adding another parameter `b` breaks them.
I will create a PR to clarify this issue, and possibly try to simplify the relevant proofs there.

classical/unstable.v
t6s: Turning this part into a structure using HB is actually good.
I somehow thought that would not work, but let us see by trying it. I will update the code in that way.

classical/set_interval.v
t6s: https://github.com/math-comp/analysis/pull/1864

affeldt-aist: I created an issue (https://github.com/math-comp/analysis/issues/1914) so that we can improve the usability of `setUitv1` and `setU1itv` independently of this PR.

classical/set_interval.v
```diff
@@ -182,38 +179,66 @@ Definition set_itvE := (set_itv1, set_itvoo0, set_itvoc0, set_itvco0, set_itvoo,
...
+
+Lemma setUitv1 a x : (a <= BLeft x)%O ->
+  [set` Interval a (BLeft x)] `|` [set x] = [set` Interval a (BRight x)].
+Proof. exact: setUitv1_gen. Qed.
+
 Lemma setU1itv a x : (BRight x <= a)%O ->
   x |` [set` Interval (BRight x) a] = [set` Interval (BLeft x) a].
+Proof. exact: setU1itv_gen. Qed.
+
+Lemma setUitv2 x y b1 b2 :
```
affeldt-aist, line 208: I guess this lemma will have the same usability problem as `setUitv1` and `setU1itv` (see issue #1914 ) and that specialization with easy-to-understand names will be called for. On the other hand, it is not used in this PR, so it could be separated in a different PR to simplify the review of this PR (#1848).

affeldt-aist:
There a few lemmas in `topology_structure.v` (`closureEbigcap_itvcy`, `interiorEbigcup_itvNyc`, `closureEbigcap_itvcc`, `interiorEbigcup_itvcc`) that look like they could be PRed independently.
By the way, I was surprised to see that in these lemmas the variable `A` is used both as a set and
as an interval bound. Isn't that surprising and deserving a comment?

affeldt-aist (changes requested):
This PR is definitely useful (as exemplified by PR #1910 which is a special case as you rightfully observed, sorry about the duplicate, my bad).

I made a couple of requests to split out a few things that look like they could be PRed independently. The goal is twofold.

On the one hand, additions to `classical_sets.v`, `topology_structure.v` can be merged quickly; the addition of `setUitv2` might have the same issue as we had with `setUitv1`/`setU1itv` which can be solved by a naming discussion that could take a bit more time so an individual PR would help.

On the other hand, the main addition about endlessness and order-density still needs to be experimented using HB.mixin's (as we already discussed). I am confident it will work out great but in any case this is a substantial addition to `unstable.v` that will need to be backported to MathComp at some point. I think that reducing this PR to endlessness and order-density will have the advantage of making it clearer when doing the HB.mixin experiment and the backport.

What do you think?

classical/set_interval.v
affeldt-aist: NB: issue #1914 has been solved and naming convention clarified for easy-to-understand/remember naming

PR #1853, opened 2026-02-20
Drop support for Coq 8.20

proux01:
Maybe something we want to do after the next release (this will also enable improving the deprecations by using `use=`).

Cc @affeldt-aist

### DIFF
--- a/.github/workflows/nix-action-9.0-2.4.0.yml
+++ b/.github/workflows/nix-action-9.0-2.4.0.yml
@@ -1,6 +1,7 @@
 jobs:
   coq:
-    needs: []
+    needs:
+    - rocq-core
     runs-on: ubuntu-latest
     steps:
     - name: Determine which commit to initially checkout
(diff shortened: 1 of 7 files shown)


affeldt-aist:
What about merging it now for the release of 1.16.0?
I am asking because
- there was a Coq 8.20 related CI problem in another PR https://github.com/math-comp/analysis/pull/1865#issuecomment-3976993897 , I don't get the error and I am not able to reproduce it by recreating an opam environment with Coq 8.20 and MathComp 2.4.0
- Rocq 9.0 and Rocq.9.1 seem to be both supported, so we abide by the "at least two versions supported" rule

affeldt-aist:
Thanks for the information. Then unless somebody figures out a solution I think we'll merge before the release.

proux01:
The importnat part is "pull_request_target" (i.e., master), there is no "pull_request" equivalent. It's a way to make explicit the CI config has changed and its indeed expected to be red.

PR #1855, opened 2026-02-24
fix #1843

### LINKED ISSUE #1843: naming of `closed_comp`
https://github.com/math-comp/analysis/blob/b1de71b13ff9846c3e5267d5a1359be96d4c39c1/theories/topology_theory/topology_structure.v#L787-L788

shouldn't this lemma be called, for example, `continuous_closed_preimage`, because there is no explicit `comp` in the statement...

@zstone1 

fyi: @holgerthies

affeldt-aist:

fix #1843

@holgerthies -> we discussed renaming on the `closed_comp` lemma recently, eventually I didn't use the name we though of at the time, what do you think? (in MathComp, when the predicate is unary, the order is reversed on purpose @CohenCyril )

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -195,6 +195,10 @@
 - in `lebesgue_integral_nonneg.v`:
   + `integral_setD1_EFin` -> `integral_setD1`
 
+- in `topology_structure.v`:
+  + `closed_comp` -> `preimage_closed`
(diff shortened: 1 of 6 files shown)


holgerthies:
I think the lemma names are more accurate now.

PR #1864, opened 2026-02-27
generalize setUitv1, setU1itv, setDitv1{l,r}, and add similar lemmas

t6s:

This PR does
- generalize `setUitv1, setU1itv, setDitv1{l,r}`,
- add similar lemmas such as `setUitv2`, `setDitv2`, `setDitvoo`, etc.

https://github.com/math-comp/analysis/pull/1848#discussion_r2857392258

I am not confident about the names of two added lemmas: `setDcitvy` and `setDcitvNy`,
and want reviewers to suggest better ones if any.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -4,8 +4,8 @@
 
 ### Added
 - in set_interval.v
-  + lemmas `setUitv2`, `setDitv2`, `setDitvoo`, `setDitvoy`, `setDitvNyo`,
-    `setDccitv`, `setDcitvy`, `setDcitvNy`
+  + lemmas `setUitv_set2`, `setDitv_set2`, `setDitvoo`, `setDitvoy`, `setDitvNyo`,
+    `setDccitv`, `setD_cbnd_bndy`, `setD_bndc_Nybnd`
 
 - in `topology_structure.v`
   + lemma `interiorS`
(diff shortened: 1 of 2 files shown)


affeldt-aist:
One CI error is the following:

```text
       > File "./reals/real_interval.v", line 61, characters 28-38:
       > Error: The reference ltrW_lteif was not found in the current environment.
       > Did you mean lt_leif?
```

Looking at MathComp's changelog, it looks like a MathComp 2.5.0 lemma.

t6s:
Thank you for reviewing!

> Since you have a doubt about the naming of `setUitv2` and `setDitv2`, what about `setUitv_set2` and `setDitv_set2`? It might make it clearer that the lemmas are about dealing with the (two) endpoints.

Unlike `setDcitvy` and `setDcitvNy`, I am fine with `setUitv2` and `setDitv2`.  Have you found these names less clear? Then I do not hesitate to rename them to `*_set2`.
 
> I observe that, when used, for example `setU1itv`, is mostly used to rewrite from right to left and requires a boolean flag whose setting always requires a bit of thought, which makes me wonder whether there isn't some improvement possible in terms of usability, but I do not have any idea (but you already observed that issue in he original PR).

One solution would be to introduce a "disjoint union" operator in classical_sets,
and use it to state the original version as a different lemmas.

> Some lemmas that you add, specifically:
> 
> * `setDitvoy`
> * `setDitvNyo`
> * `setDccitv`
> * `setDcitvy`
> * `setDcitvNy`
>   are not used in the PR. They are variations of two others,
>   so that's almost enough a justification, but are they used
>   in the original PR 1848?

I think some are used in PR 1848, but probably not all.
The others are indeed written down for symmetry and perfection.

> Looking at MathComp's changelog, it looks like a MathComp 2.5.0 lemma.

I will fix this soon.

> setDcitvy -> setD_cbnd_bndy
> setDcitvNy -> setD_bndc_Nybnd

Thanks for suggestions. I will take these names.

PR #1877, opened 2026-03-04
Tvs 20260304

affeldt-aist:

fyi: @mkerjean 

This PR moves the file `tvs.v` into `normedtype_theory` in such a way that it only depends `num_normedtype.v`
and `pseudometric_normed_Zmodule` (instead of the whole directory).

In the process:
- some structures (that are actually joins) needed to be moved out of `tvs.v` and
  placed earlier in the file hierarchy (it looks like a good change)
- some local lemmas were turned into `Let`'s (this was an oversight)
- two longish proofs could be shortened (`standard_{add,scale}_continuous`) 
- comments that were suggesting proof sharing have been updated

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -206,7 +206,6 @@
 
 - moved from `theories` to `theories/topology_theory`:
   + file `function_spaces.v`
-- moved from `theories` to `theories/topology_theory`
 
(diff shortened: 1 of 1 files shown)


mkerjean (approved):
Simplifying code and allowing for further simpler proofs in tvs.v

PR #1887, opened 2026-03-06
convex_function generalized

mkerjean:

Generalization the definition of convex functions in `convex.v`, from a function `f: R -> R^o` for `R: realType` to a function `f : E -> R^o`, for `R: numFieldType` and `E : convex_lmodType`. 

This will allow the use of `convex_function` in the statement of Hahn-Banach theorem, and more generally, the use of semi-norms as convex functions on instances of `convextvsType`.

fixes issue #1888 

Co-authored-by: Reynald Affeldt <reynald.affeldt@aist.go.jp>

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/normedtype_theory/tvs.v
+++ b/theories/normedtype_theory/tvs.v
@@ -310,8 +310,8 @@ HB.end.
 
 HB.mixin Record Uniform_isTvs (R : numDomainType) E
     & Uniform E & GRing.Lmodule R E := {
-  locally_convex : exists2 B : set (set E),
-    (forall b, b \in B -> convex b) & basis B
+  locally_convex : exists2 B : set_system E,
+    (forall b, b \in B -> convex_set b) & basis B
 }.
 
 #[short(type="tvsType")]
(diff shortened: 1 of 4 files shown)


theories/normedtype_theory/tvs.v
```diff
@@ -511,23 +509,28 @@ rewrite (@le_lt_trans _ _ (`|k - l| * M)) ?ler_wpM2l -?ltr_pdivlMr//.
 by near: l; apply: cvgr_dist_lt; rewrite // divr_gt0.
 Unshelve. all: by end_near. Qed.
 
+Local Open Scope convex_scope.
+
+Let standard_ball_convex (x : R^o) (r : R) : convex (ball x r).
```
t6s: Isn't this an instance of `ball_convex` above?

affeldt-aist: Yes but `ball_convex` is proved inside an HB builder where the carrier is an lmodType and has a norm,
but outside the builder we do not have such a structure, making it impossible to factorize. :-(

affeldt-aist:
> > > What about `convex_set`?
> > 
> > 
> > That looks specific enough. Is that conformant with other names for predicates on sets?
> 
> In infotheo, we called that `is_convex_set`.

That is maybe closer to, e.g.,:
```
(*               is_subset1 A <-> A contains only 1 element                   *)
```

t6s:
Currently there does not seem to be any rule about prefixing `is_` before predicates.
In cardinality.v:
```
Definition finite_set {T} (A : set T) := exists n, A #= `I_n.
Notation infinite_set A := (~ finite_set A).
Notation cofinite_set A := (finite_set (~` A)).
```

affeldt-aist:
For the time being, I tried with the shortest one: https://github.com/math-comp/analysis/pull/1887/changes/427bae301d9e1d70772fc57580cdc78981b861a1

PR #1889, opened 2026-03-06
Hahn banach 2026

mkerjean:
##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


affeldt-aist:
I made a pass on the file: moving a few lemmas to `mathcomp_extra.v` and `unstable.v`, linting the code, creating a `functional_analysis` sub-directory, starting the changelog (NB: it contains a fix for the changelog in master which is buggy, my bad).

Since `topU` is defined at the top-level and therefore exposed, maybe a name like `sub_initial_topology` is better (btw, it should be documented).

TODO: move the new structures out of this file.

affeldt-aist:
Compilation is fine with MathComp 2.5.0 but it looks like with a problem with the duplication of the `SubNormedZmodule` in MathComp and MathComp-Analysis:
```
Error:
You must declare the hierarchy bottom-up or add a missing join. 
There are two ways out: 
- declare structure SubNormedZmodule before structure Num.SubNormedZmodule if Num.SubNormedZmodule inherits from it; 
- declare an additional structure that inherits from both Num.SemiNormedZmodule and SubType and from which SubNormedZmodule and/or Num.SubNormedZmodule inherit.
```
It this is really the case, is there a trick to make it compile with both versions? @CohenCyril @gares
(It could even be a feature of HB? Again, not sure whether my diagnosis is correct...)

affeldt-aist:
> Compilation is fine with MathComp 2.5.0 but it looks like with a problem with the duplication of the > is there a trick to make it compile with both versions? @CohenCyril @gares (It could even be a feature of HB? Again, not sure whether my diagnosis is correct...)

Like a "the version 2.X.Y of MathComp only" option?

affeldt-aist:
we tried this trick to cheat the CI: https://github.com/math-comp/analysis/pull/1889/changes/4d044c86f6626f647605ce2b0c61c89223c4bb5f

but nix fails for other reasons so this is not yet conclusive

theories/functional_analysis/hahn_banach_theorem.v
```diff
@@ -0,0 +1,660 @@
...
+End hahn_banach.
+
+(* TODO: to define on tvs, characterize the topology of a tvs via its pseudonorms,
+and the continuity of linear continuous functions via the pseudonorms. *)
+
+Section hahn_banach_normed.
+Variable (R : realType) (V : normedModType R) (F : pred V)
+  (F' : subNormedModType F) (f : {linear_continuous F' -> R}).
+
+(*To use the thm on a F': subLmodType F, use @SubLmodule_isSubNormedmodule.Build.
```
affeldt-aist: @mkerjean I think that we have not yet addressed this comment. Should the desired factory be part of this PR or should it be an issue? (If this is going to be an issue, maybe a good description of the desired factory is necessary.)

theories/functional_analysis/hahn_banach_theorem.v
```diff
@@ -0,0 +1,660 @@
...
+pose lg := GRing.isLinear.Build _ _ _ _ g scalg.
+pose g' : {linear V -> R | *%R} := HB.pack g lg.
+exists g'.
+  by case: z gP => [c [_ _ bp _]] /= gP => x; apply: bp; exact/gP.
+by move=> z'; apply/gP; case: z {gP} => [c [_ _ _ pf]] /=; exact: pf.
+Qed.
+
+End hahn_banach.
+
+(* TODO: to define on tvs, characterize the topology of a tvs via its pseudonorms,
```
affeldt-aist: @mkerjean Should this be addressed within this PR or turned into an issue? In case of the latter, a more precise description of the desired feature is needed.

theories/functional_analysis/hahn_banach_theorem.v
```diff
@@ -0,0 +1,660 @@
...
+HB.instance Definition _ :=  isSubNbhs.Build _ _ U continuous_valE.
+
+Check U : subConvexTvsType S.
+
+Check U : subNormedModType S.
+
+HB.instance Definition _ := SubLmodule_isSubNormedmodule.Build _ _ _ U.
+HB.end.
+
+(* TODO : use a lightweight factory to make every subLmodType a subnormedmodtype *)
```
affeldt-aist, line 35: @mkerjean Can this comment be removed? It looks like we have addressed the TODO by creating the factory `SubLmodule_isSubNormedmodule` (just above in this file).

affeldt-aist:
I completed a pass of cleaning. As far as I am concerned, the scripts are good enough.
We still have to get HB structures (`SubNormedModule`, etc.) to the right files.
Right now, we are just trying to get `SubNormedZmodule` out of it, by putting into
`mathcomp_extra.v`, but we have to trick the CI to make it happy.
As soon as the CI is green (it is still red but for orthogonal reasons), this could be done quickly, and the PR merged imo.

theories/functional_analysis/hahn_banach_theorem.v
mkerjean: This comment can be deleted. The factory @SubLmodule_isSubNormedmodule.Build should ideally be instantiated on every SubLmodule by a lightweight factory. This comment can be deleted.

theories/functional_analysis/hahn_banach_theorem.v
mkerjean: `SubLmodule_isSubNormedmodule`  can indeed turn a given subLmodule into a subnormedmottype, but we struggled to define a lightweight factory doing so, because of a dependency on `K`. I would leave that comment - or create an issue.

theories/functional_analysis/hahn_banach_theorem.v
mkerjean: This is addressed by PR#1963. The comment can be deleted.

mkerjean:
> I completed a pass of cleaning. As far as I am concerned, the scripts are good enough. We still have to get HB structures (`SubNormedModule`, etc.) to the right files. Right now, we are just trying to get `SubNormedZmodule` out of it, by putting it into `mathcomp_extra.v`, but we have to trick the CI to make it happy. As soon as the CI is green (it is still red but for orthogonal reasons), this could be done quickly, and the PR merged imo.

The solution that uses isTmp in the SubNormedModule mixin apparently isn't enough to make the CI green. Should it still be left in the definition of substructures?

affeldt-aist:
> > I completed a pass of cleaning. As far as I am concerned, the scripts are good enough. We still have to get HB structures (`SubNormedModule`, etc.) to the right files. Right now, we are just trying to get `SubNormedZmodule` out of it, by putting it into `mathcomp_extra.v`, but we have to trick the CI to make it happy. As soon as the CI is green (it is still red but for orthogonal reasons), this could be done quickly, and the PR merged imo.
> 
> The solution that uses isTmp in the SubNormedModule mixin apparently isn't enough to make the CI green. Should it still be left in the definition of substructures?

I am not sure. The CI error look unrelated, they seem to be related with some build process within nix.

affeldt-aist:
Pending problem: find a trick so that the same code compiles with 2.5.0 and 2.6.0 (right now, that's not the case).

I put the changelog in a separate file (because if we can't solve the above problem, this might stay as a PR for some time, so let's avoid painful rebasings).

The documentation is complete. In particular, the new structures have been dispatched to the right files, leaving only Hahn-Banach stuff in the eponymous file.

I undrafted.

affeldt-aist:
We considered various strategies to accommodate MC 2.5.0 and master but ultimately concluded that MCA should drop MC 2.5.0 to be able to merge Hahn-Banach.

This PR therefore update the CI accordingly and could be merged as soon as the CI is green
(which is actually not immediate...).

In the meantime, we can take a last look at the diff to make sure everything is in place (as far as the Rocq scripts, documentation, and changelog are concerned).

CohenCyril (approved):
🎉

PR #1937, opened 2026-04-16
tentative removal of Rint

affeldt-aist:

@agontard : what do you think about this change?
We do not use it a lot in MathComp-Analysis per se, as a user, we'd like your opinion.

TODO: introduce "type of ring for which the property of being an integer is decidable" and refactor, even though we do not have a use for it now, this could be a useful addition to MathComp

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -64,7 +64,7 @@
 
 - in `reals.v` (use `Num.int` instead of `Rint` which is deprecated)
   + definition `Rtoint`
-  + lemmas `RtointK`, `inj_Rtoint`, `RtointN`
+  + lemmas `RtointK`, `inj_Rtoint`, `RtointN`, `isint_Rfloor`, `isint_Rceil`
 
 ### Renamed
 
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -26,6 +26,8 @@
 
 - in `normed_module.v`:
   + lemmas `closure_itvoo`
+- in `unstable.v`:
+  + lemma `ex_eqP`
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG_UNRELEASED.md
```diff
@@ -60,6 +62,10 @@
 - moved from `topology_structure.v` to `filter.v`:
   + lemma `continuous_comp` (and generalized)
 
+- in `reals.v` (use `Num.int` instead of `Rint` which is deprecated)
+  + definition `Rtoint`
+  + lemmas `RtointK`, `inj_Rtoint`, `RtointN`
```
agontard: \+ \`isint_Rfloor\`, \`isint_Rceil\`

affeldt-aist: Indeed! Thanks.

CohenCyril:
Hi, `Rint` was using classical logic to exhibit integers in any `realFieldType`, while `Num.int` requires Archimedeanity, therefore they are incomparable, and with this PR you eliminate expressivity -- albeit apparently not used according to the CI. The proper way to refactor this would be to introduce the "type of ring for which the property of being an integer is decidable".

affeldt-aist:
Note that the "generalization" to `realFieldType` was carried out some time ago
https://github.com/math-comp/analysis/commit/ef2487640c6e5d4cdbc95e939caf5c56b5b87250#diff-db2a60c8905528789fdebb5e121d48be6e89590ae44cfcdd26e1b037660707e0

It was not there originally.

affeldt-aist:
So, in other words, I do not think that it was meant to correspond the "type of ring for which the property of being an integer is decidable". Shall we however PR the latter to MathComp?

affeldt-aist:
Postponed to the next release.

PR #1940, opened 2026-04-18
fixes #1939

### LINKED ISSUE #1939: Wrongly deprecated?
https://github.com/math-comp/analysis/blob/3a27d02809d7fb711b7eacb44964890a02284e23/theories/measure_theory/measurable_structure.v#L1651-L1653

This is actually useful but maybe ill-named, `prod_measurable_rectangle` is maybe more sensible.
@shinya-katsumata

affeldt-aist:

fixes #1939 

@shinya-katsumata

We deprecated `measurable_prod_measurableType` because
this was the same lemma as `g_sigma_algebra_rectangle`
modulo unfolding.
However, when this lemma is used, the identifier one wants to unfold
is hidden behind a notation and in addition, unfolding it reveals
implementation details.
Its deprecation was maybe not a good idea after all, this PR removes
the deprecation and change the name to `prod_measurable_rectangle`
which is easier to remember and search for.

Can you have a quick look and approve if possible?
(don't mind the `Let` -> `#[local] Lemma` changes, this is orthogonal)

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -69,6 +69,9 @@
   + mixin ` PseudoMetricNormedZmod_Tvs_isNormedModule` ->
     ` PseudoMetricNormedZmod_ConvexTvs_isNormedModule`
 
+- in `measurable_structure.v`:
+  + `measurable_prod_measurableType` -> `prod_measurable_rectangle`
+
 ### Generalized
 
 - in `measurable_structure.v`:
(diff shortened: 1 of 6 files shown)


shinya-katsumata:
Hi, I quickly read the change, and I agree with bringing back "measurable_prod_measurableType" with a new name.

affeldt-aist:
Thanks for having taken a look!

PR #1949, opened 2026-04-24
measurable types are not pointed by default any more

### LINKED ISSUE #796: isMeasurable requires pointed type
Hi, I am Shin-ya Katsumata at NII, invited to this project by Reynald.

I have a question: it appears that the measurable space whose carrier is the empty set (type) seems excluded from the definition, because isMeasurable requires a pointed type. I wonder why.

The standard definition of measurable space includes the empty one. Also, categorically, excluding the empty measurable space drastically changes the property of the category of measurable spaces.

affeldt-aist:

fixes #796 

fyi: @shinya-katsumata @CohenCyril 

We would like this change to be reflected in master as soon as possible so as to be able to work
with the following definition:
```coq
Structure distr := Distr {
  mu :> T -> R ;
  _  :  forall x, 0 <= mu x ;
  _  :  summable [set: T] (EFin \o mu) ;
  _  :  \int[@counting (discrete_measurable_space T) R]_x (mu x) <= 1
}.
```
*without requiring* `T` to be a `pointedType`, only a `choiceType`,
to preserve the generality of `distr` from the `experimental_reals` package
@strub @lyonel2017 (just substituting `summable` and `psum` from their
MathComp-Analysis equivalent to use the library more easily).

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/numfun.v
```diff
@@ -1190,16 +1196,23 @@ Definition scale_fimfun k f : {fimfun aT >-> rT} := k \o* f.
 End ring.
 Arguments indic_fimfun {aT rT} _.
 
+(* NB: this notation will be taken by MathComp 2.6.0,
+   will become [SubSemiRing_isSubComSemiRing of U by <: ] but might require more  *)
+Notation "[ 'SubPzRing_isSubComPzRing' 'of' U 'by' <: ]" :=
+  (GRing.SubPzRing_isSubComPzRing.Build _ _ U)
+  (format "[ 'SubPzRing_isSubComPzRing'  'of'  U  'by'  <: ]")
+  : form_scope.
+
```
pi8027, line 1205: Did you try to remove this notation? According to `CHANGELOG.md`, it has been available since MC 2.4 (and deprecated in master but not removed).

t6s: I ran a quick search through the source tree of MC 2.5 but could not find the corresponding notation:
```
$ find . -exec grep -HnF "SubPzRing_isSubComPzRing" {} ';'
./algebra/ssralg.v:5797:HB.factory Record SubPzRing_isSubComPzRing (R : comPzRingType) S U
./algebra/ssralg.v:5800:HB.builders Context R S U of SubPzRing_isSubComPzRing R S U.
./algebra/ssralg.v:5839:HB.instance Definition _ := SubPzRing_isSubComPzRing.Build R S U.
./algebra/ssralg.v:6164:HB.instance Definition _ := SubPzRing_isSubComPzRing.Build R S U.
./CHANGELOG.md:986:  + factories `SubPzRing_isSubComPzRing`, `SubChoice_isSubPzSemiRing`,
```

Was not it like that only factories were added at MC 2.4 but not notations?

pi8027: Ok, fair enough... What about using tactic-in-term in the body of this notation to write a disjunction of terms?

pi8027: Something like this (not tested):
```suggestion
(* NB: this notation will be taken by MathComp 2.6.0,
   will become [SubSemiRing_isSubComSemiRing of U by <: ] but might require more  *)
Notation "[ 'SubPzRing_isSubComPzRing' 'of' U 'by' <: ]" :=
  (ltac:(exact (GRing.SubSemiRing_isSubComSemiRing.Build _ _ U) ||
         exact (GRing.SubPzRing_isSubComPzRing.Build _ _ U)))
  (format "[ 'SubPzRing_isSubComPzRing'  'of'  U  'by'  <: ]")
  : form_scope.

```

pi8027: ping @affeldt-aist

affeldt-aist: This causes a not-yet-implemented error.

pi8027: Indeed, we pass it to `HB.instance`, and we cannot quote tactic-in-term to Elpi. So, math-comp/math-comp#1587 now seems like a reasonable workaround to me.

affeldt-aist: Thanks for the help anyway!

affeldt-aist:
The patch we merged to MathComp, i.e., the addition of the following
between `Module Import Ring.` and `End Ring.` in `ssralg.v`:
```coq
HB.factory Record SubPzRing_isSubComPzRing (R : comPzRingType) S U
    & SubPzRing R S U := {}.

HB.builders Context R S U & SubPzRing_isSubComPzRing R S U.
HB.instance Definition _ := SubSemiRing_isSubComSemiRing.Build R S U.
HB.end.
```
does not seem to be complete. The error is now:
```
Error:
HB: GRing.SubPzRing_isSubComPzRing.axioms_ is not a factory or its library (mathcomp.algebra.algebraic_hierarchy.ssralg.GRing) was not correctly imported
```

I tested locally and this is indeed the error one gets
when the addition is inside `Module Ring.`/`End Ring.`
(without the `Import`).

Note that another way to make it work is to add
the factory between `Module GRing.` and `End GRing. HB.export GRing.`
or `Module GRing.` and `End GRing. Export GRing.`

I'll do more local checks unless somebody can spot the right fix.

affeldt-aist:
This PR is a strict generalization, solving an issue, and MathComp master has been patch so that the CI is green (and more fixes/improvements are on their way on the MathComp side).

I will merge soon unless somebody spots a problem.

PR #1961, opened 2026-05-03
bigcup and bigcap are not necessarily countable

mkerjean:
Fixes the `Example of notations` of `classical_set.v`, where `bigcap` and `bigcup` were explained as countable intersection and union.


affeldt-aist (approved):
Thanks for spotting that!

PR #1979, opened 2026-05-27
reduce dependencies in measurable_structure.v

affeldt-aist:

fyi @hoheinzollern

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/classical_sets.v
+++ b/classical/classical_sets.v
@@ -153,6 +153,8 @@ From mathcomp Require Import mathcomp_extra boolp wochoice.
 (*                                if there is one, to f0 x otherwise          *)
 (*                    F `#` G <-> intersections beween elements of F an G are *)
 (*                                all non empty                               *)
+(*                     seqDU F := sequence F_0, F_1\F_0, F_2\(F_0 U F_1),...  *)
+(*                      seqD F == the sequence F_0, F_1 \ F_0, F_2 \ F_1,...  *)
 (* ```                                                                        *)
 (*                                                                            *)
 (* ## Pointed types                                                           *)
@@ -187,8 +189,6 @@ From mathcomp Require Import mathcomp_extra boolp wochoice.
 (*                                D : set I, form a partition of A            *)
 (*          pblock_index D F x == index i such that i \in D and x \in F i     *)
 (*                pblock D F x := F (pblock_index D F x)                      *)
-(*                     seqDU F := sequence F_0, F_1\F_0, F_2\(F_0 U F_1),...  *)
-(*                      seqD F == the sequence F_0, F_1 \ F_0, F_2 \ F_1,...  *)
 (*                                                                            *)
 (*   maximal_disjoint_subcollection F A B == A is a maximal (for inclusion)   *)
 (*                                   disjoint subcollection of the collection *)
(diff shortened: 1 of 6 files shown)


affeldt-aist:
This is really just reducing dependencies in `measurable_structure.v`, so a clear improvement.

classical/classical_sets.v
```diff
@@ -2310,78 +2323,24 @@ Proof. by apply: setC_inj; rewrite setC_bigcap setC_bigsetI bigcup_seq. Qed.
...
+Implicit Types (A : set T) (F : (set T)^nat).
 
 Lemma bigcup_mkord n F : \bigcup_(i < n) F i = \big[setU/set0]_(i < n) F i.
 Proof.
 rewrite -(big_mkord xpredT F) -bigcup_seq.
 by apply: eq_bigcupl; split=> i; rewrite /= mem_index_iota leq0n.
 Qed.
 
+Lemma bigcup_bigsetU_bigcup F :
+  \bigcup_k \big[setU/set0]_(i < k.+1) F i = \bigcup_k F k.
```
t6s, line 2338: not about this PR itself, but why do lemmas around here have `.+1` in the indexing for finite parts?

t6s (approved):
looks good

t6s:
One action in the CI is failing, complaining xfinmap is not found. What is xfinmap?

classical/classical_sets.v
```diff
@@ -2181,7 +2190,11 @@ move=> Pj; apply/seteqP; split => [t [Fjt UFt] i Pi|t UFt].
 by split=> [|[k [Pk kj]] [Fjt]]; [|apply]; exact: UFt.
 Qed.
 
-Definition bigcup2 T (A B : set T) : nat -> set T :=
+Definition sequence R := nat -> R.
+
+Notation "R ^nat" := (sequence R) : type_scope.
```
hoheinzollern, line 2196: why is this notation introduced here? isn't it part of mathcomp already?

hoheinzollern (approved):
Other than the comment about notation, looks good.

theories/sequences.v
affeldt-aist: This also needs to be moved to `classical_sets.v`

classical/classical_sets.v
affeldt-aist: No, it is not part of mathcomp.
I am not sure it is going to make it to mathcomp soon because it is about sequences and their theory is really developed in MathComp-Analysis.
This is the reason why I haven't even thought of moving it to `unstable.v` (the staging area).

classical/classical_sets.v
affeldt-aist: `bigcup_bigsetU_bigcup` can indeed be changed but of course it like adding an empty set to the big union
so I am not sure it is interesting, what do you think at the look of the last commit?

If by "around" you mean the next lemma `bigcup_mkord_ord`, this is inherent to the use of `inord`.

affeldt-aist: In fact, I would like to revert the last commit (https://github.com/math-comp/analysis/pull/1979/changes/eba1042c49f3306742661a8b0c6e27b0f62206d3) that experiments with changing the statement of `bigcup_bigsetU_bigcup`, not really convincing to me.

affeldt-aist: I removed it then.

affeldt-aist:
Thanks for the reviews @t6s @hoheinzollern . I'll merge when the CI is green.

PR #2004, opened 2026-06-23
mv squeeze for R earlier

affeldt-aist:

This PR is essentially moving `squeeze_cvgr` from `normed_module.v` to the `topology_theory` directory
so that it is available to develop the theory of TVSs.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -226,6 +226,13 @@
     `ae_eq_Radon_Nikodym_SigmaFinite`, `Radon_Nikodym_change_of_variables`,
     `Radon_Nikodym_cscale`, `Radon_Nikodym_cadd`, `Radon_Nikodym_chain_rule`
 
+- moved from `normed_module.v` to `metric_structure.v`
+  + lemma `squeeze_cvgr`
+
+- moved from `pseudometric_normed_Zmodule.v` to `metric_structure.v`
+  + lemmas `real_cvgr_lt`, `real_cvgr_le`, `real_cvgr_le`, `real_cvgr_gt`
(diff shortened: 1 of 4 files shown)


affeldt-aist:
CI green (fyi: @mkerjean )

mkerjean (approved):
Looks great, thanks. As the CI is green it can be merged. Could you explain why the necessity to duplicate  `cvgrPdistC_lt ` or `cvgr_distC_lt `  ? Would solving it necessitate the creation of a pseudometric type?

affeldt-aist:
> Looks great, thanks. As the CI is green it can be merged. Could you explain why the necessity to duplicate `cvgrPdistC_lt ` or `cvgr_distC_lt ` ?

By putting a `Let` here, I can prove the needed lemma with a minimal assumption (`R : numFieldType`)
and I do not commit to anything (it is a `Let`, it can't be used outside).
If I were to put `cvgrPdistC_lt` in the module `metricType_numDomainType` along with other generalizations
(generalizations from normed modules to metric space, WIP),
I would need `R : realFieldType` to be able to use `mdist x y = mdist y x`,
which would cause the section (with the lemmas `real_cvgr_lt`, etc.) to require `R : realFieldType` instead of `numFieldType`, which would be a regression compared with the current state of MathComp-Analysis.

So this is really just a way to make an incremental progress (put the squeeze theorem in the right place and
move a minimal set of lemmas without losing "generality") and defer big changes (generalization of more lemmas
to metric spaces) to later.

> Would solving it necessitate the creation of a pseudometric type?

Sorry, I can't see it. Which type are you thinking of? Pseudometric in the MCA sense or in the sense "d(x,y)=0 does not necessarily imply x=y"? In the latter case, it does not seem to be the incriminating property, that's why I can't see how it would help right away.

PR #2009, opened 2026-06-29
Renamed and added lemmas for setD and bigcup

Brixfoly:
The lemma bigcupDr was not that well named compared to setD_bigcupl for example.
Plus the equality was not in the same way, and as a result everytime it was rewrited in mathcomp,
it was with a -. This PR renames it, adds a similar lemma for bigcap (because I will need it in my work)
and renames everywhere the previous name was used.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


affeldt-aist (approved):
Thanks for the addition and the fix!

PR #2023, opened 2026-07-14
Derivative limit theorem, exponential growth, and more

amolinamounier:
##### Motivation for this change

This PR contains several lemmas I proved during an internship based on mathcomp-analysis, most of them utility lemmas, but also some new results:

- Derivative of a function of the form `fun x => k x *: f x` for any `normedModType`
- Derivation rules for shifted functions, e.g. `(fun x => f (x + a))^`() x = f^`() (x + a)`
- "Derivative limit theorem": if $$f : \mathbb{R} \to \mathbb{R}$$ is continuous at $$a \in \mathbb{R}$$, known to be derivable around (but not at) $$a$$, and $$\lim_{x \to a} f'(x) = \ell$$, then $$f$$ is derivable at $$a$$ with derivative $$\ell$$.
- Exponential growth domination : for any $$a, b \in \mathbb{R}$$, $$b < 0$$, $$x^a e^{bx} \underset{x \to +\infty}{\to} 0$$. Stated under a few different forms for `expR`, `powR`, `exp`.
- `continuous_FTC1` analogue for integrals of the form $$x \mapsto \int_x^\infty f(t) \mathrm{d} t$$
- Versions of l'Hôpital's rule for $$\frac{+\infty}{+\infty}$$ indeterminate forms

Perhaps it would be a good idea to split this into several PRs?

##### Checklist

- [x] added corresponding entries in `CHANGELOG_TMP.md`
- [ ] merge into `CHANGELOG_UNRELEASED.md` before merge

- [x] ~added corresponding documentation in the headers~

##### Merge policy

As a rule of thumb:
- PRs with several commits that make sense individually and that
  all compile are preferentially merged into master.
- PRs with disorganized commits are very likely to be squash-rebased.

##### Reminder to reviewers

- Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs)
- Put a milestone if possible
- Check labels

PR #2024, opened 2026-07-15
Utility lemmas in filter.v

amolinamounier:

Split from #2023

##### Checklist

- [x] ~~added corresponding documentation in the headers~~

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/filter.v
+++ b/classical/filter.v
@@ -925,9 +925,22 @@ Lemma near_eq_cvg {T U} {F : set_system T} {FF : Filter F} (f g : T -> U) :
   {near F, f =1 g} -> g @ F `=>` f @ F.
 Proof. by move=> eq_fg P /=; apply: filterS2 eq_fg => x /= <-. Qed.
 
+Lemma near_eq_cvgE {T U} {F : set_system T} {FF : Filter F} (f g : T -> U) :
+  {near F, f =1 g} -> f @ F = g @ F.
+Proof.
+move=> eq_fg.
+apply/seteqP; split; apply: near_eq_cvg => //.
+by near do symmetry.
+Unshelve. all: by end_near.
+Qed.
+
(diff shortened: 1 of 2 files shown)


classical/filter.v
```diff
@@ -937,6 +950,14 @@ Lemma eq_is_cvg (T : Type) (T' : pnbhsType) (F : set_system T) (f g : T -> T') :
   f =1 g -> cvg (f @ F) = cvg (g @ F).
 Proof. by move=> /funext->. Qed.
 
+Lemma near_eq_lim (T : Type) (T' : pnbhsType) {F : set_system T} {FF : Filter F} (f g : T -> T') :
+  {near F, f =1 g} -> lim (f @ F) = lim (g @ F).
+Proof. by move=> /near_eq_cvgE ->. Qed.
+
+Lemma cvg_to_eq {T : nbhsType} {F : set_system T} (l l' : T) :
```
affeldt-aist: That does not look very useful a priori.

amolinamounier: I found it useful in cases where the convergence is easy to close out with lemmas like `cvgD`, `cvgM` with an existential as the target, but where the end result needs some rewriting to be shown equal to the desired quantity, somewhat like `is_derive_eq`.

affeldt-aist: I think that `is_derive_eq` comes with typeclass instances so that when it is applied, it returns only one goal (the equality). Maybe this is a difference with `cvg_to_eq`?

affeldt-aist: What about we remove this lemma from this PR and defer its inclusion to a PR that actually uses it? (That would unlock this PR because the other lemmas looks ok to me.)

affeldt-aist (approved):

PR #2029, opened 2026-07-15
normed module identity convergence lemmas

amolinamounier:

Split from #2023. Adds convenient `cvg` lemmas that involves the additive and multiplicative identities of normed modules.

##### Checklist

- [x] ~added corresponding documentation in the headers~

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels


theories/normedtype_theory/pseudometric_normed_Zmodule.v
```diff
@@ -1098,6 +1098,45 @@ Qed.
 Lemma is_cvgDrE f g : cvg (f @ F) -> cvg ((f + g) @ F) = cvg (g @ F).
 Proof. by rewrite addrC; apply: is_cvgDlE. Qed.
 
+Lemma cvgDl f a b : f @ F --> b -> a + f x @[x --> F] --> a + b.
```
affeldt-aist: What about `cvgcstD` to suggest that the constant is on the left (and `cvgDcst` instead of `cvgDr`)?
Because `cvgDl` does not really say what is on the left.

But in fact I am not sure that `cvgDl`, `cvgDr`, `cvgBl`, `cvgBr` are very useful,
because it is just `cvg{D,B}` with `cvg_cst`.
The application of `cvg_cst` should maybe be automatic instead
(and maybe it was thought as such because there is actually a Hint in `topology_structure.v` but it is not working here). Fixing the latter Hint could be better.

amolinamounier: Yes, I think that would be a better idea; I did find it a bit odd that `cvg_cst` didn't trigger automatically.

affeldt-aist: Trying to fix that: https://github.com/math-comp/analysis/pull/2046

affeldt-aist: So now `cvg_cst` and `is_cvg_cst` are true hints, this PR can maybe be simplified.

theories/normedtype_theory/pseudometric_normed_Zmodule.v
```diff
@@ -1098,6 +1098,33 @@ Qed.
 Lemma is_cvgDrE f g : cvg (f @ F) -> cvg ((f + g) @ F) = cvg (g @ F).
 Proof. by rewrite addrC; apply: is_cvgDlE. Qed.
 
+Lemma cvg0D f g a : f @ F --> 0 -> g @ F --> a -> f x + g x @[x --> F] --> a.
+Proof. by move=> /cvgD /[apply]; rewrite add0r. Qed.
+
+Lemma cvg0DC f a : f @ F --> 0 -> f x + a @[x --> F] --> a.
```
affeldt-aist: I was puzzled at first by the naming because `C` as a suffix often means "commutative". Here it is "constant", right?

amolinamounier: Yes, I remember seeing `C` for constant being used in some places/doc. Should it change to `cst`?

affeldt-aist: Iirc it is used for polynomials in MathComp, but indeed in MathComp-Analysis we have been using `cst`, which I find less surprising.

amolinamounier: Alright, I'll change it in a bit.

affeldt-aist (approved):
I have added `cvgZ` variants and tried to find a handful of places in the code where the new lemmas can actually be used. Since `cvg_cst` is now a proper hint, I am not sure that the `cst` variants are still useful. I tentatively removed them. If you are fine with that change, I think that we can merge. Note that we can always reintroduce the `cst` versions later if really needed.

amolinamounier:
I generalized the lemmas to be in line with the others, and added a `cvgZ0` lemma. If there was a file for normed algebras I would have added a `cvgZ1` as well.

PR #2027, opened 2026-07-15
New topology lemmas

amolinamounier:
##### Motivation for this change

Split from #2023.
Generalized and moved lemma `cvg_at_right_left_dnbhs` from `metric_structure.v` to `num_topology.v`

##### Checklist

- [x] added corresponding entries in `CHANGELOG_UNRELEASED.md`

- [x] ~~added corresponding documentation in the headers~~

Reference: [How to document](https://github.com/math-comp/math-comp/wiki/How-to-document)

##### Merge policy

As a rule of thumb:
- PRs with several commits that make sense individually and that
  all compile are preferentially merged into master.
- PRs with disorganized commits are very likely to be squash-rebased.

##### Reminder to reviewers

- Read [this Checklist](https://github.com/math-comp/math-comp/wiki/Checklist-for-creating-and-review-PRs)
- Put a milestone if possible
- Check labels

PR #2030, opened 2026-07-15
normed type neighborhood lemmas

amolinamounier:

Split from #2023. Depends on #2024 and #2027.

##### Checklist

- [x] Merge #2024 
- [x] Merge #2027 

- [x] ~added corresponding documentation in the headers~

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -20,215 +20,19 @@
 - in `num_topology.v`:
   + lemmas `at_rightD`, `at_leftD`, `near_at_rightD`, `near_at_leftD`,
     `at_left_shift`, `at_right_shift`
-- in `classical_sets.v`:
-  + lemmas `setI_closed_setT`, `setI_closed_set0`
 
-- in `measurable_function.v`:
-  + lemma `g_sigma_algebra_preimage_comp`
+- in `num_normedtype.v`,
+  + lemmas `pinftyV`, `ninftyV`, `cvgryV`, `cvgrNyV`, `lt0_cvgMlNy`, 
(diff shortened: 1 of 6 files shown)


affeldt-aist:
(not yet reviewed but rebased)

affeldt-aist:
The statement of `near_shiftE` and `nbhsDl` are the same but on different structures that maybe ought to share an interface. This is an observation that has already been done for other lemmas, so maybe this is time to reconsider the hierarchy of mathematical structures to factorize code.

See
https://github.com/math-comp/analysis/blob/master/theories/normedtype_theory/tvs.v#L122-L128

affeldt-aist:
The duplication of code is calling for a more in-depth review (which might end up in adding structures) so we postpone to the next release (fyi: @mkerjean )

affeldt-aist:
Just a thought:
We long had `PseudoMetricNormedZmod` defined with a `numDomainType` parameter.
Recently, we have added `PseudoMetricNormedZmod0` and made  `PseudoMetricNormedZmod` derive from `Metric`.
In fact, as a consequence, many lemmas can be changed from `PseudoMetricNormedZmod` down to `PseudoMetricNormedZmod0` without impacting the code.
Which makes me wonder whether we shouldn't make `PseudoMetricNormedZmod` have a `numFieldType` as a parameter instead of sticking to `numDomainType`: this actually comes in the way when trying to refactor the code as hinted at in this conversation.

mkerjean:
I'm happy to do a review but I'm lacking a bit of context. Why are these lemmas duplicated ? What's the goal of this PR ?

PR #2060, opened 2026-07-29
is_derive_trmx

yosakaon:
This PR introduces `is_derive_trmx` instance, which provides automatic differentiation for transposed matrices. 
It is particularly useful as the development uses a row-vector convention rather than the more common column-vector convention found in physics and mechanics literature.

Most of the code is from robot-rocq, co-authored by @affeldt-aist. 

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -647,3 +647,21 @@ End Theory.
 Module Import Exports. HB.reexport. End Exports.
 End Norm.
 Export Norm.Exports.
+
+From mathcomp Require Import interval_inference.
+Section nng_comlaw.
+Import Num.Def.
+Context {K : realFieldType}.
+
+Let nng_max0r : left_id ((0 : K)%:nng) (@maxr {nonneg K}).
+Proof.
+move=> x; rewrite /maxr; case: ifPn => //.
+rewrite -leNgt => x0.
+apply/eqP; rewrite eq_le x0 andbT.
+by have : 0 <= x%:nngnum by []. (* NB: why isn't this automatic? *)
+Qed.
+
+HB.instance Definition _ :=
+  Monoid.isComLaw.Build {nonneg K} 0%:nng maxr maxA maxC nng_max0r.
+
+End nng_comlaw.
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -38,6 +38,8 @@ From mathcomp Require Import vector archimedean interval matrix.
 (*                           K is a numDomainType.                            *)
 (*                           L is a lmodType K.                               *)
 (*                           The HB class is Norm.                            *)
+(*    Module MaxNngComLaw == contains on instance of Monoid.isComLaw on       *)
+(*                           (@maxr {nonneg _}) to be used with caution       *)
 (* ```                                                                        *)
 (*                                                                            *)
 (******************************************************************************)
@@ -649,7 +651,10 @@ End Norm.
 Export Norm.Exports.
 
 From mathcomp Require Import interval_inference.
-Section nng_comlaw.
+(* NB: This module adds an instance of `Monoid.Law.sort` on `Order.max`
+   although there is already one. Use with caution. *)
+Module MaxNngComLaw.
+Section max_nng_comlaw.
 Import Num.Def.
 Context {K : realFieldType}.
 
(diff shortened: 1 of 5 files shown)


classical/unstable.v
```diff
@@ -647,3 +647,21 @@ End Theory.
...
+Section nng_comlaw.
+Import Num.Def.
+Context {K : realFieldType}.
+
+Let nng_max0r : left_id ((0 : K)%:nng) (@maxr {nonneg K}).
+Proof.
+move=> x; rewrite /maxr; case: ifPn => //.
+rewrite -leNgt => x0.
+apply/eqP; rewrite eq_le x0 andbT.
+by have : 0 <= x%:nngnum by []. (* NB: why isn't this automatic? *)
```
affeldt-aist: I was a bit surprised that we need to insert this `have`. Maybe there is some automation in `interval_inference.v` that does not work as expected? @proux01 ?

theories/lebesgue_integral_theory/radon_nikodym.v
affeldt-aist, line 266: @proux If this file (`radon_nikodym.v`) imports `unstable.v` (which now contains an instance of ComLaw for `(@maxr {nonneg K})`, then this line produces an error that I cannot explain.
The issue is easily avoided by not importint `unstable.v` but this is certainly a problem that will occur again later. Could you take a look?

affeldt-aist:
@holgerthies I found two things puzzling about the the instance on `@maxr {nonneg _}`, I pinged @proux01 but maybe you already figured it out?

theories/lebesgue_integral_theory/radon_nikodym.v
proux01: Well, using `@maxr {nonneg K}` is probably asking for troubles. Seems safer to use `@maxr K` and reinfer the nonnegativity of the result when needed with an interval instance on maxr (which we should already have). I would refrain from adding this kind of instances.

theories/derive.v
```diff
@@ -2511,6 +2511,54 @@ apply/derivable_mxP => i0 j0.
 by have [] := MdM i0 j0.
 Qed.
 
+Lemma derivable_trmx {m n} (M : V -> 'M[R]_(m, n)) t v :
+  derivable (fun x => (M x)^T) t v = derivable M t v.
+Proof.
+rewrite propeqE; split; rewrite /derivable/=.
```
proux01: This looks like essentially two times the same proof, couldn't it be factored? (I would expect ``derivable M -> derivable M^T` to be enough to prove `derivable M^T -> derivable M` since `derivable M^T -> derivable M^T^T` and `M^T^T = M`)

theories/lebesgue_integral_theory/radon_nikodym.v
affeldt-aist: When dealing with bigmax, `{nonneg _}` is nice because we have a neutral elements (`0`) and therefore access to more bigop lemmas. We will try again without it.

proux01: Yes but then everything comes with an interval and a proof that the value is in that interval, so you loose equality, you'll probably rapidly need morphism lemmas for `%:num` and bigops and thing can become painful. So not sure we want to engage in that direction. I agree that writing `(\bigmax_i (F i)%:num)%:nng` can be a bit heavy though, not sure what's best.

proux01: And to explain the error:
```rocq
Print Canonical Projections Order.max (* HB.about could maybe also do the job *)
(* without loading unstable.v *)
(* Order.max <- Monoid.Law.sort ( Order_max__canonical__Monoid_Law ) *)
(* Order_max__canonical__Monoid_Law : forall {R : realDomainType}, Monoid.Law.type -oo%E (from constructive_ereal.v) *)
(* after loading it *)
(* Order.max <- Monoid.Law.sort ( unstable.Order_max__canonical__Monoid_Law ) *)
```
due to canonical structures matching on a single key, there can only be a single instance of `Monoid.Law.sort` on `Order.max` and we have to choose between them. By the way, having it on ereal is already a bit strange.
 I guess the correct fix would be to have a structure of "max is a monoid operator", which maybe is `Order.BPreorder`, and put the monoid instance on it (and then ensure that both ereal and nonneg have that BPreorder instance).

classical/unstable.v
proux01: Goal is `widen_itv 0%:itv <= x` (with the order on Itv.t), of course this is convertible to `0 <= x%:num` (as shown by `rewrite /Order.le/= /Order.PreCancelPartial.le/=`) but trigerring hints for auto for `ge0` do not match that syntactic pattern https://github.com/math-comp/math-comp/blob/900e37912dbecbd3c04d3c1b320efe721d3f56dd/algebra/interval_inference.v#L770 . So I'd say it's not expected to be simpler, except maybe `by rewrite -num_le/=`

affeldt-aist: Thanks for the explanation! I'll record it as a comment inside the script.

theories/lebesgue_integral_theory/radon_nikodym.v
affeldt-aist: For the time being, I have recorded the problem as an issue https://github.com/math-comp/analysis/issues/2093.

PR #2062, opened 2026-07-29
Add set of lemmas for esum

lyonel2017:

Set of lemmas for `esum` extracted from #2049.

##### Checklist

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/theories/esum.v
+++ b/theories/esum.v
@@ -542,7 +542,8 @@ Lemma esum_ge {R : realType} {T : choiceType} (I : set T) (f : T -> \bar R) x :
   x <= \esum_(i in I) f i.
 Proof. by move=> f0 If; rewrite ge0_esum// PosEsum.pos_esum_ge. Qed.
 
-Lemma esum_unit {R : realType} {T : choiceType} (f : T -> \bar R)  x :
+(* TODO: is this lemma really useful? *)
+Lemma esum_if_eq_op_set1 {R : realType} {T : choiceType} (f : T -> \bar R)  x :
   \esum_(i in [set: T]) (if x == i then f i else 0) = f x.
 Proof. by rewrite esum_if_eq_op esum_set1. Qed.
 
@@ -1149,7 +1150,7 @@ Proof.
 move=> h d0 dfin.
 have -> : d = (fine d)%:E by rewrite fineK.
 have ? : (0 <= fine d)%R  by rewrite -lee_fin fineK.
-rewrite esumE ge0_funeposM// ge0_funenegM// !esumZ// -muleBr//.
+rewrite esumE (ge0_funeposM f)// (ge0_funenegM f)// !esumZ// -muleBr//.
   by rewrite fin_num_adde_defr// esummable_esum_funepos.
 by rewrite [in RHS]esumE.
 Qed.
--- a/theories/esum.v
+++ b/theories/esum.v
@@ -19,7 +19,7 @@ From mathcomp Require Import topology sequences normedtype numfun.
 (*                       function whose codomain is included in the extended  *)
 (*                       reals; it is 0 if I = set0 and sup(\sum_A a) where A *)
 (*                       is a finite set included in I o.w.                   *)
-(*       summable D f := \esum_(x in D) `| f x | < +oo                        *)
+(*      esummable D f := \esum_(x in D) `| f x | < +oo                        *)
 (* ```                                                                        *)
 (*                                                                            *)
 (******************************************************************************)
@@ -38,7 +38,7 @@ Local Open Scope ring_scope.
 Local Open Scope ereal_scope.
 
 Section set_of_fset_in_a_set.
-Variable (T : choiceType).
+Context {T : choiceType}.
 Implicit Type S : set T.
 
 Definition fsets S : set_system T := [set F | finite_set F /\ F `<=` S].
@@ -64,7 +64,7 @@ End set_of_fset_in_a_set.
   define a generic `esum` below, which should be preferred *)
 Module PosEsum.
 Section posesum.
-Variables (R : realFieldType) (T : choiceType).
+Context {R : realFieldType} {T : choiceType}.
 Implicit Types (S : set T) (f g : T -> \bar R).
 
 Definition pos_esum S g := ereal_sup [set \sum_(x \in B) g x | B in fsets S].
@@ -117,7 +117,7 @@ End posesum.
 Arguments eq_pos_esum {R T} S f g.
 
 Section posesum_realType.
-Variables (R : realType) (T : choiceType).
+Context {R : realType} {T : choiceType}.
 Implicit Types (S : set T) (f g : T -> \bar R).
 
 Local Notation "\esum_ ( i 'in' P ) A" := (pos_esum P (fun i => A)).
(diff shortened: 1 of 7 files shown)


theories/esum.v
```diff
@@ -155,13 +155,51 @@ Lemma pos_esum_ge (T1 : choiceType) (I : set T1) (a : T1 -> \bar R) x :
...
 
-Lemma le_pos_esum S f g : (forall i, S i -> f i <= g i) ->
+Lemma pos_neq0_esum (I : set T) (a : T -> \bar R) :
+  \esum_(i in I) a i <> 0 -> exists i,  a i <> 0.
+Proof.
+move=> ?. apply/existsp_asboolPn /asboolPn => h.
+have // : (\esum_(i in I) a i = 0); by apply pos_esum1.
+Qed.
+
+Lemma pos_esum_ge1 (I : set T) (f: T -> \bar R) :
```
affeldt-aist:
```suggestion
Lemma pos_esum_ge1 (I : set T) (f : T -> \bar R) :
```

theories/esum.v
```diff
@@ -155,13 +155,51 @@ Lemma pos_esum_ge (T1 : choiceType) (I : set T1) (a : T1 -> \bar R) x :
   x <= \esum_(i in I) a i.
 Proof. by move=> [X IX /le_trans->//]; apply: ereal_sup_ubound; exists X. Qed.
 
-Lemma le_pos_esum S f g : (forall i, S i -> f i <= g i) ->
+Lemma pos_neq0_esum (I : set T) (a : T -> \bar R) :
+  \esum_(i in I) a i <> 0 -> exists i,  a i <> 0.
```
affeldt-aist:
```suggestion
  \esum_(i in I) a i <> 0 -> exists i, a i <> 0.
```

theories/esum.v
```diff
@@ -155,13 +155,51 @@ Lemma pos_esum_ge (T1 : choiceType) (I : set T1) (a : T1 -> \bar R) x :
...
+Lemma le_pos_esum {U : choiceType} (S: set U) (f g: U -> \bar R) : (forall i, S i -> f i <= g i) ->
   \esum_(i in S) f i <= \esum_(i in S) g i.
 Proof.
 move=> fg; rewrite ge_ereal_sup => //= _ [X [finX XS]] <-.
 by rewrite pos_esum_ge//; exists X => //; apply: lee_fsum => // t /XS /fg.
 Qed.
 
+Lemma le_pos_esum_fine {U : choiceType} (f: T -> U -> \bar R):
+  (forall x y, 0 <= f x y)%E ->
+  (\esum_(i in [set: U]) (fine (\esum_(x in [set: T]) f x i))%:E <=
```
affeldt-aist: Can't this be generalized to any `A : set U` and `B : set T`?

theories/esum.v
```diff
@@ -361,6 +399,12 @@ rewrite /esum PosEsum.ge0_pos_esum_funepos// PosEsum.ge0_pos_esum_funeneg//.
 by rewrite sube0.
 Qed.
 
+Lemma esum_pos_esum S f : (forall x, S x -> 0 <= f x) ->
```
affeldt-aist: This lemma does not look necessary since it is essentially `ge0_esum`.

theories/esum.v
```diff
@@ -385,13 +429,60 @@ Section esum_realType.
...
+Qed.
+
+(* Lemma le_esum S f g : (forall x, S x -> 0 <= f x) -> *)
+(*   (forall x, S x -> f x <= g x) -> *)
+(*   \esum_(x in S) f x <= \esum_(x in S) g x. *)
+(* Proof. *)
+(* move=> f0 leS; have g0 x : S x -> 0 <= g x. *)
+(*   by move=> /[dup] Ax /leS; apply: le_trans; exact: f0 Ax. *)
+(* by rewrite !ge0_esum// PosEsum.le_pos_esum. *)
+(* Qed. *)
```
affeldt-aist:
```suggestion
```

theories/esum.v
```diff
@@ -483,6 +581,13 @@ Lemma esum_ge {R : realType} {T : choiceType} (I : set T) (f : T -> \bar R) x :
   x <= \esum_(i in I) f i.
 Proof. by move=> f0 If; rewrite ge0_esum// PosEsum.pos_esum_ge. Qed.
 
+Lemma esum_unit {R : realType} {T : choiceType} (f : T -> \bar R)  x :
+  \esum_(i in [set:T]) (if x == i then f i else 0) = f x.
+Proof.
+  rewrite esum_if_eq_op.
+  by rewrite esum_set1.
```
affeldt-aist:
```suggestion
rewrite esum_if_eq_op.
by rewrite esum_set1.
```

theories/esum.v
```diff
@@ -493,6 +598,18 @@ exists [set x]; first by split => // t ->.
 by rewrite -esum_set1 esum_fset// => i ->; exact: f0.
 Qed.
 
+Lemma neq0_esum {R : realType} {T : choiceType} (I : set T) (a : T -> \bar R) :
+  \esum_(i in I) a i <> 0 -> exists i,  a i <> 0.
```
affeldt-aist:
```suggestion
  \esum_(i in I) a i <> 0 -> exists i, a i <> 0.
```

theories/esum.v
```diff
@@ -493,6 +598,18 @@ exists [set x]; first by split => // t ->.
...
+Proof.
+move=> ?. apply/existsp_asboolPn /asboolPn => h.
+have // : (\esum_(i in I) a i = 0); by apply esum1.
+Qed.
+
+Lemma esum_ge1 {R : realType} {T : choiceType} (I : set T) (f: T -> \bar R) :
```
affeldt-aist:
```suggestion
Lemma esum_ge1 {R : realType} {T : choiceType} (I : set T) (f : T -> \bar R) :
```

theories/esum.v
```diff
@@ -812,6 +945,50 @@ apply: PosEsum.le_pos_esum => t Dt.
...
+under eq_esum do rewrite abseM.
+by rewrite esumZ // lte_mul_pinfty //= abse_fin_num.
+Qed.
+
+Lemma summableZr D f c :
+c \is a fin_num -> summable D f -> summable D (fun x => f x * c).
+Proof. by move=> ??; apply/summable_muleC /summableZ. Qed.
+
+Lemma summableMl D f1 f2 :
+  (exists M, (forall x, D x -> `|f1 x| <= M) /\ M \is a fin_num) ->
```
affeldt-aist: Use `exists2` is often a good idea.

affeldt-aist:
Sorry for the random, minor comments.
Yet, I am already concerned by the fact that we have `summable` defined both in `esum.v` and in `realsum.v`: we might want to distinguish one or the other by adding a small `e` or a small `r` in the identifier.

affeldt-aist (changes requested):

affeldt-aist:
I made a first pass. I have not yet looked at section `esum_summable`.
I'll complete this review asap.

Maybe we do not need `pos_neq0_esum`
(btw, let's keep the substring `pos_esum` in identifiers
so that we remember easily that we are inside the `PosEsum` module)
since it is not used and in fact the more general
version `neq0_esum` (rather: `esum_neq0`) does not require
non-negativity as an hypothesis
(so far, the rationale was to use the `PosEsum` module to hold
restricted versions of `esum` lemmas for non-negative functions).

`pos_sum_esum_ge`: we can suppress because this is really
just a special case of `pos_esum_ge` and is used only once.

`summable_muleC` is maybe not necessary since it is really
just `under eq_fun do rewrite muleC`

Pending:
`pos_esum_ge1`/`esum_ge1`: there might be a better name
(`_ge1` looks too much like `1 <= _`) but I have not idea yet

`esum_unit` does not mention `unit`, maybe `esum_if_eq_op_set1`
but the script is so short that maybe the lemma is not really useful

TODO: the changelog should mention that `le_esum` has been generalized

fyi: @strub

affeldt-aist (approved):
Thank you very much for having taken the time to sort out these lemmas!

I have generalized lemmas (e.g.,  eq_summable (now eq_esummable),
 le_summable (now le_esummable) to use {in D, f = g} instead of f =1 g, etc.).

`esummable_esumZ_pos` is not a `Let`: it is used to prove a more general lemma.

I have added `esummable_esum_funeneg` since we have its counterpart.

`esum_posneg` has been renamed `esumE` and used to simplify
proof (by not making appeared the "positive" version).

I also used the new lemmas to simplify a bit `realsum.v`.

I have not yet checked whether `le_esum_fine` and `le_pos_esum_fine` 
are really useful but we will use them in a future PR so we can defer this
discussion.

I haven't found a better name for `esum_ge1`.

Unless somebody jumps in (@strub , @t6s ?), we'll merge soon.

PR #2082, opened 2026-08-15
generalize continuous_comp_cvg

t6s:

This PR is a strict generalization of lemma `continuous_comp_cvg` that has been in ~`pseudometric_structure.v`~ `pseudometric_normed_Zmodule.v`.

NB: the name `continuous_comp_cvg` does not look very good considering similar lemmas `continuous_cvg` and `cvg_comp`.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -84,7 +84,7 @@
 - in `esum.v`:
   + lemmma `le_esum`
 
-- from `pseudometric_structure.v` to `topology_structure.v`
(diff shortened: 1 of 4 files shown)


affeldt-aist (approved):
Thanks for the generalization! Let's add `id_continuous` because incidentally we also needed the lemma in another piece of work (and I also remember this was due some performance problem but maybe not a `Search`).

PR #2083, opened 2026-08-16
add compatibility lemmas for Stdlib `Rcos` and `Rsin` 

t6s:

Context:
https://rocq-prover.zulipchat.com/#narrow/channel/237666-math-comp-analysis/topic/relating.20trigo.2Ecos.20to.20Rtrigo_def.2Ecos.2C.20etc.2E

This PR adds compatibility lemmas for a few trigonometry functions defined in Stdlib Reals,
namely `Rcos` and `Rsin`.   A lemma for π is also added.

##### Checklist

##### Merge policy

As a rule of thumb:

##### Reminder to reviewers

- Put a milestone if possible
- Check labels

### DIFF
--- a/classical/unstable.v
+++ b/classical/unstable.v
@@ -647,39 +647,3 @@ End Theory.
 Module Import Exports. HB.reexport. End Exports.
 End Norm.
 Export Norm.Exports.
-
-(* PR to mathcomp in progress (#1637) *)
-Section big_nat_dvdn.
-
-Lemma iotaS (m n : nat) : iota m n.+1 = rcons (iota m n) (m + n)%N.
-Proof. by rewrite -addn1 iotaD cats1. Qed.
-
-Lemma index_iotaS (m n : nat) :
-  (m <= n)%N -> index_iota m n.+1 = rcons (index_iota m n) n.
-Proof. by move=> ?; rewrite /index_iota subSn// iotaS subnKC. Qed.
-
-Lemma big_nat_recr_op (R : Type) (idx : R) (op : R -> R -> R)
-  (n m : nat) (P : pred nat) (F : nat -> R) :
-  (m <= n)%N ->
-  let idx' := if P n then op (F n) idx else idx in
-  \big[op/idx]_(m <= i < n.+1 | P i) F i = \big[op/idx']_(m <= i < n | P i) F i.
-Proof. by move=> ?; rewrite index_iotaS// big_rcons_op. Qed.
-
-Lemma big_nat_dvdn (R : Type) (idx : R) (op : R -> R -> R)
-  (n d : nat) (F : nat -> R) :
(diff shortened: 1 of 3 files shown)


t6s:
The proof turned out to be not as simple as for the `exp` case, due to a subtle difference in the defining series for `Rcos` and `cos`:
the former is
1 - (x^2/2!) + (x^4/4!) ...
while the latter is
1 + 0 - (x^2/2!) + 0 + (x^4/4!) + 0 ...

That said, the proof script for `RcosE` seems a bit too long, and there could be a better proof strategy.

classical/unstable.v
```diff
@@ -647,3 +647,39 @@ End Theory.
 Module Import Exports. HB.reexport. End Exports.
 End Norm.
 Export Norm.Exports.
+
+(* PR to mathcomp in progress (#1637) *)
+Section big_nat_dvdn.
```
proux01: Since https://github.com/math-comp/math-comp/pull/1637 is now essentially nmerged, we can probably move that to `mathcomp_compat.v`

proux01 (approved):

proux01:
> The proof turned out to be not as simple as for the `exp` case, due to a subtle difference in the defining series for `Rcos` and `cos`: the former is 1 - (x^2/2!) + (x^4/4!) ... while the latter is 1 + 0 - (x^2/2!) + 0 + (x^4/4!) + 0 ...

Could `cos_coeff'` and `cos_coeff'E` help here?

t6s:
> > The proof turned out to be not as simple as for the `exp` case, due to a subtle difference in the defining series for `Rcos` and `cos`: the former is 1 - (x^2/2!) + (x^4/4!) ... while the latter is 1 + 0 - (x^2/2!) + 0 + (x^4/4!) + 0 ...
> 
> Could `cos_coeff'` and `cos_coeff'E` help here?

That helps so much!!

classical/unstable.v
t6s: By `cos_coeff'` this part of code is now a stray (and thus removed from this PR).

What do you think about the MC PR 1637?  Would it be useful in its own?

proux01: You can consider https://github.com/math-comp/math-comp/pull/1637 merged (just waiting for CI to complete before pressing the button)

affeldt-aist:
By the way, this is not a useful comment anymore since the need has disappeared, but `big_nat_dvdn` should have gone to `mathcomp_compat.v` rather than unstable.

classical/unstable.v
proux01: Ho, I see, no longer needed. Well, let's merge it anyway, it can still be useful to someone in the future.

proux01:
Let's merge after CI confirmation.

